PR #2, opened 2026-05-11
Add polyglot example, channels integration, and Last-Event-ID SSE resume

joshbrooks:
## Summary

- **`examples/polyglot/`** — a standalone Django sample where every visible string on a marketing landing page is a `Translatable` row. Editing a row via the right-pane editor emits a `StreamEvent` to `translations:{langcode}` and every browser pinned to that language updates its preview in place. Mirrors `examples/chat/` layout (dev + prod settings, justfile recipes).
- **`fix(django_rakaia): Last-Event-ID resume`** — the SSE replay used to hardcode `[:20]` historical entries, so clients reconnecting to streams with more than 20 events never received the remainder until a new live event arrived. Now honors the `Last-Event-ID` header (sent automatically by `EventSource` on reconnect) and emits `id: <offset>` per event so the browser tracks its cursor.
- Rolls up the rest of the branch: ASGI async SSE endpoints, the django_rakaia API rework + Channels integration, the chat example, the test suite, and the move to a Zensical-rendered docs site.

## Test plan

- [x] `just test` — 188 tests pass
- [x] `just polyglot-dev` → http://localhost:8001, edit a string, click Save, verify another tab on the same `?lang=` updates without refresh
- [x] curl `Last-Event-ID: N` against `/streams/api/streams/<stream_id>/sse/` returns only entries with `offset > N`
- [x] `just dev` (chat sample) still works end-to-end (create/update/delete messages, SSE delivers `create`/`update`/`delete` events)
- [ ] `just serve` prod-style hypercorn + Redis run (smoke)


joshbrooks:
Closing as fully superseded. This branch's tip (`992d770`) is a direct ancestor of the `versioned-handlers` branch merged in #3 — that PR was built on top of this one, so its squash-merge already absorbed 100% of the commits here, including both the polyglot and chat example apps.

A rebase onto current `main` produces add/add conflicts across nearly every `django_rakaia` file (the same files, redeveloped), and the only content unique to this branch is the *pre-hardening* version of `admin.py` / `decorators.py` / `protocol_views.py` / `views.py` (f-string XSS, `@csrf_exempt`, `str(e)` error leak) — all of which were fixed on `main` after #3 landed. Merging this would regress those fixes.

Nothing to salvage; everything is on `main`.

PR #3, opened 2026-05-26
feat: versioned event handlers, upcasters, and replay

joshbrooks:
## Summary
- Adds a `replay()` subsystem that re-derives materialized state from a Rakaia stream by running pure handlers and applying returned `Effect`s via a pluggable executor.
- Handlers are registered with `[from_seq, to_seq)` sequence ranges; the dispatcher picks the version covering each event so a bug fix doesn't silently rewrite history.
- Upcasters normalise older event shapes forward by `schema_version` so handlers only ever see one shape.
- Ships `DjangoExecutor`, app autodiscovery (`handlers.py` / `upcasters.py` per app), `python manage.py replay`, and two read-only adapters (`ModelStreamReader` for virtual streams over a queryset, `DjangoStreamReader` for real `@stream_model`-populated streams).

## Motivation
Real production codebases consuming durable streams hit the same problem: business logic that interprets an event changes, but events already in storage were processed under the old interpretation. The standard workaround — iterate historical rows and re-save them so newer signals fire — is fragile, hand-rolled per-incident, and untracked. This subsystem makes the pattern first-class: handler versions are durable, replay is idempotent, drift is detected.

## What's in it
- **\`rakaia.effects\`** — \`Effect\` dataclass (\`op=\"update_or_create\" | \"external\"\`), \`Executor\` protocol, \`EffectCollisionError\`, disjoint-defaults check.
- **\`rakaia.registry\`** — \`HandlerRegistry\`, \`UpcasterRegistry\`, \`register_handler\` / \`register_upcaster\` decorators. Registrations persisted to reserved meta-streams (\`__rakaia__:handlers\`, \`__rakaia__:upcasters\`) so they survive restarts and provide their own audit log. Re-registration with identical identity is a no-op.
- **\`rakaia.replay\`** — \`replay(store, stream_path, executor, …)\` orchestrator. Pipeline: load → upcast chain → handler resolve (raises on gap) → drift check → call → effect collision check → executor.apply.
- **\`rakaia.source_hash\`** — SHA-256 of a function's source body for drift detection.
- **\`django_rakaia.effect_executor.DjangoExecutor\`** — applies update_or_create Effects in an atomic transaction.
- **\`django_rakaia.handlers_autodiscover\`** — walks INSTALLED_APPS on app ready().
- **\`django_rakaia.management.commands.replay\`** — \`python manage.py replay <stream> [--from N --to M --strict-drift --dry-run --include-external]\`.
- **\`django_rakaia.streams\`** — \`ModelStreamReader\` (queryset → virtual stream) and \`DjangoStreamReader\` (Stream/StreamEvent/StreamEntry → reader).
- **\`docs/versioned-handlers.md\`** — user guide with worked example.

## Design decisions worth flagging
- **Half-open intervals** \`[from, to)\` for version ranges; overlap rejected at registration, gaps raise at replay.
- **\`update_or_create\` is the only DB op**; covers create+update and makes replay idempotent by construction.
- **External effects** (email, third-party) tagged \`op=\"external\"\` and skipped on replay unless \`--include-external\`.
- **Producer declares \`schema_version\`** in the event payload; not added to \`AppendOptions\` so the wire protocol is unchanged.
- **Registries are themselves streams** — registration is an append, current state is a fold; no new storage primitive.
- **Drift policy** is warn-and-continue by default (developer-friendly), \`--strict-drift\` raises (CI/pre-deploy).

## Test plan
- [x] \`uv run pytest\` — 308 passed, 1 skipped (pre-existing)
- [x] \`uv run ruff check src/ tests/\` — clean
- [x] \`uv run pyright src/<new files>\` — clean
- [x] Effect collision detection (disjoint-defaults invariant)
- [x] Registry: overlap rejected, gap raises, version dispatch correct, source_hash captured, persistence durable across instances, re-registration idempotent
- [x] Upcasters: chain composes, missing link raises, target-lower-than-current raises
- [x] Replay: end-to-end synthetic stream, idempotency (run twice → identical effects), external-effect skip, gap-raises, drift policies (warn vs raise)
- [x] DjangoExecutor: idempotent apply, batch, unknown model raises clearly, external skipped, collision raises
- [x] Autodiscovery: handlers + upcasters in a test app are registered on app ready, autodiscover is idempotent
- [x] \`manage.py replay\`: basic apply, \`--dry-run\`, range bounds, \`--strict-drift\`
- [x] \`ModelStreamReader\` + \`DjangoStreamReader\`: ordered reads, multi-stream isolation, end-to-end replay through each

## Risk / rollout
- **Zero-impact opt-in**: nothing in this PR runs unless someone imports \`register_handler\`, registers something, and calls \`replay()\`. No schema changes.
- **Public API additions only** — no removals, no behavioural changes to existing rakaia functionality.

## Follow-up PRs (not in this branch)
- Wire \`@stream_model\` to optionally auto-dispatch handlers on append.
- Disk-persistent \`StreamStore\` for non-Django mode.

## Commits
1. \`feat(rakaia): add versioned event handlers, upcasters, and replay\` — core (11 files, ~2300 LOC).
2. \`feat(django_rakaia): wire versioned handlers into Django\` — executor, autodiscovery, replay command, stream adapters.
3. \`docs: add versioned-handlers user guide\` — docs/versioned-handlers.md + nav.


joshbrooks:
Pushed a hardening commit (f898282) on top of this branch. **Two behavior changes worth flagging for reviewers/clients:**

- **BREAKING — CSRF now enforced on `translation_create_update_api`**: removed `@csrf_exempt`. Any non-browser client POSTing to this endpoint must now send a CSRF token, or it gets a 403. The HTMX UI path was already protected, so the app's own UI is unaffected.
- **Auth tightened on read endpoints**: added `@login_required` to all stream/translation GET views, including `translations_api`, which was previously anonymous. Anonymous callers of these endpoints now get redirected to login.

Other (non-breaking) changes in the same commit:
- admin: escape stream identifiers via `format_html_join` (was an f-string XSS hole)
- offset allocation: `select_for_update()` row lock in `_get_next_offset`/`get_next_offset` to serialize concurrent appends (no-op on SQLite; `unique_together` is still the backstop there)
- `stream_events_api`: 400 on non-integer `after_offset`/`limit`, clamp `limit` to 1..200
- error handling: log traceback + generic 500 instead of leaking `str(e)`

Issue #22, opened 2026-07-19
RFC: Event envelope + history read-model (retire pghistory) — grounded in Partisipa/FormKit audit


joshbrooks:
## Summary

Design for a first-class **event envelope** (per-event metadata) and a **history read-model** in rakaia, so a durable stream can carry what `django-pghistory` carries and adopters can retire it. Last core promotion from the pghistory-retirement spike (#12).

**Revised after read-side (partisipa-import) and write-side (formkit-ninja) adopter review.** Both seats confirmed the envelope *shape* — `{label, ts, metadata, snapshot}` with an open `metadata` dict — is sufficient; no consumer needs something it can't express. The changes below are about the **decisions the first draft filed under "edge cases"**, chiefly **stream identity**, which is the actual gate on adoption.

Still a scoping RFC — not code. Comments welcome.

---

## What the real consumers read (evidence)

Audited against django-pghistory 3.9 + the live code. Three consumers + the auto-registered pghistory admin:

**1. `GET /history` audit API** (`partisipa-import` `form_submission/schema.py:607-671`). Per event: `version` (=`pgh_id`), `key` (obj UUID), `user`+`user_details`, `type` ∈ `{+,~,-}`, `fields` (**full snapshot**), `status`, `created_at`, `form_type`. Marker from label: `insert→+`, `delete→-`, else `→~`. **Actor precedence:** `pgh_context.metadata['user']` (editor, a user **PK int**) ?? `event.user_id` (row's own snapshotted `user` FK, the owner) → resolved to `full_name or username` by the consumer.

**2. `repair_blank_save_dataloss`** — lives on the **`formkit-ninja` write side** (correction from v1, which mis-attributed it to partisipa-import). Raw SQL over the event table: peak = `max(count(jsonb_object_keys(fields)))` per obj; restore = newest snapshot whose key-count equals the peak. Needs the **complete `fields` snapshot per event**. Reads no label/actor.

**3. pghistory admin** — `pgh_created_at, obj, pgh_diff (adjacent-snapshot diff), user, url` — `user`/`url` from `Context.metadata`.

**Actor capture:** `partisipa-import` registers `HistoryMiddleware` (`settings.py:102`) → `metadata = {"user": <pk int>, "url": path}`, only on mutating HTTP methods. `formkit-ninja` registers **no** middleware → `metadata['user']` is always null there and the row's `user` FK is the *sole* actor. Management commands / imports / GET → **no context → null actor** everywhere.

[...]

Issue #11, opened 2026-07-18
Partisipa still has to adopt the streams work we finished


joshbrooks:
## Context

Follow-up to #5 (single-form adoption gaps, closed by #6) and #7 (multi-form staged replay, spiked in #8). This issue tracks the remaining, mostly *low-effort* moves toward the real goal: **replacing Partisipa's `Submission` table with a rakaia stream** (Timor-Leste PNDS; `catalpainternational/partisipa-import` + `formkit-ninja`).

Two things were confirmed by reading the live code this session:

1. **The projector is already a hand-rolled event-sourcing system.** `Submission.fields` is the canonical log; `SeparatedSubmission` and `ida_forms.*` are read-model projections rebuilt by signals; `reimport_failed.py` literally "replays" a submission by calling `Submission.objects.get(pk).save()`; `post_migration_tasks.py` is a pile of keyed, idempotent, dry-run-able "rebuild the projection to match the log" jobs. Most of the quick wins below are *recognising something rakaia already does and deleting the bespoke version.*

2. **`pghistory` is load-bearing, not dead weight** — see Part A. Dropping it is the *destination*, not a quick win, and it carries an audit/recovery risk surface.

---

## Part A — "Drop pghistory for a streams-based alternative?"

**Verdict: it is used and it matters. Retire it *last*, behind a proven parity spike — not as step one.**

`django-pghistory` 3.9 tracks `Submission`, `SeparatedSubmission`, and `FormKitSchemaNode` (`@pghistory.track()`), and has **three live consumers**:

- **`GET /history` audit API** — `backend/form_submission/schema.py:606` → `api.py:1637/1665`. Reads the `pgh_event` model, maps `pgh_label` to `+`/`~`/`-`, returns per-version fields **plus the editing actor**.
- **Per-change actor context** — `pghistory.middleware.HistoryMiddleware` (`settings.py:102`) records who made each change; the API reads `h.pgh_context.metadata['user']`, with a regression test for the fallback path.
- **Disaster recovery** — `repair_blank_save_dataloss.py:105` queries the event table for each submission's historical *peak* key-count and restores a truncated record. It is the only source for that recovery.

Two facts that shape the port:

- **`pgtrigger` is a *separate* dependency and stays regardless** — change_id sync sequences, soft-delete, protect, and ordering triggers are unrelated to history.
[...]

PR #29, opened 2026-07-20
spike: SubmissionEvent (append log) → Submission (latest projection) — Decision #13

joshbrooks:
## What

A runnable spike proving **RFC #22 Decision #13 — the arrow-flip**: make the
event log the source of truth and the row a projection.

```
pghistory today:   Submission  ──trigger──▶  SubmissionEvent   (log is a shadow)
this design:       SubmissionEvent  ──project──▶  Submission     (Submission = LATEST)
```

New `examples/formkit_submissions/submission_stream/` app. The event log is
rakaia's durable `DjangoStreamStore` (`StreamEvent`/`StreamEntry`);
`Submission`/`SubmissionHistory` are projections rebuilt from it. A "save" is
`record_submission` — **append a `SubmissionEvent` and reproject in one
`transaction.atomic()`**. Provenance rides `append()` (not a `post_save` signal,
which would miss bulk writes).

## Demo assertions (`manage.py demo_submission_stream`)

- **[1] APPEND→PROJECT + REPLAY** — resolves to the latest event; rebuild reproduces it.
- **[2] HISTORY == the log** — `history_effects` fan-out; marker/actor/**url** from the envelope.
- **[3] SELF-HEALING** — a direct write to the projection is overwritten by reprojection (durable state *is* the log; this is why Decision #10's guard is belt-and-suspenders here).
- **[4] MODE B** — a context-less import still logs a full event; only actor/url null (pghistory-equivalent).
- **[5] TOMBSTONE** — a `delete` removes the row but keeps the `+`/`~`/`-` trail in history.
- **[6] DURABILITY** (`--reproject-only`) — a second process rebuilds state from the persisted log.

## Tests (`manage.py test submission_stream`, 7 tests)

latest-wins across interleaving · **append+project atomicity (Decision #11 rollback)** · tombstone projection · reproject/history idempotency · mode-B null actor · provenance-on-append.

## Scope / caveats

- Runs on **sqlite**, so the Postgres coverage guard (Decision #10) is out of scope by design — this is the topology where it's *not* load-bearing.
- `reproject_all` does a full rebuild for clarity; real code would project incrementally.
- Example-only; outside the `src/`+`tests/` CI-lint surface. `ruff check` clean locally.


joshbrooks:
Keeping this as a **reference spike** (converted to draft) — it proves Decision #13 end-to-end but is not itself for merge. The reusable primitives it exercises by hand (`reproject_all` → `project_latest`, and the `history_effects`+executor dance → `materialize_history`) are being promoted to core in a separate PR; this branch stays as the worked example.

Issue #1, opened 2026-03-13
Django integration


joshbrooks:
I want to plan the "Django Integration" part of this durable streams implementation
 - A decorator on a model will turn a Django model into a "streamed" model
 - A "streamed" model will on "save" or "delete" update a stream table, with a JSON field, with some serialization method
 - We'll stick to plain Python and Django as much as possible (no pydantic at this stage)
 - We'll use TDD

Our canonical example will be:
 - A User model
 - An Areas model
 - A Projects model
 - User has m2m to Area
 - Area has FK to Project
 - User subscribes to "Projects in an Area" as a stream

joshbrooks:
Reviewed against what's shipped — everything this issue planned is delivered and under test:

| Planned | Where it landed |
|---|---|
| Decorator turns a model into a "streamed" model | `@stream_model` — `src/django_rakaia/decorators.py` |
| save/delete updates a stream table with a JSON field + serialization | `create_stream_event` + `post_save`/`post_delete` signals; JSON produced via `to_dataclass` |
| Plain Python + Django, no pydantic | no pydantic in `src/` or dependencies |
| TDD | `tests/test_django_rakaia/` (`test_decorators.py`, `test_channels.py`, `test_protocol.py`, …) |
| Canonical User / Area / Project example | `tests/test_django_rakaia/models.py` — `Area`, `Project`, User FKs |
| Subscribe to "Projects in an Area" as a stream | `area:{id}:projects` stream; Projects fan out to **both** the area stream and the creator's `user:{id}:projects` stream, demonstrating multi-stream events |

Walkthrough docs: `docs/django-integration.md`.

**One deviation from the original plan:** it proposed "User has m2m to Area." The implementation instead links them through `Project` (Project → Area FK, Project → User FK) rather than a direct User↔Area m2m. The "Projects in an Area" subscription semantics are implemented exactly as specified; only the intermediate relationship shape differs.

Follow-on work since this issue (versioned handlers/upcasters/replay in #3, and runnable chat/polyglot/orders examples) built on top of this foundation.

PR #33, opened 2026-07-20
Subscriber cursors: incremental per-consumer stream reads (quick win #4, #11)

joshbrooks:
Draft — the rakaia-side of **quick win #4** from #11 (Partisipa streams migration). Autonomous session; **not** for merge without review.

## Why
Most of #11's "unlock" already shipped (event envelope, history read-model + `recover_peak_snapshot`, the pghistory-retirement design doc; quick wins #1–#3 map to shipped primitives/examples). The one genuinely-unbuilt rakaia-side piece was **#4**: Partisipa's `sync_identity/watermarks.py` + `list_options(last_change_id)` hand-roll a per-consumer watermark + DB-rewind detection. Once `Submission` is a stream, that machinery *is* a subscriber cursor over stream offsets. This ships that primitive.

## What
**Core** — `rakaia.subscription.poll(store, path, cursor) -> Poll` (store-agnostic, pure over the store):
- `status ∈ {fresh, advanced, caught_up, rewound, absent}`; `Poll.messages` + `Poll.cursor` (new watermark).
- **Rewind detection**: cursor sorts past the head (truncation / delete+recreate) → re-read from start. Offsets compared **numerically** (parse the `_`-joined int parts), so it's correct even for the Django store's non-zero-padded offsets — `"10"` is later than `"2"` despite sorting before it lexicographically (a test pins exactly this).
- `poll` never advances; caller applies **then** commits → **at-least-once**.

**Durable (Django)** — `ConsumerCursor` model (`(consumer_id, stream_path) → offset`, migration `0003`) + `django_rakaia.subscription`: `load_cursor` / `commit_cursor` / `poll_consumer`.

**Docs** — `docs/subscriber-cursors.md` (+ nav/index), incl. the Partisipa `last_change_id` mapping.

## Tests (TDD)
7 core + 4 durable (`tests/test_rakaia/test_subscription.py`, `tests/test_django_rakaia/test_subscription.py`) — fresh/advance/caught-up/rewound/absent, no-gaps-no-dupes, resume-across-restart, and the unpadded-offset rewind case. **515 passed**, ruff check + format clean, migration graph clean.

## Known limitation (documented)
A stream deleted and re-grown *past* the old cursor before the next poll reads as a normal `advanced` — offsets alone can't see it; a stream-generation token would close it (future extension).

## Scope
rakaia repo only (primitive + docs + tests). Wiring the client sync protocol (frontend IndexedDB) onto this stays the Partisipa-side follow-on, as #11 notes.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -124,6 +124,8 @@ the channel layer.
   label, and no-op suppression on append.
 - [History read-model](history-read-model.md) — Latest-state vs a queryable
   audit trail, both derived from one log.
+- [Subscriber cursors](subscriber-cursors.md) — Incremental per-consumer reads
+  over a stream's offsets, with rewind detection and durable watermarks.
 - [Alerts projection](alerts-projection.md) — Human judgment and machine rules
   in one projection, without clobber.
 - [Dry-run & executors](dry-run-and-executors.md) — Preview a replay's writes
--- a/docs/subscriber-cursors.md
+++ b/docs/subscriber-cursors.md
@@ -52,16 +52,26 @@ already is).
 
 ## Rewind detection
 
-If the stored cursor sorts *after* the current head, the log shrank beneath it —
-a truncation, or a delete-and-recreate — so the consumer must reset and re-read
-from the start. Offsets are compared **numerically** (parsing the `_`-joined
-integer parts), so detection is correct even for the Django store's
-non-zero-padded offsets, where `"10"` is later than `"2"` despite sorting before
-it as a string.
+If the stored cursor sorts *after* the current head, the log shrank beneath it,
+so the consumer resets and re-reads from the start. Offsets are compared
+**numerically** (parsing the `_`-joined integer parts), so detection is correct
+even for the Django store's non-zero-padded offsets, where `"10"` is later than
+`"2"` despite sorting before it as a string.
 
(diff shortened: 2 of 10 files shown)


joshbrooks:
**Adversarial self-review ran** (one finder, focused on offset/rewind edges). One real defect found and addressed:

- **delete+recreate offset collision (data loss):** if a stream is deleted and recreated so its new head lands on the *exact* offset a consumer last committed, `poll` returned `caught_up` and silently dropped the new content — a hole in at-least-once. Only across a delete+recreate; **append-only streams (the intended use — Submission events are appended) are correct.**

Since the proper fix (a per-stream **generation token**, e.g. the Django `Stream` pk, threaded through the cursor) is a multi-file core change touching both stores + the protocol + the cursor shape, I did **not** rush it in while unattended. Instead this commit:
- corrects the docstring/docs (the limitation was understated as benign) and scopes the guarantee to append-only;
- adds a **strict-xfail** regression test pinning the desired `rewound`+redeliver behavior, so it auto-flips green when the token lands.

Everything else in the review verified safe (numeric offset compare, caught_up/advanced boundary, absent-vs-rewound window, commit concurrency). **Follow-up before this leaves draft:** implement the generation token. Recommend deciding the token's home (protocol method vs. passed-in) together.

PR #42, opened 2026-07-21
feat(upcaster): content-routing via match_field (#37)

### LINKED ISSUE #37: register_upcaster: support match_field (content routing), mirroring register_handler
**Problem.** `register_handler` supports `match_field` (route on an event-body field, e.g. `form_type`), but `register_upcaster` does **not** — upcasters are only matched against the stream-path/`event_match` string (`fnmatch(event_match_str, pattern)`).

**Why it matters.** On per-submission streams (`submissions:<uuid>`), the path carries no form type, so a **per-form upcaster cannot be routed at all** — it silently never fires. This is exactly what left a consumer's TF1321 quantize upcaster dormant. Handlers content-route; upcasters must too, or per-form schema evolution is structurally impossible on entity-keyed streams.

**Proposal.** Add `match_field` to `register_upcaster` / `UpcasterRegistry`, mirroring the handler dispatch path (match the pattern against `event[match_field]` when set). Preserve current behavior when unset.

**Acceptance.** An upcaster registered with `match_field="form_type"` fires for matching events on a `submissions:<uuid>` stream; a test covers it; docs updated. Refs ADR-0002.

joshbrooks:
Adds `match_field` to `register_upcaster`, mirroring `register_handler`: the `event_match` pattern is tested against `event[match_field]` (e.g. `form_type`) instead of the stream/event-match string. This lets per-form upcasters route on a per-entity stream (`submissions:<uuid>`) whose path carries no type — previously structurally impossible, so such upcasters were silently dormant.

- Threaded through `UpcasterVersion`, `register()`, the decorator, both matching sites (`current_version`/`apply_chain` via a shared `_subject` helper), and the persisted identity.
- `replay`/`merge_replay` now resolve the target version **per event** (`upcast_to_current`) instead of once-per-stream from the path.
- +6 content-routing tests; full suite + ruff green.

See ADR-0002. Closes #37.

### DIFF
--- a/src/rakaia/registry.py
+++ b/src/rakaia/registry.py
@@ -582,7 +582,11 @@ class UpcasterRegistry:
             raise UpcasterConflictError(
                 f"Upcaster already registered for event_match={event_match!r}, "
                 f"from_version={from_version} (existing dotted_path="
-                f"{existing.dotted_path!r}, new={new.dotted_path!r})"
+                f"{existing.dotted_path!r} match_field={existing.match_field!r}, "
+                f"new dotted_path={new.dotted_path!r} match_field={new.match_field!r})."
+                " The (event_match, from_version) key ignores match_field, so a "
+                "stream-routed and a content-routed upcaster cannot share it — "
+                "give them distinct event_match patterns."
             )
 
         if self._store is not None:
@@ -595,11 +599,23 @@ class UpcasterRegistry:
         up: UpcasterVersion, event: dict[str, Any] | None, event_match_str: str
     ) -> str:
         """The string an upcaster's pattern is matched against: a content field
-        when `match_field` is set (mirrors handler routing), else the stream/
-        event-match string. A content-routed upcaster with no event to read
-        yields "" (so it simply does not match)."""
+        when `match_field` is set (mirrors `HandlerRegistry._match_subject`), else
+        the stream/event-match string.
+
+        A content-routed upcaster matched with `event=None` raises — like handler
+        dispatch — to catch a caller who forgot to pass the event (which would
+        otherwise leave the event silently un-upcast). An event that is present
+        but lacks the field yields "" and simply does not match (a different
+        form_type on the stream is normal, not an error)."""
         if up.match_field is not None:
-            return str((event or {}).get(up.match_field, ""))
+            if event is None:
+                raise ValueError(
+                    f"Upcaster (event_match={up.event_match!r}) routes on "
+                    f"match_field={up.match_field!r} but was matched without an "
+                    "event; pass the decoded event to current_version()/"
+                    "apply_chain() (or use upcast_to_current)."
+                )
+            return str(event.get(up.match_field, ""))
         return event_match_str
 
     def current_version(
@@ -634,6 +650,13 @@ class UpcasterRegistry:
         (default 1). Each step finds the upcaster matching the event_match (or
         `event[match_field]` for content-routed upcasters) and the current
         version, then bumps schema_version on the result.
+
+        Content-routing note: each step is re-matched against the *progressively
+        upcasted* event, so a content-routed chain's `match_field` discriminator
+        must stay stable across the chain. An upcaster that renames or rewrites
+        the discriminator (e.g. v1→v2 renaming `form_type`) re-routes subsequent
+        steps off the new value — keep the discriminator field constant, or route
+        those steps on the stream path.
 
         If `drift_callback` is provided, it is called once per step with
         `(version, current_hash)` if the upcaster's source hash has changed
--- a/src/rakaia/registry.py
+++ b/src/rakaia/registry.py
@@ -519,6 +519,12 @@ class UpcasterVersion:
     dotted_path: str
     source_hash: str
 
+    match_field: str | None = None
+    """When set, the ``event_match`` pattern is tested against
+    ``event[match_field]`` (content routing) instead of the stream/event-match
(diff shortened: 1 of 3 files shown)


src/rakaia/registry.py
```diff
@@ -576,12 +590,30 @@ def register(
...
-    def current_version(self, event_match_str: str) -> int:
-        """Highest schema version reachable for an event matching this string."""
+    @staticmethod
+    def _subject(
+        up: UpcasterVersion, event: dict[str, Any] | None, event_match_str: str
+    ) -> str:
+        """The string an upcaster's pattern is matched against: a content field
+        when `match_field` is set (mirrors handler routing), else the stream/
+        event-match string. A content-routed upcaster with no event to read
+        yields "" (so it simply does not match)."""
```
joshbrooks: **Silent no-op vs. explicit raise — inconsistent with `HandlerRegistry._match_subject`.** When `match_field` is set but `event` is `None` (or lacks the field), `_subject` returns `""`, which silently fails to match any registered pattern (short of `""` or `*`). The existing, analogous `HandlerRegistry._match_subject` raises `ValueError` in this exact situation ("routes on match_field=... but resolve() was called without an event"), specifically to catch a caller who forgot to pass `event`.

The PR frames this as "mirrors register_handler," but the one behavior that would catch caller misuse — the loud failure — was dropped for a silent one here. Any call site that invokes `current_version`/`apply_chain` without `event` for a content-routed upcaster will silently skip the upcast (event stuck on an old schema) instead of erroring. Worth raising here too, or documenting why upcasting should be forgiving where handler dispatch is not.

None of the 6 new `TestContentRouting` tests exercise the `event is None` case.

src/rakaia/registry.py
```diff
@@ -558,6 +571,7 @@ def register(
             fn=fn,
             dotted_path=f"{fn.__module__}.{fn.__qualname__}",
             source_hash=hash_function_source(fn),
+            match_field=match_field,
```
joshbrooks, line 574: The `_upcasters` dict key is still `(event_match, from_version)`, unaffected by `match_field`. A stream-routed upcaster (`match_field=None`) and a content-routed one (`match_field="form_type"`) sharing the same `event_match`/`from_version` collide on this key. Since `_u_identity` now includes `match_field`, they aren't identical, so this correctly raises `UpcasterConflictError` rather than silently overwriting — but the conflict error message (below) doesn't mention `match_field` as the actual differentiator, which will confuse anyone converting a stream-routed upcaster to content-routed under the same identifiers. Consider naming `match_field` explicitly in that message.

src/rakaia/registry.py
```diff
@@ -616,7 +649,10 @@ def apply_chain(
             matching = [
                 up
                 for (pattern, fv), up in self._upcasters.items()
-                if fv == current and fnmatch.fnmatchcase(event_match_str, pattern)
+                if fv == current
+                and fnmatch.fnmatchcase(
+                    self._subject(up, event, event_match_str), pattern
```
joshbrooks, line 677: `_subject` is recomputed each loop iteration against `event` as progressively upcast by prior steps (`event = step.fn(event)` a few lines below). If an upcaster mutates or renames the very field used for routing (e.g. a v1→v2 step that renames `form_type` to `entity_type`, or rewrites its value), the *next* step's routing decision is based on the already-upcasted value, not the original — silently breaking the chain for a "rename the discriminator on migrate" pattern, which is exactly the kind of thing the PR's own `form_type` example invites. No test in `TestContentRouting` exercises a multi-step chain to confirm this is safe; all six use a single from_version=1 step.

tests/test_rakaia/test_upcasters.py
```diff
@@ -189,6 +189,70 @@ def my_upcaster(event):
...
+    instead of the stream/event-match string — so per-form upcasters work on a
+    per-entity stream (submissions:<uuid>) whose path carries no type."""
+
+    def test_current_version_matches_on_content_field(self, reg: UpcasterRegistry):
+        reg.register("TF_13_2_1", 1, _tf_v1_to_v2, match_field="form_type")
+        # path string alone can't resolve a content-routed upcaster
+        assert reg.current_version("submissions:abc") == 1
+        # with the event, it routes on form_type
+        assert reg.current_version("submissions:abc", {"form_type": "TF_13_2_1"}) == 2
+        assert reg.current_version("submissions:abc", {"form_type": "SF_1_1"}) == 1
```
joshbrooks, line 215: Coverage gap: none of `TestContentRouting`'s tests cover (1) `event=None`/missing-field routing (the silent-`""` behavior flagged on registry.py), (2) two upcasters that structurally both match the same event (ambiguous match — `apply_chain` raises `UpcasterChainError`, but `current_version` doesn't detect the ambiguity when computing target, so the failure surfaces later than it could), (3) registering a content-routed upcaster under the same `(event_match, from_version)` as an existing stream-routed one, or (4) converting a previously-persisted stream-routed registration to content-routed on restart (identity mismatch → `UpcasterConflictError`, a migration foot-gun). These are the edge cases most likely to bite in production for a PR specifically about routing correctness.

joshbrooks (commented):
Reviewed the content-routing (match_field) feature for UpcasterRegistry. The core mechanism is sound and the happy path is well tested, but it diverges from the existing HandlerRegistry content-routing precedent in an important way (silent no-match vs. explicit raise when event is missing), and a few registration/chain edge cases aren't covered by the new tests.

PR #43, opened 2026-07-21
feat(effects): validate exclude/spare_keys; add dispatch_external seam (#40)

### LINKED ISSUE #40: Effect model: exclude vs spare_keys collision; external-effect dispatch seam
Two `Effect` model rough edges:

1. **`exclude` vs `spare_keys` can silently AND.** Both spare rows from a delete/retire — `exclude` is a flat `.exclude(**exclude)`, `spare_keys` is composite `.exclude(Q(..)|Q(..))`. If a handler supplies both, `DjangoExecutor` applies **both** exclusions conjunctively (`.exclude(**exclude).exclude(Q|..)`), which is almost never intended and is unvalidated. **Fix:** validate mutual-exclusion (raise if both set), or collapse `exclude` into the `spare_keys` form.

2. **`external` effects have no dispatch seam.** `op="external"` carries opaque `kind`/`payload`; rakaia only *filters* them on replay. There's no registry/dispatcher, so every consumer re-implements "route `kind` → side-effect," and machine-reconcile resolutions can't emit per-row transitions (the R2 gap seen in the alerts consumer). **Fix:** offer an optional external-effect dispatch seam (map `kind` → handler) and/or an executor-level retire-diff so a reconcile that flips N rows can emit N transitions.

**Acceptance.** Mixing `exclude`+`spare_keys` is validated or unified; a documented pattern (or seam) exists for routing external effects. Refs ADR-0002.

joshbrooks:
Two Effect-model hygiene fixes from the seams review:

1. `Effect.__post_init__` rejects (a) setting both `exclude` and `spare_keys` (alternative row-sparing mechanisms the executor would silently AND), and (b) `exclude` on a non-`delete` op (the retire path ignored it). Fails fast in the handler.
2. `dispatch_external(effects, {kind: fn})` — a documented, exported seam for routing `op="external"` effects to per-`kind` handlers, so consumers stop re-implementing the `kind` switch.

+9 tests; full suite + ruff green. See ADR-0002. Closes #40.

### DIFF
--- a/src/rakaia/effects.py
+++ b/src/rakaia/effects.py
@@ -10,7 +10,7 @@ state) and keeps handlers trivially testable without a database.
 from __future__ import annotations
 
 import json
-from collections.abc import Iterable
+from collections.abc import Callable, Iterable, Mapping
 from dataclasses import dataclass
 from typing import Any, Literal, Protocol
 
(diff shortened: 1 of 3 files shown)


src/rakaia/effects.py
```diff
@@ -137,3 +157,41 @@ def check_disjoint_defaults(effects: Iterable[Effect]) -> None:
...
+    Pass a ``{kind: fn}`` map; each external effect is routed to
+    ``handlers[effect.kind]`` (non-external effects are skipped). Returns the
+    number dispatched. An effect whose ``kind`` has no handler raises ``KeyError``
+    by default, or is skipped with ``on_unknown="ignore"``.
+    """
+    dispatched = 0
+    for eff in effects:
+        if eff.op != "external":
+            continue
+        handler = handlers.get(eff.kind) if eff.kind is not None else None
```
joshbrooks, line 190: Coverage gap: `Effect(op="external", kind=None, ...)` is never exercised. Manually verified behavior is correct (falls through to the `handler is None` branch and raises `KeyError: 'No handler for external effect kind=None'` under the default `on_unknown="raise"`, or is silently skipped under `on_unknown="ignore"`), but there's no regression test pinning this — a future refactor of this ternary (e.g. `handlers.get(eff.kind, default)`) could silently change that behavior. Consider adding a `test_kind_none_treated_as_unknown` case alongside `TestDispatchExternal`.

joshbrooks (commented):
Reviewed the exclude/spare_keys validation and dispatch_external seam. Verified against all existing call sites (reconcile_children, reconcile_by_key, DjangoExecutor) — no bypass paths and no call site is broken by the new __post_init__ checks. Only finding is a minor test-coverage gap; no correctness bugs found.

Issue #41, opened 2026-07-21
Architecture: make the framework ↔ protocol-server boundary explicit (ADR-0002)


joshbrooks:
**Epic / tracking issue for ADR-0002** (`docs/adr/0002-framework-vs-protocol-server-boundary.md`).

rakaia is effectively **two products in one package**: a *Durable Streams protocol server* (`handler.py` ASGI, producer epoch/seq fencing, CDN cursors, SSE, the full `StreamStore` lifecycle) and an *event-sourcing / projection framework* (`Effect`/registry/`replay`/projections, `ReadableStore`/`ProjectionReader`). The framework tier is what consumers build on and is well-designed, but its extension seams are *under-formalized* relative to how polished the protocol tier is, and nearly every rough edge lives at the boundary where the two fuse (envelope bolted onto the transport message; two timestamps named "ts"; two "cursor" concepts; a store interface that is really two interfaces).

**Goal:** make the boundary explicit and bring the framework-tier seams up to the protocol tier's standard.

**Child work items**
- [ ] #36 — write-side Store contract + conformance suite
- [ ] #37 — `register_upcaster` `match_field`
- [ ] #38 — injectable/resettable registries
- [ ] #39 — replay determinism traps
- [ ] #40 — Effect model cleanups
- [ ] Docs: package-boundary doc + "what needs Django" matrix (ADR-0002 consequence)
- [ ] Export the extension protocols (`ReadableStore`, `ProjectionReader`) in `rakaia.__all__` (also in #36)

**Related, needs verification (not yet a confident bug):** the standalone ASGI server (`handler.py`) and the Django `protocol_views.py` reportedly emit **different offset formats** for the same protocol — confirm and, if real, file as a bug.

Context: this came out of an architecture/seams review; see ADR-0002.

joshbrooks:
**Consumer note (Partisipa `rebuild_ida`) — a concrete boundary instance for the "what needs Django" matrix: `apps.ready()` hard-imports channels.**

`django_rakaia/apps.py:10` runs `import django_rakaia.channels_signals` unconditionally in `AppConfig.ready()`. So a consumer that uses **only the framework tier** (durable `StreamStore` + `Effect`/registry/`replay`/projections) and none of the protocol-server features (SSE/ASGI/consumers) is still forced to have `channels` installed — importing `django_rakaia` at app-load fails with `ModuleNotFoundError: channels` otherwise.

[...]

Issue #36, opened 2026-07-21
Formalize the write-side Store contract + shared conformance suite


joshbrooks:
**Problem.** `protocols.py` defines only `ReadableStore` (read side). The **write/lifecycle side has no protocol** — `create`/`append`/`has`/`get_current_offset`/`close` are duck-typed on the concrete `StreamStore`, and `RAKAIA_STORE="memory"|"durable"` swaps implementations by string with no interface check. `ReadableStore` isn't even exported in `rakaia.__all__`.

**Why it matters.** The standard pattern is *test on the in-memory store, ship on `DjangoStreamStore`* — and they diverge silently:
- JSON **array/batch** appends flatten to N events in-memory but stay 1 event in Django (single-dict payloads are unaffected);
- offset format differs (compound `{seq}_{byte}` vs `BigInteger`);
- timestamp precision differs (`time.time()` vs `DateTimeField`) — and `merge_replay` orders on `ts`;
- `DjangoStreamStore` implements **no producer dedup / close / TTL / long-poll**. (A consumer had to add a `pgh_id` watermark for backfill idempotence precisely because the durable store won't dedup, and nothing documents that.)

**Proposal.**
- Add a `WritableStore`/`AppendableStore` protocol (`create`, `append`, `has`, `read`, `current_offset`), and **export both store protocols** in `__all__`.
- Ship one **shared conformance test suite** that every store (in-memory, Django, third-party) must pass — covering offset semantics, JSON mode, timestamp contract, and the append-requires-existing-stream behavior.
- Document explicitly which lifecycle methods a *framework* store must implement vs which belong only to the *protocol server* store.

**Acceptance.** Protocols exported; conformance suite green for both built-in stores; a divergence (e.g. array-flatten) either unified or documented as an explicit, tested difference. Refs ADR-0002.

PR #44, opened 2026-07-21
docs(adr): ADR-0002 — framework vs protocol-server boundary

joshbrooks:
Records the two-tier thesis from the architecture/seams review: rakaia is a Durable Streams *protocol server* fused with an event-sourcing *projection framework*, and the framework-tier extension seams are under-formalized relative to the protocol tier. Decision: name the boundary and bring the framework seams (store-write protocol + conformance, upcaster `match_field`, injectable registries, determinism contract, Effect hygiene) up to standard.

Tactical work tracked in #36–#40 under epic #41. Refs #41.

### DIFF
--- a/docs/adr/0002-framework-vs-protocol-server-boundary.md
+++ b/docs/adr/0002-framework-vs-protocol-server-boundary.md
@@ -1,0 +1,131 @@
+# ADR 0002 — Name the framework ↔ protocol-server boundary; formalize the framework seams
+
+- **Status:** Proposed
+- **Date:** 2026-07-21
+- **Deciders:** rakaia maintainers
+- **Related:** [`django-integration.md`](../django-integration.md),
+  [`dry-run-and-executors.md`](../dry-run-and-executors.md),
+  [`staged-replay.md`](../staged-replay.md), [`protocol.md`](../protocol.md),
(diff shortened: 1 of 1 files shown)


joshbrooks (commented):
Reviewed ADR-0002. This is a docs-only PR (1 file, no code changes) — test coverage is N/A. I cross-checked the concrete technical claims against `main` (StreamMessage's `label`/`metadata` fields and the exact "ignored by the transport" wording, `merge_replay`'s `order_key="ts"` default and its "envelope timestamp" docstring phrasing vs. the actual envelope timestamp living on `StreamMessage.timestamp`, `protocols.py`'s `ReadableStore`/`ProjectionReader` and the absence of `WritableStore`/export in `__all__`, `DjangoStreamStore`'s method set, `RAKAIA_STORE`'s string-based dispatch with no conformance check, the `register_upcaster`/`register_handler` `match_field` asymmetry, and the `exclude`/`spare_keys` composability gap in `effect_executor.py`). All of them hold up — every cited seam is real, and the linked docs (django-integration.md, dry-run-and-executors.md, staged-replay.md, protocol.md, ADR 0001) all exist and are consistent in format with this ADR. Nothing actionable to flag.

Issue #39, opened 2026-07-21
Close replay determinism traps: dual 'ts', upcaster-rewrites-history, offset format


joshbrooks:
Replay's headline promise is deterministic, idempotent projection. Three traps undermine it:

1. **Dual "ts".** `StreamMessage.timestamp` is the transport time (`time.time()` at append), but `merge_replay`'s default `order_key="ts"` reads the **payload body**, and its docstring calls that "the envelope timestamp." Two different timestamps, one name — a consumer hit this directly (their payload had no `ts`, so merge ordering / soft-retire silently broke). **Fix:** make the envelope timestamp first-class and have `merge_replay` read it, or rename/clarify the payload order key; align the docstring with the code.

2. **Upcaster edits rewrite history retroactively.** Because upcasters aren't versioned by event range, editing an upcaster changes the effective shape of *all* historical events at that schema step. Source-hash drift *detects* a change but the semantics are undocumented. **Fix:** document the contract explicitly; consider requiring an explicit version bump.

3. **Offset format divergence across stores** (compound `{seq}_{byte}` in-memory vs `BigInteger` in Django) forces consumers to `int(m.offset)` and makes cursors non-portable. **Fix:** one canonical offset type/contract across stores (see #A2).

**Acceptance.** `merge_replay` timestamp source is unambiguous and documented; upcaster-rewrites-history contract documented; offset contract unified. Refs ADR-0002.

joshbrooks:
**Consumer note (Partisipa `rebuild_ida`) — the `ts` fix must keep the envelope timestamp producer-settable to a *logical* time, not transport time.**

We're the consumer that hit #39.1: our appended payloads originally carried no `ts`, so `merge_replay`'s `order_key="ts"` (which reads the payload body) silently mis-ordered and broke soft-retire. We worked around it by stamping `ts` into the payload.

The caveat for whatever fix lands: **don't resolve this by having `merge_replay` read `StreamMessage.timestamp` (the append/transport time).** Our two producers set `ts` to *different logical times on purpose*:

- **live save** → `ts = now()` (append time ≈ event time — fine)
- **one-time backfill from pghistory** → `ts = the original historical event time` (`extract(epoch from pgh_created_at)`), **not** the append time.

[...]

PR #45, opened 2026-07-21
feat(store): WritableStore protocol + shared conformance suite (#36)

### LINKED ISSUE #36: Formalize the write-side Store contract + shared conformance suite

joshbrooks:
Formalizes the framework **store seam** (ADR-0002).

- `protocols.py` gains **`WritableStore`** (= `ReadableStore` + `has`/`create`/`append`) — the create/append/read/has surface producers and the meta-stream registry rely on. It deliberately **excludes** the Durable Streams protocol-server lifecycle (producer epoch/seq fencing, close, TTL, `wait_for_messages`), which is not required to back `replay()`/projections.
- **`ReadableStore`, `WritableStore`, `ProjectionReader` are now exported** from `rakaia`, so consumers writing a custom backend/reader have something to import and type against (previously they were unexported).
- **`tests/store_contract.py`** — a shared `StoreContract` run against **both** the in-memory `StreamStore` **and** the durable `DjangoStreamStore` (22 cases). This makes 'test on memory, ship on durable' *enforced*, not assumed: create-idempotence, append-requires-stream, ordered round-trip, partial read from offset, envelope round-trip (labelled + raw), float timestamp, current-offset advance, and `isinstance` protocol satisfaction.

Return types stay loose (`Stream`/`AppendResult` vs `StreamEntry`); the contract is the read-back behaviour. Known cross-store differences (offset format, JSON-array flattening) are out of scope and tracked in #39.

542 passed; ruff clean. Closes #36.

### DIFF
--- a/src/rakaia/protocols.py
+++ b/src/rakaia/protocols.py
@@ -5,9 +5,14 @@ extension seams of the event-sourcing framework (ADR 0002).
 `replay()` only ever *reads* a stream, so it depends on the narrow
 `ReadableStore` protocol rather than the concrete in-memory `StreamStore`.
 Producers and the meta-stream registry additionally *write*, captured by
-`WritableStore`. Staged handlers read committed projections through
+`WritableStore`. Subscribers additionally need the head offset, captured by
+`CursorStore`. Staged handlers read committed projections through
 `ProjectionReader`. Any backend satisfying the relevant protocol — the in-memory
 `StreamStore`, a durable DB-backed store, or a third-party one — plugs in.
+
+All store-facing protocols live here so they read as one coherent seam:
+`ReadableStore` (read), `WritableStore` (+ create/append/has), `CursorStore`
+(+ get_current_offset).
 """
 
 from __future__ import annotations
@@ -54,22 +59,41 @@ class WritableStore(ReadableStore, Protocol):
         """Whether a stream exists at `path`."""
         ...
 
-    def create(self, path: str, **kwargs: Any) -> Any:
+    def create(self, path: str) -> Any:
         """Idempotently create the stream at `path`. `append` requires it to exist.
 
-        Backends may accept extra keyword options (content type, TTL, …) or
-        ignore them; only idempotent creation is part of the contract.
+        Only idempotent creation by `path` is part of the contract. Concrete
+        stores may accept extra keyword-only options (content type, TTL, …) via
+        their own richer signatures — deliberately *not* on this protocol, so a
+        backend that ignores such an option (e.g. `DjangoStreamStore`) can't be
+        called with it through the `WritableStore` type and silently no-op.
         """
         ...
 
     def append(self, path: str, data: bytes, options: Any = None) -> Any:
         """Append one (optionally enveloped) event to `path`.
 
-        Raises `KeyError` if the stream does not exist (create it first). The
-        envelope — `label` and `metadata` on `options` (an `AppendOptions`) — is
-        recorded and read back by `read`; ambient `provenance()` merges under
+        Raises `KeyError` if the stream does not exist (create it first) — the one
+        guaranteed exception. Beyond that, backends may raise for their own
+        validation (the in-memory `StreamStore` raises `ValueError` on a
+        content-type or Stream-Seq conflict) or signal a closed stream without
+        raising; `DjangoStreamStore` has no such concept. Depend only on the
+        `KeyError` here; treat richer errors as backend-specific.
+
+        The envelope — `label` and `metadata` on `options` (an `AppendOptions`) —
+        is recorded and read back by `read`; ambient `provenance()` merges under
         explicit metadata.
         """
+        ...
+
+
+@runtime_checkable
+class CursorStore(ReadableStore, Protocol):
+    """A `ReadableStore` that also exposes its current head offset — what a
+    subscriber (`poll()`) needs to detect new messages and a rewound log."""
+
+    def get_current_offset(self, path: str) -> str | None:
+        """The offset after the last message, or None if the stream is absent."""
         ...
 
 
--- a/src/rakaia/protocols.py
+++ b/src/rakaia/protocols.py
@@ -1,10 +1,13 @@
 """
-Structural protocols for pluggable storage backends.
+Structural protocols for pluggable storage and projection backends — the
+extension seams of the event-sourcing framework (ADR 0002).
 
 `replay()` only ever *reads* a stream, so it depends on the narrow
-`ReadableStore` protocol rather than the concrete in-memory `StreamStore`. Any
-backend that can return a stream's messages in order — the in-memory
-`StreamStore` or a durable, DB-backed store — satisfies it.
+`ReadableStore` protocol rather than the concrete in-memory `StreamStore`.
+Producers and the meta-stream registry additionally *write*, captured by
+`WritableStore`. Staged handlers read committed projections through
+`ProjectionReader`. Any backend satisfying the relevant protocol — the in-memory
+`StreamStore`, a durable DB-backed store, or a third-party one — plugs in.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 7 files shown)


tests/store_contract.py
```diff
@@ -0,0 +1,107 @@
...
+        store.create("s")
+        store.create("s")
+        assert store.has("s") is True
+
+    def test_has_reflects_existence(self, store):
+        assert store.has("s") is False
+        store.create("s")
+        assert store.has("s") is True
+
+    def test_append_requires_existing_stream(self, store):
```
joshbrooks, line 58: Coverage gap: the shared contract only asserts `KeyError` for append-to-missing/read-of-missing. It never exercises `create()`'s conflicting-config rejection (`StreamStore.create` raises `ValueError` on a re-create with different `content_type`/`ttl_seconds`/etc., per `src/rakaia/store.py:144-146`), nor the closed-stream/`ValueError`(seq-conflict, content-type mismatch) paths that `StreamStore.append` documents. `DjangoStreamStore.create` silently ignores all kwargs and has no closed/seq concept at all (`src/django_rakaia/django_store.py:51-55`). A producer that relies on in-memory's conflict-rejection behavior would pass this suite today and then silently no-op against the durable backend — exactly the 'test on memory, ship on durable' gap this PR's contract is meant to close. Worth adding a `test_create_conflicting_config_rejected_or_ignored` case (or explicitly documenting that create-conflict detection is out of contract scope, since Django doesn't implement it).

src/rakaia/protocols.py
```diff
@@ -29,6 +32,47 @@ def read(
...
+
+    def create(self, path: str, **kwargs: Any) -> Any:
+        """Idempotently create the stream at `path`. `append` requires it to exist.
+
+        Backends may accept extra keyword options (content type, TTL, …) or
+        ignore them; only idempotent creation is part of the contract.
+        """
+        ...
+
+    def append(self, path: str, data: bytes, options: Any = None) -> Any:
```
joshbrooks, line 73: `WritableStore.append`'s docstring promises only `KeyError` on a missing stream, but the in-memory `StreamStore.append` also raises `ValueError` (content-type mismatch, Stream-Seq conflict — `src/rakaia/store.py:253,277`) and can return `AppendResult(stream_closed=True)` without raising at all for a closed stream (`store.py:246`). A caller coded strictly against this protocol docstring won't know those branches exist, and `DjangoStreamStore.append` never raises `ValueError` or short-circuits on 'closed' since it has no such concept — so the exception contract silently diverges between backends. Consider either documenting the divergence explicitly or scoping the protocol docstring to say 'at least KeyError; backends may raise ValueError for backend-specific validation.'

src/rakaia/protocols.py
```diff
@@ -29,6 +32,47 @@ def read(
...
+        """Append one (optionally enveloped) event to `path`.
+
+        Raises `KeyError` if the stream does not exist (create it first). The
+        envelope — `label` and `metadata` on `options` (an `AppendOptions`) — is
+        recorded and read back by `read`; ambient `provenance()` merges under
+        explicit metadata.
+        """
```
joshbrooks, line 86: `create(self, path: str, **kwargs: Any) -> Any` accepts any keyword at all, so a typo'd kwarg (e.g. `inital_data=`) or a kwarg meaningful only to one backend (e.g. `ttl_seconds=` passed to `DjangoStreamStore.create`, which silently swallows all kwargs per `django_store.py:51-53`) passes type-checking on both backends with no error — a caller can believe TTL is configured when it silently isn't. Since the docstring already scopes out TTL/content-type/etc. as backend-specific, a stricter protocol signature (`def create(self, path: str) -> Any`, letting each concrete class carry its own richer keyword-only params via structural narrowing) would make that scoping decision enforced by the type system rather than only documented in prose.

tests/store_contract.py
```diff
@@ -0,0 +1,107 @@
...
+            return MyStore()
+    """
+
+    def test_satisfies_writable_store_protocol(self, store):
+        from rakaia import ReadableStore, WritableStore
+
+        assert isinstance(store, WritableStore)
+        assert isinstance(store, ReadableStore)
+
+    def test_create_is_idempotent(self, store):
```
joshbrooks, line 48: Reuse: this contract (create-idempotence, append-requires-stream, envelope round-trip, offset/timestamp checks) substantially duplicates tests already present in `tests/test_django_rakaia/test_django_store.py` (e.g. `test_create_is_idempotent`, `test_append_requires_existing_stream`, `test_read_from_offset_partial`, `test_get_current_offset`). Now the same behavior is asserted in two places per backend — a future change to append/read semantics needs updating both, and if only one is touched the suites can silently diverge, undermining the 'shared conformance suite' guarantee this PR is introducing. Worth trimming the now-redundant cases from `test_django_store.py` (and the in-memory `test_store.py` equivalent) to leave only backend-specific tests (row locking, durability-across-instances) there, with `StoreContract` as the single source of truth for the shared surface.

src/rakaia/protocols.py
```diff
@@ -29,6 +32,47 @@ def read(
         ...
 
 
+@runtime_checkable
+class WritableStore(ReadableStore, Protocol):
```
joshbrooks, line 41: There's already a second narrow extension of `ReadableStore` in the codebase — `CursorStore(ReadableStore, Protocol)` in `src/rakaia/subscription.py:49`, adding `get_current_offset`. `WritableStore` and `CursorStore` are structurally-similar-but-independently-defined store-facing protocols living in different modules with no shared naming/discovery convention (and the new conformance suite doesn't check `CursorStore` conformance either, even though `get_current_offset` is exercised by `test_get_current_offset_none_then_advances`). A future protocol author extending `ReadableStore` a third way won't find this one from `protocols.py` alone. Consider consolidating store-facing protocols in `protocols.py` (or cross-referencing them in the module docstring) so `WritableStore`, `CursorStore`, and `ReadableStore` read as one coherent seam.

joshbrooks (commented):
Reviewed the WritableStore protocol addition and the new shared conformance suite (tests/store_contract.py + its two backend subclasses). The protocol/exports change is purely additive and safe — five independent finder angles (line-by-line scan, removed-behavior audit, cross-file tracer, cleanup, conventions) turned up no active bugs. The findings below are coverage gaps in the conformance suite and reuse/typing nits worth addressing before this becomes the load-bearing cross-backend contract.

PR #58, opened 2026-07-25
feat(alerts): machine-resolution transitions via executor retire-diff (#32)

### LINKED ISSUE #32: Alerts: external transitions for machine (reconcile) resolutions
**Narrowed from the original Phase-3 scope.** The alerts derived⊕authored composition **shipped in #31** (`TestComposedAlerts`, `docs/alerts-projection.md`): staged replay's `ProjectionReader` was the `refs` substrate, so no new core primitive was needed. Oracle criteria 2 (zero-clobber) and 3 (dismissals honored) are covered. This issue now tracks only the one remaining gap.

## The gap — machine-resolution notifications

`reconcile_by_key` emits **no** `external` transition on retire. One retire Effect covers N rows, and a pure handler can't see which rows actually flipped, so it can't emit "one transition per real machine resolution."

- **Authored** transitions already fire correctly (Phase 1): each `alert_raised`/`alert_dismissed` event is itself the transition, one per event, dropped on replay.
- **Machine** resolutions (a validator flag that stops violating → soft-deleted by the reconcile retire) are currently **silent**.

## Proposed fix

Have the executor's `retire` pass **return the rows it actually updated** (the keys whose `resolved_at` went NULL→set), so the replay orchestrator can emit one `external` `alert_transition` per real state change. This keeps handlers pure and needs no cross-stream read.

- Replay-safety: still skipped on replay (external effects are), so a rebuild never re-spams.
- Alternative considered: derive transitions from the reader (prev open-set vs new) — rejected, more coupling than the executor-diff.

**Not blocking** any shipped correctness; alerts are fully functional without it.

joshbrooks:
Closes #32.

## The gap

`reconcile_by_key` soft-deletes stale rows in a **single** retire Effect covering N rows, so a pure handler can't see which rows actually flipped — machine (validator) resolutions were **silent**, while authored `alert_raised`/`alert_dismissed` transitions fired correctly (Phase 1).

## The fix — executor retire-diff

The retire is already **open-guarded** (`resolved_at IS NULL`), so the rows its WHERE clause matches are *exactly* the ones flipping NULL→set. No diff needed — just their identities.

- Opt in with `reconcile_by_key(..., transition_kind="alert_transition")`.
- The executor `SELECT`s the flipped rows' natural-key identities **before** the `UPDATE` (same atomic txn, paid **only** when opted in) and returns them in the new `ApplyReport.retire_flips`.
- The replay orchestrator emits **one `external` transition per real resolution**, routed through the **same gate** as authored transitions.

```
executor.apply(effects) → ApplyReport
  _retire (only when transition_kind set):
    rows = qs.order_by(*key).values(*key_fields)   ← capture flip set (same txn)
    qs.update(**patch)                             ← same WHERE ⇒ rows == flips
_synth_transitions(report): per flipped key →
  Effect(op="external", kind="alert_transition",
         payload={"key": k, "state": "resolved", **patch})
_apply_effects(synth):
  include_external=False (replay/rebuild) → counted skipped, NOT delivered ✓ never re-spams
  include_external=True  (live)           → ONE transition per real resolution ✓
```

## Design notes

- **Identity = natural key** (`{stream_key, alert_type, field_key}`), matching authored transitions, via the executor folding scope-equality columns + reconcile key-fields — not pk. Costs one extra Effect field (`transition_key_fields`) to keep the executor model-agnostic.
- **`Executor.apply` return type `None → ApplyReport | None`** (breaking protocol change, one prod impl + a couple test doubles). `_synth_transitions` tolerates `None`, so existing capture executors were untouched.
- **Determinism:** the pre-`UPDATE` `SELECT` uses `order_by(*key_fields)`, so transitions stay a pure function of the events.

## Tests (TDD, red→green)

- `TestReconcileByKey` — retire carries/omits `transition_kind`; upserts never carry it.
- `TestEffectValidation` — `transition_kind` rejected on non-retire ops.
- `TestRetireFlipReport` — reports only open rows that flipped (deterministic order; already-resolved excluded); no report/`SELECT` without opt-in.
- `TestMachineResolutionTransitions` — one transition per real resolution; still-violating spared key silent; skipped-but-counted on replay by default; rebuild silent.

## Verification

`just check` — **600 passed, 1 skipped**, ruff check + format clean, docs build clean.

Docs: `docs/alerts-projection.md` "Remaining core gap" → shipped; CHANGELOG entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,16 @@ is not yet tagged in a release.
 
 ### Added
 
+- **Machine-resolution transitions for `reconcile_by_key`** (#32). A retire that
+  soft-deletes stale rows can now notify. Opt in with
+  `reconcile_by_key(..., transition_kind="alert_transition")`: the executor
+  captures the identities of the rows the open-guarded retire actually flipped
+  (NULL→set) and returns them in `ApplyReport.retire_flips`, and the replay
+  orchestrator emits one `external` transition per real resolution — skipped and
+  counted on replay by default, delivered under `include_external=True`, so a
+  rebuild never re-spams. `Executor.apply` now returns an `ApplyReport` (was
+  `None`). → [`docs/alerts-projection.md`](docs/alerts-projection.md).
+
 - **Versioned event handlers, upcasters & replay** (#3). Register handlers
   against a sequence range so old events run through the business logic that was
   correct *when they happened*; replay is idempotent and detects handler drift.
--- a/docs/alerts-projection.md
+++ b/docs/alerts-projection.md
@@ -45,7 +45,7 @@ re-derivation must never clobber a human's judgment, and vice-versa.
 | Layer | Owner | rakaia mechanism | Status |
 |---|---|---|---|
 | **1. Authored / informational** (`alert`, `backfilled`, …) | an actor | appended `alert_raised` / `alert_dismissed` events → natural-key `update_or_create` (+ `external` transition) | **shipped (Phase 1)** |
-| **2. Machine-reconciled** (validator violations) | the rules | `reconcile_by_key(..., retire={"resolved_at": ts})`, scoped by `retire_filter` to machine types | **shipped (Phase 2)** |
+| **2. Machine-reconciled** (validator violations) | the rules | `reconcile_by_key(..., retire={"resolved_at": ts})`, scoped by `retire_filter` to machine types (+ opt-in `external` transition per real resolution) | **shipped (Phase 2 + #32)** |
 | **3. Dismissable warnings** (derived ⊕ authored) | rules *and* actor | staged replay: stage-1 reconcile reads stage-0 `alert_dismissed` via the `reader` — authored wins for user-resolvable types | **shipped (Phase 3)** |
 
 ## Phase 1 — authored alerts (shipped, zero core changes)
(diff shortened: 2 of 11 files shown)


joshbrooks:
Follow-up hardening from post-merge review landed on `main` in 4fcfa28 (four findings on the retire-flip path):

1. **Locked the flip-capture SELECT** with `select_for_update()` so a concurrent txn can't resolve/mutate a captured row between the SELECT and the UPDATE.
2. **Explicit identity columns** — `reconcile_by_key` now passes scope + natural key in `transition_key_fields` and the executor uses them verbatim, so a scope key written as a lookup span (e.g. `org__id`) is no longer mistaken for a filter and dropped from the transition payload. This also removed the duplicated concrete-column filter in the executor.
3. **Typed `_synth_transitions(report: ApplyReport | None)`** (was `object`) with an explicit None guard, so a rename of `retire_flips` is a type error rather than a silent 'synthesise nothing' regression.

### ⚠️ Residual concern on #1 — needs a follow-up decision
`select_for_update()` locks the **existing** captured rows; it does **not** prevent a concurrent transaction from `INSERT`ing a **new** open row into the same reconcile scope, which the later `UPDATE` would then flip **but not report** — a silently-missed transition. Fully closing this 'phantom insert' window requires `SERIALIZABLE` isolation (or an `UPDATE ... RETURNING`-based capture, which isn't portable to SQLite). The lock closes the realistic 'captured row changed under us' case (false/duplicate transitions); the phantom-insert case is left open and should be resolved by either raising isolation for this apply or documenting the assumption of no concurrent inserts into a live reconcile scope.

Issue #68, opened 2026-07-27
Projection API ergonomics from the Partisipa port: verification helper, overlay executor, symbolic effect refs, match_field globs, cookbook


joshbrooks:
## Context

Follow-up to #7, which shipped staged replay + `DjangoProjectionReader` +
`CollectingExecutor` and proved them against Partisipa's multi-form projections.
Those capabilities work — this issue is the **ergonomics gap** surfaced by
actually wiring them in the port (`replay_tf611`, `replay_tf1321`, `replay_forms`).
The reporter's summary:

> The pieces are all there and well-docstringed, but every `replay_*` proof
> reduces to the same shape, and I hand-rolled the same normalization and
> staging ceremony three times. Each item below is "this worked, but it should
> have been in the library."

— Partisipa port adoption feedback, 2026-07-27.

Roughly ordered core → docs → minor.

## #1 · Verification-diff helper (`diff_effects_against_rows`)

**Problem.** Every replay proof is "run `replay()` with `CollectingExecutor`,
then diff each `update_or_create` effect's `defaults` against the live row." The
reporter wrote the same `Decimal`-quantize + `UUID`→`str` + FK-`attname`
normalization three times (`replay_tf611`, `replay_tf1321`, `replay_forms`).

**Proposed change.** Ship a `django_rakaia` helper:
`diff_effects_against_rows(effects, *, normalizers=…)` handling the universal
cases — `Decimal` → column `decimal_places`, `UUID` → `str`, `submission_id` key.
"Diff projected effects vs current rows" is *the* migration-verification pattern;
it belongs in the library.

**Acceptance.** A single call diffs a batch of collected effects against live
rows with the common normalizers applied by default; a test covers Decimal/UUID/FK
normalization.

## #2 · Overlay/buffered executor for verifying new-entity creation

**Problem.** Stage-1 handlers resolve FKs via `DjangoProjectionReader`, which
reads live tables. Under `CollectingExecutor` nothing is applied, so stage-1 only
resolves **pre-existing** targets. The reporter's proofs pass only because
legacy-bind targets already exist — a from-scratch replay that creates an
`ida_options.Project` and then binds to it **cannot be verified read-only** today.

**Proposed change.** A buffered/overlay executor that applies effects to an
in-memory shadow the reader can see, so staged from-scratch replay is verifiable
without touching the DB.

**Acceptance.** A staged replay that creates an entity in stage 0 and binds to it
in stage 1 verifies end-to-end with no DB writes.

[...]

PR #69, opened 2026-07-27
Projection API ergonomics: verification helper, event_match sets, register_simple, replay_stream (#68)

joshbrooks:
Addresses the registry/verification tranche of #68 (projection API ergonomics
from the Partisipa port) — the items that don't change replay/executor core
semantics. The two invasive items (#2 overlay executor, #3 symbolic refs) and
the #5 cookbook follow as separate PRs.

## #68 item 1 — `diff_effects_against_rows` verification helper

`django_rakaia.verification.diff_effects_against_rows(effects, *, reader,
normalizers, ops)` ships the projection-verification pattern every `replay_*`
proof re-derives by hand: run `replay()` with a `CollectingExecutor`, then diff
each write effect's `defaults` against the live row — read-only, no DB writes.

- Returns a `DiffReport` (`.ok` / `.problems` / `.rows` / `.raise_if_diff()`),
  with `RowDiff` / `FieldDiff` carrying expected-vs-actual per field, plus a
  readable `__str__`.
- Default normalizers cover the two universal "log carries a different
  representation than the column" cases: **UUID ↔ str** and an over-precise /
  JSON-float value vs the column's **Decimal** scale (quantized to
  `decimal_places`). Normalizers are pluggable — pass `[]` for raw comparison or
  a list to extend.
- FK `attname` (`x_id`) is resolved when locating the model field.
- Adds a `Measure` test model (`UUIDField` + `DecimalField`) + migration as the
  normalizer fixture.

## #68 item 4 — `event_match` accepts a set of patterns

`register_handler` / `register_simple` now take `event_match` as a glob **string**
(unchanged) **or an iterable of glob strings**, so one registration covers several
unrelated form_types that share no glob (e.g. `{"TF_6_1_1", "SF_1_2", "POM_1"}`)
instead of a register-once-per-value loop. A collection matches the routing
subject against **any** member; elements may themselves be globs.

- Canonicalised to a `frozenset`: hashable (works as the series key and in the
  identity tuple) and order-independent (same members in any order dedup).
- Persisted as a **sorted list** in the meta-stream and reconstructed to the same
  frozenset identity on reload, so store-backed dedup/rehydrate is preserved.
- Empty collections and non-string members are rejected.
- **Predicates are out of scope** — a callable can't be rehydrated from the
  persisted meta-stream.

## #68 minors

- **`register_simple(name, event_match, *, match_field, stage)`** — the always-on,
  single-stage "just project" case without the `effective_from=0` /
  `effective_to=None` ceremony. Delegates to `register_handler`; exported from
  `rakaia`.
- **`django_rakaia.replay.replay_stream(...)`** — defaults **both** the executor
  (`DjangoExecutor`) and reader (`DjangoProjectionReader`), so a staged replay
  from Django no longer fails with *"stage > 0 but no reader"* when the caller
  didn't wire a reader by hand. `manage.py replay` now routes through it, fixing
  staged replay via the command.

## Tests

23 new tests (11 verification, 3 `register_simple`, 2 `replay_stream`, 7
`event_match` sets incl. a persistence round-trip); staged replay through
`manage.py replay` covered. Full suite: **727 passed, 1 skipped**. `ruff check` +
`ruff format --check` clean.

### DIFF
--- a/src/django_rakaia/__init__.py
+++ b/src/django_rakaia/__init__.py
@@ -28,6 +28,9 @@ Public API:
     - stream_model: Decorator for automatically streaming model changes
     - create_stream_event: Helper function for manually creating stream events
     - register_stream_event_admin: Register stream event model with Django admin
+    - diff_effects_against_rows: Verify replayed effects reproduce the projection
+      (import from django_rakaia.verification — see that module; not eagerly
+      imported here because it pulls in the ORM before apps are ready)
 """
 
 # We intentionally do not eager import models here
--- a/src/django_rakaia/management/commands/replay.py
+++ b/src/django_rakaia/management/commands/replay.py
@@ -9,10 +9,8 @@ from typing import Any
 
 from django.core.management.base import BaseCommand, CommandParser
 
-from django_rakaia.effect_executor import DjangoExecutor
-from django_rakaia.store import get_store
+from django_rakaia.replay import replay_stream
 from rakaia.executors import CollectingExecutor
-from rakaia.replay import replay
 
 
 class Command(BaseCommand):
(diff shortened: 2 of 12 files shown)


joshbrooks:
**Review: approve** ✅

Well-scoped and well-tested — full suite 727 passed / 1 skipped, ruff clean. No blockers or majors. Minor doc-scope items only:

- **`verification.py:348` (minor)** — `update`-op verification uses `.get(...)` → `.first()`, so a multi-row `update` lookup only spot-checks the first row; rows 2..N could diverge and still report `ok`. Consistent with the docstring's "the row it targets," but worth a one-line note that broad `update` lookups are only spot-checked.
- **`registry.py:237,357,538` (minor)** — overlap/gap checks are per-series `(name, event_match)`. Two sets sharing a member (`{A,B}` vs `{A,C}`) won't flag an overlap on the shared member, and a gap in one series raises even if a sibling covers the seq. Mirrors the pre-existing glob behavior, so not a regression — but sets make it easier to reach. Consider documenting that overlap/coverage are per-registration.
- **`verification.py:295,364` (nit)** — `_resolve_field` re-scans `get_fields()` per field/side; negligible at proof scale.

Verified correct: frozenset canonicalization round-trips + dedups; `replay_stream` kwarg pass-through with lazy executor/reader; the `manage.py replay` staged-reader fix is the intended behavior change.

PR #70, opened 2026-07-27
Symbolic effect refs: bind to a sibling's generated key (#68 item 3)

joshbrooks:
Implements **#68 item 3**. A handler can bind one effect to the pk of the row a
sibling effect materialises — in the same batch, with no replay-stage split and
no natural-key reader lookup.

## The change

```python
def tf_6_1_1(event):
    return [
        Effect(op="update_or_create", model_label="ida_options.Project",
               lookup={"suku": event["suku"], "output": event["output"]},
               defaults={"name": event["project_name"]},
               produces="proj"),                                # names this row
        Effect(op="update_or_create", model_label="ida_forms.Tf_6_1_1",
               lookup={"submission_id": event["key"]},
               defaults={"project_id": Ref("proj")}),           # → Project.pk
    ]
```

New in `rakaia.effects` (exported from `rakaia`):

- **`Ref(produces, field="pk")`** — a placeholder value dropped into another
  effect's `lookup`/`defaults`/`patch`/`payload`/`spare_keys`. Defaults to the
  produced row's primary key; `Ref("proj", "suku")` picks another column.
- **`Effect.produces`** — a batch-local correlation id naming the single row an
  upsert materialises. Rejected on any op other than `update_or_create` (an
  `update` may match many rows; delete/retire/external produce none).
- **`UnresolvedRefError`** — a `Ref` to an id no earlier effect produced (forward
  ref / typo) raises, never binds a silent `None`.
- **`RefResolver`** — the shared resolution the applying executors use (kept in
  core so the forthcoming `OverlayExecutor` reuses it verbatim).

`DjangoExecutor` now, per `apply()` batch: resolves each effect's `Ref` values
before applying it, and records a `produces=` row's accessor right after
upserting it. Writes apply before deletes/retires (unchanged), so a producer is
always recorded before any later effect that refs it. `CollectingExecutor`
deliberately does **not** resolve — a dry run keeps the literal `Ref`s, which is
correct.

## Determinism

A `Ref("proj")` resolves to the **DB-assigned pk**. That's exactly right for an
FK column (referential integrity holds), but pks are not stable across a full
from-scratch rebuild — use a natural-key `field` when you need a rebuild-stable
value. Documented on `Ref`.

## Tests

12 new (7 core: `Ref`/`produces` validation, `RefResolver` incl. spare_keys +
no-ref fast path; 5 Django: FK bind, named-field ref, forward-ref raises,
`skip_unchanged` path). Full suite: **718 passed, 1 skipped**. `ruff check` +
`ruff format --check` clean.

Independent of #69 (touches `effects.py` / `effect_executor.py`, which #69 does
not). Next: #68 item 2, reframed as "replay into a disposable database"
(connection-alias seam), which reuses `RefResolver` from this PR.


joshbrooks:
**Review: request changes** 🔧

Mechanism is correct and well-tested (14 passed) — self-refs/cycles are impossible, forward refs raise loudly, fast-path identity holds. One issue I'd like closed before merge because it's the only *silent* failure in the change:

- **`effects.py:235` / `effect_executor.py:86` (blocker for this PR's own invariant)** — **duplicate `produces` id silently overwrites.** `RefResolver.record()` does `self._symbols[produces_id] = accessor` with no duplicate check. Two effects in one batch both tagged `produces="area"` silently overwrite: a downstream `Ref("area")` binds to the *second* producer's row and the first is unreachable — a silent wrong binding, which cuts against this PR's "never a silent None" principle. A cheap guard in `record()` (raise on re-declared id) closes it.

Minor / nits (deferrable):
- **`effect_executor.py:83-89` (minor)** — within writes, a consumer `update`/`update_or_create` must appear *after* its producer in list order (they share the first loop). This raises loudly, but the PR body's "producer always recorded first" overstates it — worth a one-line doc clarification.
- **`effects.py:72,315` (nit)** — `check_disjoint_defaults` runs pre-resolution, so two effects with different `Ref`s resolving to the same row won't collision-detect.
- **`effects.py:230,252` (nits)** — `payload` in `_MAPPING_FIELDS` is dead for `DjangoExecutor` (external-only); Refs nested inside list values aren't resolved and there's no guard rejecting them.

PR #71, opened 2026-07-27
using= connection seam: replay a from-scratch rebuild into a disposable DB (#68 item 2)

joshbrooks:
Implements **#68 item 2** — reframed. Rather than build an in-memory shadow that
imitates the ORM, point the **real** executor and reader at a **disposable
database**. Full ORM fidelity, no query engine to write, nothing touches
production.

> **Stacked on #70** (symbolic refs). Review/merge that first — this branch shares
> `effect_executor.py` and reuses its `RefResolver`. Base is
> `feat/symbolic-effect-refs`; it retargets to `main` once #70 lands.

## Why this shape

A `CollectingExecutor` answers a **regression** question ("does replay reproduce
the rows I already have?") — it writes nothing, so it leans on the targets already
existing. It can't answer the **rebuild** question ("can I build the whole
projection from the log into an empty schema?"), because a stage-1 handler
resolving a reference another form created would find nothing. That's the gap item
2 was about.

The honest fix isn't a bespoke plain-dict overlay (more code, *less* fidelity —
exact-match only, no constraints/defaults). It's a `using=` alias on the real
components:

```python
DATABASES["rebuild"] = {"ENGINE": "django.db.backends.sqlite3", "NAME": ":memory:"}

replay(
    store, "submissions",
    DjangoExecutor(using="rebuild"),                 # writes → scratch DB
    reader=DjangoProjectionReader(using="rebuild"),  # stage-1 reads them back
    handler_registry=reg,
)
# diff rebuilt rows vs prod on natural keys, then discard the scratch DB.
```

## The change

- **`DjangoExecutor(using=alias)`** routes every ORM call **and** the
  `transaction.atomic()` block to a named connection alias. Internal helpers
  (`_update`/`_delete`/`_retire`/`_upsert_skip_unchanged`) moved from static/class
  to instance methods so they share the alias via a `_manager(model)` helper.
- **`DjangoProjectionReader(using=alias)`** reads from the alias.
- `using=None` keeps default routing (`.using(None)` / `atomic(using=None)` are
  no-ops), so **single-DB behavior is byte-for-byte unchanged** — the whole
  existing suite still passes untouched.
- Refs from #70 resolve under the alias too.

## Docs

`docs/dry-run-and-executors.md` gains a **"Verifying a from-scratch rebuild"**
section: the regression-vs-rebuild distinction, plus two caveats stated plainly —
diff on **natural keys** (a rebuild assigns fresh pks), and sqlite `:memory:` is
full *ORM* fidelity but not full *Postgres* fidelity (use a scratch Postgres alias
for the cut-over proof; same code, different `using=`).

## Not built (deliberately)

The plain-dict in-memory `OverlayExecutor`. It's more code for less fidelity, and
only a *non-Django core* consumer — none today — would need it. YAGNI; the seam
keeps that door open.

## Tests

Test settings gain an `overlay` alias. New `test_using_seam.py` (6 tests): writes
route to the alias only; the default alias is untouched when `using=None`; reader
`get/filter/query` route; a **staged from-scratch rebuild links in the overlay
while `default` (standing in for prod) stays empty**; refs resolve under the
alias. Full suite: **724 passed, 1 skipped**; `ruff` + `ruff format` + `zensical
build` clean.

Foundation for the Partisipa from-scratch rebuild proof + natural-key rebuild diff
(tracked in `docs/streams-migration.md`).

### DIFF
--- a/docs/dry-run-and-executors.md
+++ b/docs/dry-run-and-executors.md
@@ -102,6 +102,47 @@ Drop `--dry-run` to apply the same effects via the `DjangoExecutor`. See
 [Versioned handlers](versioned-handlers.md) for the full command reference
 (`--from`, `--to`, `--strict-drift`, `--include-external`).
 
+## Verifying a from-scratch rebuild: the `using=` seam
+
+A `CollectingExecutor` answers a **regression** question: *"does replaying the log
+reproduce the rows I already have?"* It writes nothing, so it leans on the target
+rows already existing. It cannot answer the **rebuild** question — *"can I build
+the whole projection from the log into an empty schema, correctly?"* — because a
+stage-1 handler that resolves a reference another form created would find nothing
+(the reference was recorded, never applied), so the link can't be verified.
+
+The honest answer is not a bespoke in-memory shadow that imitates the ORM, but the
+**real** executor and reader pointed at a **disposable database**. Both
+`DjangoExecutor` and `DjangoProjectionReader` take a `using=` database alias:
+
+```python
+# settings: a throwaway alias — in-memory sqlite for fast/CI proofs,
+# or a scratch Postgres for a full-fidelity cut-over rehearsal.
+DATABASES["rebuild"] = {"ENGINE": "django.db.backends.sqlite3", "NAME": ":memory:"}
+
+replay(
+    store, "submissions",
+    DjangoExecutor(using="rebuild"),                 # writes land in the scratch DB
+    reader=DjangoProjectionReader(using="rebuild"),  # stage-1 reads them back
+    handler_registry=reg,
+)
+# ...then diff the rebuilt rows against production and throw the scratch DB away.
+```
+
+Because it is the real ORM, you get full fidelity — constraints, defaults, joins,
+`__gte` — for free, with no query engine to write and nothing written to
+production.
+
+Two caveats worth stating plainly:
+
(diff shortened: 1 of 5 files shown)


joshbrooks:
**Review: approve** ✅ (stacked on #70 — merge that first)

Seam correctly and uniformly applied; `test_using_seam.py` 6 passed + 34 regression pass. `using=None` is a byte-for-byte no-op (untouched suite proves it). No blockers/majors.

- **`test_using_seam.py` (minor)** — test gap: the alias is routed through 5 methods but new tests only cover `update_or_create` + refs. Untested: `op="update"`, `delete`, `retire`, and `skip_unchanged`. The riskiest are `_retire` (`effect_executor.py:188` `select_for_update`) and `_upsert_skip_unchanged` (`effect_executor.py:142` `row.save(using=...)`) — all currently correct, but a future refactor could silently drop the alias and leak writes into the production DB with no test catching it. Suggest adding these as a fast-follow.
- **`effect_executor.py:142` (nit)** — `row.save(using=self._using)` is the one non-uniform call (not via `_manager`); correct only because the fetch and save share the alias.

Verified correct: event log stays on default/production while only projection R/W route to the alias (the rebuild design); both new params are keyword-only so no positional breakage; no router/teardown problem with the two `:memory:` connections.

PR #90, opened 2026-08-14
Refuse an unknown RAKAIA_STORE instead of silently using memory

joshbrooks:
Rakaia picks its event store from one Django setting, `RAKAIA_STORE`, which is
free-form text. Until now anything that wasn't exactly `"durable"` got the
in-memory store — so a single mistyped character, or a stray space picked up
from a `.env` file, quietly selected a store that keeps everything in process
memory. Nothing failed and nothing warned: appends succeeded, the app looked
healthy, and the entire event log disappeared at the next restart. A setting
that is one keystroke away from losing all your data should not fail silently.

An unrecognised backend is now refused outright, with an error that names what
you typed and what the valid choices are. A new startup check reports the same
problem when you run `manage.py check`, so a misconfigured deployment fails
before it serves a single request rather than hours later when a background
worker happens to write its first event. Running in memory deliberately is still
fine; doing it with `DEBUG` switched off now gets a warning you can silence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,6 +149,20 @@ is not yet tagged in a release.
 
 ### Fixed
 
+- **`RAKAIA_STORE` no longer fails open.** `get_store()` returned the
+  *in-memory* store for any backend string that wasn't exactly `"durable"`, so a
+  one-character typo (`"durrable"`, a stray space from a `.env` file, `"Durable"`)
+  selected a process-local dict while the deployment believed it was durable:
+  appends succeeded, nothing warned, and the entire event log vanished on the
+  next restart. An unknown backend now raises `ImproperlyConfigured` naming the
+  valid choices, and a refused backend is not cached, so correcting the setting
+  works without a restart. A new system check reports the same problem at
+  startup (`rakaia.E001`) rather than on the first append — which in a worker
+  process could be hours in — and `"memory"` with `DEBUG = False` warns
+  (`rakaia.W001`). ADR 0002 named this gap ("`RAKAIA_STORE` swaps stores by
+  string with no interface check"); the first production consumer had written
+  its own check downstream to catch it.
+
 - **A store now refuses an offset it did not issue.** Both stores accepted the
   other's offset format and resolved it to an unrelated position instead of
   failing: `int("0_5")` is `5` in Python, so the durable store read the wrong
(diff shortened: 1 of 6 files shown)


joshbrooks:
### Technical detail

**RED first.** `tests/test_django_rakaia/test_store_selection.py` before the fix:

```
8 failed, 3 passed in 0.20s
FAILED ...::test_unknown_backend_raises_instead_of_falling_back[durrable]
FAILED ...::test_unknown_backend_raises_instead_of_falling_back[durable ]
FAILED ...::test_unknown_backend_raises_instead_of_falling_back[Durable]
FAILED ...::test_unknown_backend_raises_instead_of_falling_back[postgres]
FAILED ...::test_unknown_backend_raises_instead_of_falling_back[db]
FAILED ...::test_unknown_backend_raises_instead_of_falling_back[]
FAILED ...::test_the_error_names_the_backend_and_the_valid_choices
FAILED ...::test_a_refused_backend_is_not_cached
E  Failed: DID NOT RAISE <class 'django.core.exceptions.ImproperlyConfigured'>
```

The 3 that passed are the happy paths (`unset` → memory, `"memory"`, `"durable"`),
which pins that this change is a narrowing and not a behaviour swap.

**The shape.** `_build_store` was `if backend == "durable": ... ; return StreamStore()`
— a fallback in the `else` position. It is now an explicit two-arm match with a
raise, plus `BACKENDS` / `DEFAULT_BACKEND` constants that `checks.py` imports so
the selector and the check cannot disagree about what is valid.

**Cache poisoning.** `_stores[backend] = _build_store(backend)` evaluates the
right-hand side first, so a raise leaves no entry behind. `test_a_refused_backend_is_not_cached`
pins that: correcting the setting works without restarting the process. Worth
having a test for because the obvious alternative implementation (assign a
sentinel, validate after) would regress it silently.

**Check registration** is via importing `django_rakaia.checks` first in
`AppConfig.ready()`, ahead of autodiscovery — so a bad `RAKAIA_STORE` is reported
even when the rest of `ready()` would fail for the same reason.

`ARG001` is suppressed on `check_store_backend`: Django's check API mandates the
`(app_configs, **kwargs)` signature.

**Gate:** 996 passed (981 baseline + 15 new), `ruff check` and `ruff format --check` clean.

**Not done here:** `get_store()` still returns `Any`. Typing it as
`StreamServerStore` is a real improvement but it changes the public signature,
so it belongs with the protocol work rather than a config fix.

PR #91, opened 2026-08-14
A verification that compared nothing should not report itself clean

joshbrooks:
The way you prove a migration is safe with rakaia is to replay the log, collect
what it would write, and diff that against the rows you already have. If nothing
disagrees, you cut over. The problem is what happens when the diff has nothing to
compare: it finds no disagreements, so it reports success. A run that examined
ten thousand rows and a run that examined none give you the same green light.

That is easy to hit by accident — a store pointed at the wrong backend, a replay
over a stream path that got renamed, a filter that quietly stopped matching, a
registry that failed to load. Any of those hands you an empty batch, and the
proof tells you everything is fine. This adds a third answer alongside pass and
fail: *nothing was compared, so this run certifies nothing*. A real failure still
takes priority over it, so an empty-looking result can never mask a genuine one.
If you honestly expect nothing to compare, you now have to say so at the call
site rather than getting it by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -163,6 +163,22 @@ is not yet tagged in a release.
   string with no interface check"); the first production consumer had written
   its own check downstream to catch it.
 
+- **A verification sweep no longer certifies a population of zero.**
+  `diff_effects_against_rows(...).ok` answers "did anything disagree?", which is
+  vacuously `True` when nothing was compared — so an empty effect list reported
+  a clean bill of health with nothing behind it. The ways that happens are all
+  ordinary: a store on the wrong backend, a replay over a renamed stream path,
+  an `event_match` filter that stopped matching, a registry that failed to
+  autodiscover. `DiffReport` now carries a three-state `verdict`
+  (`GREEN`/`RED`/`VACUOUS`) plus `compared` and `certified`, and
+  `raise_if_diff()` raises the new `VacuousVerification` rather than passing.
+  Failures are checked **before** vacuity, so a real failure is never hidden
+  behind "empty"; pass `allow_empty=True` where a zero population is genuinely
+  expected. `ok` keeps its original meaning, so existing callers are unaffected.
+  Upstreamed from the first production consumer, which named this **the vacuous
+  green** and had built the same guard downstream.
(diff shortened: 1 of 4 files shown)


joshbrooks:
### Technical detail

**RED first.** On `main`, the vacuous green reproduced directly:

```
>>> r = DiffReport(rows=[])
ok            = True
raise_if_diff = None
str           = DiffReport: 0 row(s) verified, no differences
```

`tests/test_django_rakaia/test_verification_vacuous.py` fails at import against
`main` (`ImportError: cannot import name 'RED'`), then 14 pass after the change.

**The rule, and why the order matters.** `verdict` checks `problems` *before*
`compared == 0`:

```python
if self.problems:      return RED
if self.compared == 0: return VACUOUS
return GREEN
```

If anything disagreed, something was evidently compared — so RED is strictly
safer than reporting "empty" and letting a real failure hide behind it. Pinned by
`TestVerdictOrdering`, including the case worth being explicit about: a **missing
row** is `RED`, not `VACUOUS`. The lookup ran and found nothing; that is a
failure, not an empty population.

**`compared` counts checked rows, not input effects.** Non-write effects
(delete/retire/external) carry no values to diff and are skipped, so a batch of
only those compares nothing and is correctly `VACUOUS` — see
`test_only_non_write_effects_were_produced`. That case is the one most likely to
surprise, since the caller did pass a non-empty list.

**Compatibility.** `ok` is untouched (`not self.problems`), so existing callers
and the `VerificationError` message are unaffected — the full suite goes 981 →
995 with no existing test modified. `VacuousVerification` is a *sibling* of
`VerificationError`, not a subclass: they mean opposite things (wrong vs. no
evidence). Both subclass `AssertionError`, so a proof written as
`except AssertionError` still catches either.

**Opt-out** is `raise_if_diff(allow_empty=True)` — explicit, at the call site,
and it still raises on a real failure (`test_allow_empty_does_not_suppress_a_real_failure`).

**Gate:** 995 passed, `ruff check` + `ruff format --check` clean.

**Follow-up not done here:** the same guard belongs on any future `verify_*`
management command, and `docs/projection-cookbook.md` now recommends
`assert report.certified` over `assert report.ok`. Existing example code still
using `.ok` is correct but weaker; sweeping those is left for the demo pass.

PR #92, opened 2026-08-14
Ship the other half of the rebuild guard

joshbrooks:
Rakaia can rebuild your database tables from scratch by replaying the event log
into a throwaway database, then comparing the result against what you already
have. For that comparison to mean anything, the rebuild has to leave the real
database completely alone — it must not read from it, and it must not write to
it. We shipped the half that catches reads. The half that catches writes was
described in our own documentation and in ADR 0003, but never actually included,
so anyone adopting rakaia had to write it themselves.

It ships now. You wrap a rebuild, name the tables that matter, and if the real
database gains or loses a single row during the run you get a loud failure that
says which table and by how much. The specific accident it is built for is easy
to have and hard to notice: Django save-hooks keep firing during a rebuild, and
one that forgets to say which database it is writing to lands on the live one —
quietly corrupting production during the very run that was supposed to prove
nothing was touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,21 @@ is not yet tagged in a release.
   staged replay. Upstreamed from the Partisipa migration's `PreloadReader`
   workaround. → [`docs/projection-cookbook.md`](docs/projection-cookbook.md).
 
+- **`assert_no_live_writes` — the rebuild gate's write-side guard** (ADR 0003).
+  `django_rakaia.hermeticity` has documented this as the mirror of
+  `deny_database_access` since ADR 0003 landed, but only shipped the read half;
+  the write half lived in the first consumer's tree and every adopter would have
+  re-derived it. It now ships here: wrap a from-scratch rebuild in
+  `assert_no_live_writes(*models, using="default")` and any change to the live
+  database's row counts raises `LiveWriteLeaked` naming the drift. The leak it
+  exists to catch is a `post_save`/`pre_save` receiver saving without a `using=`
+  — which Postgres' `session_replication_role = replica` does *not* disable, so
+  a rebuild can silently mutate production while reporting itself green.
+  Unlike the read guard (an `execute_wrapper`, so it cannot be armed where any
+  legitimate query to the alias happens), counting rows tolerates reads — so this
+  is the guard that can wrap a whole rebuild whose event log lives on the
+  guarded alias.
+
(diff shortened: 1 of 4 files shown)


joshbrooks:
### Technical detail

**RED first.** `tests/test_django_rakaia/test_rebuild_isolation.py` against `main`:

```
E  ImportError: cannot import name 'LiveWriteLeaked' from 'django_rakaia.hermeticity'
```

12 pass after the change; suite goes 981 → 993 with no existing test touched.

**Why row counts and not a statement wrapper.** The obvious symmetry would be to
reuse `execute_wrapper` and reject writes. That does not work for the case this
guard is for: an `execute_wrapper` cannot be armed anywhere a *legitimate* query
to the alias happens, and a rebuild's event log frequently lives on the very
alias being guarded — the constraint `deny_database_access`'s docstring already
records ("a `DjangoStreamStore` on `default` would itself trip the guard"). So
the read guard can only cover a hand-picked region. Counting rows tolerates
arbitrary reads, which makes this the guard you can wrap the *whole* rebuild in.
`test_reads_of_the_live_alias_are_tolerated` pins that difference.

**`finally`, not `else`.** The counts are compared even when the block raised.
A rebuild that fell over half-way may already have leaked, and that is precisely
the leak likely to go unnoticed. Both directions are pinned:
`test_the_original_exception_propagates_when_nothing_leaked` (the block's own
error is the more informative one and must survive) and
`test_a_leak_is_reported_even_if_the_block_raised`.

**Deletes count too** — `test_a_delete_on_the_default_alias_raises`. A rebuild
that *removes* live rows is as much a live mutation as one that adds them, and
an implementation that only checked `after > before` would pass it.

**Known limit, documented on the function:** it compares counts, so an in-place
`UPDATE` of an existing row is invisible. That is a deliberate trade — counts are
what a rebuild leak actually looks like (a receiver minting rows), they are cheap
enough to wrap a whole replay, and they need no per-model knowledge of which
columns matter. Field-level assurance is `diff_effects_against_rows`'s job, and
the docstring points there.

**Also updated:** the module docstring no longer describes the write guard as
living downstream — it now introduces both halves and explains why neither
subsumes the other. ADR 0003 gets a dated update note recording that the
implementation moved upstream; the decision itself is unchanged.

**Gate:** 993 passed, `ruff check` + `ruff format --check` clean.

PR #93, opened 2026-08-14
@stream_model was throwing away who made the change

joshbrooks:
Rakaia can record who made each change and what request it came from, and it
ships middleware that fills that in automatically for every web request. It turns
out the one append path most Django users actually use — the `@stream_model`
decorator on their models — was quietly dropping all of it. The middleware was
doing its job; the decorator just wasn't listening. Every event it wrote had an
empty author field.

The visible symptom is that the audit history answered the wrong question. Asked
who saved a record, it fell back to whoever owns the record — usually a different
person, and no way to tell from the log which one you were looking at. The
decorator now fills in the same envelope every other append path does, including
the logical timestamp that replay uses to order events. Nothing changes for code
that wasn't setting this in the first place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -194,6 +194,18 @@ is not yet tagged in a release.
   green** and had built the same guard downstream.
   → [`docs/projection-cookbook.md`](docs/projection-cookbook.md)
 
+- **`@stream_model` now records the ambient envelope.** The decorator wrote
+  `StreamEvent` rows directly rather than through a store, so `metadata` was
+  always `{}` and `event_ts` always `NULL` — on the most-used append path in the
+  Django integration. Two consequences: `ProvenanceMiddleware`, which exists to
+  stamp actor and URL onto envelopes, could not reach the appends it was built
+  for, and `history.envelope_actor` silently fell back to the payload's owner FK,
+  answering "who owns this" in place of "who saved this". `event_ts` being NULL
+  also forced `merge_replay` onto transport time, the ordering trap ADR 0002
+  item 5 closed. The decorator now calls `merge_provenance` and sets `event_ts`
+  exactly as both stores do; a fan-out to several streams shares one envelope,
+  since it is one event. Appends outside a `provenance()` block are unchanged.
+
 - **A store now refuses an offset it did not issue.** Both stores accepted the
   other's offset format and resolved it to an unrelated position instead of
   failing: `int("0_5")` is `5` in Python, so the durable store read the wrong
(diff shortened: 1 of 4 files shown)


joshbrooks:
### Technical detail

**RED first.** `tests/test_django_rakaia/test_decorators_provenance.py` on `main`:

```
7 failed, 2 passed in 0.53s
E  KeyError: 'user'
FAILED ...::TestAmbientProvenanceIsRecorded::test_a_save_inside_provenance_records_the_actor
FAILED ...::TestAmbientProvenanceIsRecorded::test_every_ambient_field_is_recorded
FAILED ...::TestAmbientProvenanceIsRecorded::test_an_update_records_provenance_too
FAILED ...::TestAmbientProvenanceIsRecorded::test_a_delete_records_provenance_too
FAILED ...::TestEnvelopeTimestamp::test_event_ts_is_set
FAILED ...::TestEnvelopeTimestamp::test_event_ts_is_monotonic_across_successive_saves
FAILED ...::TestFanOutSharesOneEnvelope::test_a_multi_stream_event_records_provenance_once
```

The 2 that passed are `TestNoAmbientProvenance` — the no-block behaviour, which
pins that this is purely additive. Suite goes 981 → 990.

**Root cause.** `create_stream_event` calls `StreamEvent.objects.create(data=…,
event_type=…)` directly. Both stores route through `merge_provenance` —
`store.py:321` and `django_store.py:542` / `:768` — but this third write path
never did. Grepping `merge_provenance` before the change returns those two stores
and nothing else, which is the whole bug in one command.

**Fix** is two lines at the call site: `merge_provenance(None) or {}` and
`time.time()`. `merge_provenance(None)` returns `None` when no block is active,
hence the `or {}` — preserving the existing empty-dict shape rather than writing
`NULL` into a column that is `{}` everywhere else.

**One envelope per event, not per entry.** `metadata` / `event_ts` go on
`StreamEvent`, which the `StreamEntry` rows share. A `@stream_model` fan-out is
*one* event appearing in several streams, so it has one author and one logical
time. `test_a_multi_stream_event_records_provenance_once` asserts exactly that
(`entries.count() > 1`, one new `StreamEvent`).

**Scope.** This is the provenance half of the wider finding that
`create_stream_event` bypasses the store entirely. The bypass also means
`@stream_model` skips the closed-stream and expiry checks and ignores
`content_type`/`payload_encoding`. Closing that properly needs the store to grow
the one capability the bypass exists for — writing one event into N streams
(`append_to_many`) — so `create_stream_event` can become resolve-paths →
build-payload → call the store. That is a bigger change with its own contract
tests and is deliberately **not** in this PR; this one is a correctness fix with
a one-line RED.

**Gate:** 990 passed, `ruff check` + `ruff format --check` clean.

PR #94, opened 2026-08-14
Stop making every project retype the same two rituals

joshbrooks:
There are two small pieces of boilerplate that anyone using rakaia with Django
ends up writing over and over: the handful of lines that record an event with its
author and timestamp, and the handful that run a batch of events through your
projection logic right now rather than waiting for a rebuild. Neither is hard.
Both are easy to get subtly wrong, and a copy that drifts — a missing timestamp,
an author field left empty — produces events that behave differently from every
other event in the same stream, with no test anywhere looking at the difference.
The first project to use rakaia in production ended up with around thirty-seven
copies of one and eleven of the other.

Both now ship with rakaia. The tests check them against the longhand they
replace rather than against something hand-written, so switching over is pure
deletion with nothing to re-verify. Along the way this also pins down a rule the
codebase had been relying on without stating: asking to create a stream that
already exists leaves its contents alone. That was true of both stores but
untested, and it's what makes the shorter form safe — so rakaia's own code now
uses the shorter form too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,24 @@ is not yet tagged in a release.
   `except KeyError` code and tests are unaffected. A new failure type without a
   status now fails the suite rather than 500ing at runtime.
 
+- **`append_event` and `fold_events` — the two rituals every consumer retypes.**
+  Appending an enveloped event (JSON-encode with `DjangoJSONEncoder`,
+  create-the-stream-if-missing, wrap label/actor/`event_ts` in `AppendOptions`)
+  and folding a batch live (seed a scratch in-memory `StreamStore`, replay it
+  through a registry with a reader bound) were hand-written at nearly every
+  durable call site — ~37 and 11 copies respectively in the first production
+  consumer, whose own module carried the warning that motivates this: *"a second
+  write path which re-implements the envelope is a path no gate covers."* Both
+  now ship from `django_rakaia.envelope`, pinned byte-for-byte against the
+  longhand they replace. Upstreamed as its ADR-0020 anticipated.
+
+  The store contract also gained the property that makes the create-if-missing
+  shorthand safe: a redundant `create()` on a populated stream preserves its
+  messages and does not rewind its offsets. Both stores already satisfied it;
+  nothing had said so, and `test_create_is_idempotent` would have passed a
+  `create()` that truncated. With it pinned, the `has()`/`create()` dance is
+  provably redundant — `registry.py`'s three copies are now a single `create()`.
+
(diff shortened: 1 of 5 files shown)


joshbrooks:
### Technical detail

**RED first.** `tests/test_django_rakaia/test_envelope.py` against `main`:

```
E  ModuleNotFoundError: No module named 'django_rakaia.envelope'
```

14 pass after. Suite 981 → 999 (14 envelope + 4 contract cases × both stores).

**The oracle pattern.** These helpers are only worth adopting if adopting them is
a pure deletion, so `TestAppendEventMatchesTheLonghand` keeps `_hand_rolled()` —
the exact longhand being replaced — in the test file and asserts the two produce
the same `data` / `label` / `metadata`. Pinning against a hand-written expectation
instead would prove the helper does *something* consistent, not that it does the
same thing the 37 existing call sites do.

**`metadata={"user": actor}`, always a dict.** That is the key
`rakaia.history.envelope_actor` already reads, pinned by
`test_the_actor_lands_where_envelope_actor_reads_it` — the helper adopts the
existing convention rather than inventing one. `test_ambient_provenance_still_merges`
checks the explicit `user` doesn't shut ambient `provenance()` out of the other
fields, since `merge_provenance` layers ambient *under* explicit.

**The contract addition is the load-bearing part.** `create()` unguarded is only
safe if a redundant `create()` is a no-op rather than a reset — and
`test_create_is_idempotent` did **not** establish that: it asserts `has("s")`,
which a `create()` that truncated would also satisfy. Two new cases close it:

- `test_create_on_an_existing_stream_preserves_its_messages`
- `test_create_on_an_existing_stream_does_not_rewind_offsets`

Both stores already passed them, so these are characterization tests, not a bug
fix — they license the simplification. With them in place `registry.py`'s three
`if not store.has(p): store.create(p)` sites become one `create()` each,
halving the round trips on the meta-stream path.

**Placement.** `envelope.py` is in `django_rakaia`, not `rakaia`: the encoder is
`DjangoJSONEncoder` and the fold's default executor writes Django models. The
core package stays dependency-free.

**`fold_events` semantics** worth flagging: fresh scratch store per call
(`test_successive_folds_do_not_leak_into_each_other`), in-memory so a fold leaves
no durable trace (`test_it_leaves_no_durable_trace`), and an empty batch returns
before touching anything. Its reason to exist is that write-time and rebuild-time
projection run the *same* handler code — otherwise a rebuild that disagrees with
live is indistinguishable from a rebuild that is correct.

**Gate:** 999 passed, `ruff check` + `ruff format --check` clean.

PR #95, opened 2026-08-14
Bulk appends were never reaching live subscribers

### LINKED ISSUE #82: append_many bypasses post_save, so bulk appends never reach SSE subscribers (diverges from append)
`DjangoStreamStore.append_many` `bulk_create`s its `StreamEvent`s and then its
`StreamEntry`s (`src/django_rakaia/django_store.py:157-174`). `bulk_create` does
not send `post_save`, so `channels_signals.handle_stream_entry_created` never
fires and nothing is broadcast. Single `append` goes through
`StreamEntry.objects.create` (`django_store.py:104`) and broadcasts normally.

Live subscribers silently miss every bulk-appended event. There is no error and
no gap in the durable stream — a client that reconnects catches up, because the
SSE view replays from `Last-Event-ID` (`channels_views.py:33-57`, filtering
`offset__gt=after_offset`) — but a client that stays connected simply never sees
them. The failure is invisible from the writer's side, which is what makes it
worth fixing rather than documenting.

Scope: only bites when the SSE broadcast is wired, i.e. `RAKAIA_ENABLE_SSE` and
`channels` (`apps.py:28-35`). The durable rows are correct either way.

The docstring currently claims the opposite (`django_store.py:110-116`):

> Semantically identical to calling :meth:`append` once per item — same event
> rows, same envelope mapping (``label``->``event_type``, per-event
> ``merge_provenance`` for ``metadata``, ``event_ts``) [...]

That equivalence holds on the durable-row axis and breaks on the signal axis.

## Repro

Reproduced on `main` at `5e4a6e3`. Patch a recording channel layer over
`channels_signals.get_channel_layer`, then append the same events both ways:

```python
store.append("bulk:probe", json.dumps({"n": 1}).encode(), None)
# -> 1 broadcast

store.append_many(
    "bulk:probe",
    [(json.dumps({"n": i}).encode(), None) for i in (2, 3, 4)],
)
# -> 0 broadcasts, expected 3
```

```
append          -> 1 broadcast(s)
append_many(3)  -> 0 broadcast(s)
```

## Introduced by

#79 (`feat(store): append_many bulk-append on both stores`). Not a regression in
any prior behaviour — `append_many` did not exist before — but it lands on the
same fan-out path #80 has been hardening, so the two interact.

## Options

1. Send `post_save` explicitly for each created `StreamEntry` after
   `bulk_create`, restoring true equivalence at the cost of the per-row signal
   overhead `bulk_create` exists to avoid.
2. Broadcast from the store directly — one `group_send` per batch, or a single
   batched frame — and leave the signal path for the ORM-driven writes.
3. Narrow the docstring claim to the durable-row axis and document that bulk
   appends are not broadcast, if catch-up-on-reconnect is considered sufficient.

Option 2 seems the most honest fit: a batch write probably wants a batch frame
rather than N frames, and it keeps `bulk_create`'s reason for existing intact.

Found during `/review` of #81.

joshbrooks:
Rakaia can push new events to connected browsers as they happen. It also has a
fast path for writing many events at once, added so that seeding and replaying
don't cost one database round trip per event. Those two features did not work
together: anything written through the fast path was saved correctly but never
announced, so live pages simply never heard about it. You would only notice on
reload, when the missing events suddenly appeared.

The cause is that live notification had been riding on a database hook that the
fast bulk write deliberately skips — that skipping is the whole reason the fast
path is fast. So the store now announces its own writes rather than hoping
something downstream notices, and there is one shared definition of what a
notification looks like so the two paths cannot describe the same event
differently. Fixes #82.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -229,6 +229,18 @@ is not yet tagged in a release.
   exactly as both stores do; a fan-out to several streams shares one envelope,
   since it is one event. Appends outside a `provenance()` block are unchanged.
 
+- **A bulk append now reaches live subscribers** (#82). `append_many` persists
+  with `bulk_create`, which does not fire `post_save` — and on the durable store
+  publication *was* a `post_save` receiver on `StreamEntry`, so every bulk append
+  was silently invisible to SSE subscribers. The docstring's promise that it is
+  "semantically identical to calling `append` once per item" was true of the rows
+  written and false of the subscribers reached; nothing caught it because that
+  semantic was never in the interface. The wire frame now has a single definition
+  (`channels_signals.broadcast_entries`) which both the receiver and the store
+  call, and `DjangoStreamStore` publishes its own appends rather than relying on
+  a signal that its own write strategy bypasses. Restoring the signal by saving
+  rows one at a time would have undone the reason `append_many` exists.
+
 - **A store now refuses an offset it did not issue.** Both stores accepted the
   other's offset format and resolved it to an unrelated position instead of
   failing: `int("0_5")` is `5` in Python, so the durable store read the wrong
--- a/src/django_rakaia/channels_signals.py
+++ b/src/django_rakaia/channels_signals.py
@@ -141,10 +141,13 @@ def broadcast_entries(stream_id: str, entries: Sequence[StreamEntry]) -> None:
 def handle_stream_entry_created(sender, instance, created, **kwargs):  # noqa: ARG001
     """Broadcast an ORM-written stream entry to the stream's channel group.
 
-    Covers the write paths that do not go through `DjangoStreamStore` — chiefly
-    `@stream_model`. The store publishes its own appends via `broadcast_entries`
-    and writes with `bulk_create`, so this receiver does not double-fire for
-    them.
+    Covers every entry written through the ORM: `@stream_model`, and
+    `DjangoStreamStore.append`, which persists one entry with `create()`.
+
+    It does *not* cover `DjangoStreamStore.append_many`, which persists with
+    `bulk_create` — that path calls `broadcast_entries` itself. So the two
+    mechanisms partition the write paths rather than overlapping, and no entry is
+    published twice.
     """
     # Skip fixture rows. Beyond the phantom frame, the `instance.stream` /
     # `instance.event` dereferences are separate queries that raise DoesNotExist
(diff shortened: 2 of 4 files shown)


joshbrooks:
### Technical detail

**RED first.** `tests/test_django_rakaia/test_publish_on_append.py` on `main`:

```
4 failed, 2 passed in 13.09s
E  AssertionError: assert 0 == 2
E   +  where 0 = len([])          # two events via append_many  -> 0 frames
E   +  and   2 = len([{...}, {...}])  # two events via append   -> 2 frames
```

The 2 that passed are the single-append case and the empty batch — kept so the
change reads as a widening, not a swap. Suite 981 → 987.

**Why the tests drive the store, not the ORM.** `test_channels.py` already covers
this area, and passes on `main`, because it creates `StreamEntry` rows through
the ORM — which fires `post_save`. The bug only exists when you go through
`DjangoStreamStore`, so these tests do, via `run_sync`. That difference is the
whole reason the gap survived: the existing tests were exercising the mechanism
rather than the behaviour.

**Shape of the fix.** The frame body moved out of the receiver into
`_frame(stream_id, entry)`, with `broadcast_entries(stream_id, entries)` around
it. Two callers:

- `handle_stream_entry_created` — now a four-line receiver, covering ORM writes;
- `DjangoStreamStore._publish` — called from `append_many` after `bulk_create`.

`_publish` imports `channels_signals` lazily inside a `try/except ImportError`,
because `channels` is an optional extra (ADR 0002 / #41) and a framework-tier
consumer must still be able to append.

**No double-publishing.** `append_many` uses `bulk_create` (no signal) and
publishes explicitly; `append` uses `objects.create` (signal fires) and does not.
`test_one_bulk_append_delivers_what_n_single_appends_deliver` pins the parity
from the subscriber's side — exactly two frames either way.

**Wire shape is unchanged.** The frame carries the *raw* entry offset, not the
zero-padded protocol form `read()` returns. That is what the receiver has always
sent and what live SSE consumers already parse, so
`test_offsets_are_the_ones_the_store_assigned` asserts against the raw column
rather than "fixing" the format. Changing it here would be an unannounced break
for anything currently connected.

**Publishing happens inside the transaction**, matching the receiver's existing
timing on the `append` path. Publishing before commit means a rolled-back
transaction can still have broadcast — that is pre-existing on the single-append
path, and making both paths commit-hooked is a separate behaviour change I did
not fold in here.

**Scope note.** This does not delete the `post_save` receiver. Doing so requires
`@stream_model` to stop writing `StreamEntry` rows behind the store's back —
tracked separately (the same bypass #93 flagged), and it needs the store to grow
one-event-into-N-streams first. Until then the receiver is the correct home for
ORM-path publication, and it now shares the store's frame definition rather than
owning a second copy.

**Gate:** 987 passed, `ruff check` + `ruff format --check` clean.

PR #96, opened 2026-08-14
The two stores disagreed about whether an append was allowed

joshbrooks:
Rakaia has two places to keep an event log — one in memory for tests and small
apps, one in your database for real deployments — and the whole point is that
they behave identically, so you can develop against one and ship on the other.
They didn't. The rules for deciding whether a write is allowed at all (is the
stream closed, does the content type match, is this a duplicate retry, is the
sequence number sane) were written out twice, once in each store, and the two
copies had already drifted apart. The database-backed one silently ignored the
duplicate-detection information callers passed it, while its documentation
claimed the opposite.

Those rules now live in one place that both stores consult, so a store's job is
to fetch some facts, ask, and then save — not to re-implement the protocol. The
shared test suite that both stores must pass has grown seven cases covering the
door this slipped through. As a bonus the rules are now testable on their own,
without spinning up a store or a database, which is what makes the ordering
between them — genuinely fiddly, and previously implicit in the order of some
if-statements — something we can actually pin down.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -241,6 +241,26 @@ is not yet tagged in a release.
   a signal that its own write strategy bypasses. Restoring the signal by saving
   rows one at a time would have undone the reason `append_many` exists.
 
+- **`append` now honours producer options on both stores.** `StreamStore.append`
+  validated producer epoch/seq inline and recognised the idempotent
+  close-duplicate; `DjangoStreamStore.append` ignored `options.producer_id`
+  entirely — while its docstring claimed "outcomes, all now matching the
+  in-memory store". Nothing caught it because every producer test routed through
+  `append_with_producer`, but `WritableStore.append` is public and its
+  `AppendOptions` carries those fields, so a consumer calling it directly got
+  adapter-dependent behaviour.
+
+  The whole admission sequence — closed → content-type → producer fencing →
+  Stream-Seq — now lives in one pure module, `rakaia.append_decision`, which both
+  stores ask. The ordering is the subtle part and was previously encoded only as
+  the order of if-statements in two separate methods: fencing runs **before**
+  Stream-Seq, because a retried append carries the same `Stream-Seq` it did the
+  first time, so checking Stream-Seq first would raise `SequenceConflict` on
+  exactly the retry that fencing exists to absorb. Seven new cases in
+  `tests/server_store_contract.py` hold both stores to the same verdicts, and the
+  rules themselves are now unit-testable as a table with no store, no database
+  and no async.
+
 - **A store now refuses an offset it did not issue.** Both stores accepted the
   other's offset format and resolved it to an unrelated position instead of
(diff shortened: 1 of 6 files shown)


joshbrooks:
### Technical detail

**RED first.** The seven new cases in `tests/server_store_contract.py`, run
against both backends on `main`:

```
6 failed, 32 passed, 100 deselected
FAILED test_django_rakaia/...::test_append_honours_producer_fencing
FAILED test_django_rakaia/...::test_append_refuses_a_stale_epoch
FAILED test_django_rakaia/...::test_append_refuses_a_sequence_gap
FAILED test_django_rakaia/...::test_append_accepts_a_well_formed_producer_write
FAILED test_django_rakaia/...::test_append_shares_producer_state_with_append_with_producer
FAILED test_django_rakaia/...::test_append_to_a_closed_stream_reports_the_duplicate_close
```

Every failure is `test_django_rakaia`; the in-memory store passes all seven. That
asymmetry *is* the finding — this is a divergence, not a missing feature, so the
in-memory behaviour is the reference and the durable store moves to it.

Suite 981 → 1015 (14 contract = 7 × 2 stores, plus 20 unit).

**Why it hid.** `server_store_contract.py` had thorough producer coverage, all of
it through `append_with_producer`. A protocol server routes fenced writes there,
so nothing exercised the same options through `append` — which is public, and
whose `AppendOptions` carries `producer_id`/`producer_epoch`/`producer_seq`.

**Shape.** `rakaia/append_decision.py` is pure: `decide_append(facts, opts, *,
producer_state, now) -> AppendVerdict`. `StreamFacts` is the whole surface a
backend must expose (closed, closed_by, content_type, last_seq); `producer_state`
is the one lookup the store must do, since only it knows where that lives — a new
`_producer_state()` on each store, reading `stream.producers` and `StreamProducer`
respectively. Both `append` methods are now gather → ask → persist.

**Ordering, and why it's the reason to centralise.** The relative order of the
four checks was previously implicit in the sequence of if-statements inside two
methods, where an unrelated edit could reorder them with nothing to catch it.
`TestOrdering` pins each pair. The load-bearing one is
`test_fencing_is_checked_before_stream_seq`: a retried append carries the *same*
`Stream-Seq` it did the first time, so checking Stream-Seq first raises
`SequenceConflict` on precisely the retry that fencing exists to absorb. That
comment existed in `store.py` (`"before Stream-Seq check for retry dedup"`) and
was the kind of thing only a comment protected.

**Raise vs return** is preserved and now stated: conflicts (`ContentTypeMismatch`,
`SequenceConflict`) raise because they are caller errors; fencing outcomes return
because they are protocol results with their own statuses and headers.

**Deliberately out of scope:** `append_many` still has its own validation loop.
It has a genuine extra concern — all-or-nothing across a batch, cut at the first
close — and folding it in is a second step best taken once this one has settled.
The same applies to `append_with_producer`, which reaches the same
`validate_producer` but by its own route.

**Gate:** 1015 passed, `ruff check` + `ruff format --check` clean.

PR #97, opened 2026-08-14
Demos could fail and still report success

joshbrooks:
The example projects are how rakaia explains itself: each one seeds some data,
replays it, and then checks that the thing it just claimed actually happened —
no orphaned rows, replay matches the old code, running twice changes nothing.
Seven of those checks printed a red cross when they failed and then exited
successfully anyway. So `just demo` could report a clean run against a library
that had quietly stopped doing what the demo was demonstrating, and you would
only know if you happened to read the scrollback.

Every check now fails the command it is part of. There is also a new `just demos`
that runs all eleven examples and stops at the first failure, and CI runs it —
until now the examples directory was excluded from both the test run and the lint
run, so nothing in it was ever executed or checked automatically, which is how it
drifted far enough to need a repair pass earlier this year. The formatting churn
in the diff is that first lint sweep.


joshbrooks:
### Technical detail

**RED first — by sabotage, since the guard's absence is the defect.** Neutered
`reconcile_children`'s orphan-pruning delete in `src/rakaia/projections.py`, then
ran the same demo with and without the guard:

```
# guard present
EXIT=1
CommandError: reconcile left an orphan: expected 1 activity row, got 2
[3] RECONCILE: IR-108 resubmitted with 1 activity (was 2) -> 2 row(s) — ORPHAN ✗

# guard stashed (i.e. main)
EXIT_WITHOUT_GUARD=0
[3] RECONCILE: IR-108 resubmitted with 1 activity (was 2) -> 2 row(s) — ORPHAN ✗
```

Identical broken output, opposite exit codes. `src/` was restored before
committing — `git diff --stat src/` is empty.

Worth noting the first sabotage I tried (changing the delete op's *name*) raised
a `ValueError` and would have failed on `main` too. That proves nothing about
this change, so I replaced it with one that produces a silently wrong result,
which is the case the guards actually exist for.

**Count correction.** My earlier read of this said 13 unguarded sites. On close
inspection only **7** are silent — in `demo_submissions.py` (×5),
`demo_submission_stream.py` (×1) and `demo_orders.py` (×1). The other six already
raised. The changelog and commit message say 7.

**Placement.** `demo_submission_stream.py` already funnelled every check through a
`_say(ok, msg)` helper, so it needed one change rather than one per call site.
The others raise at each site with a message naming the actual numbers
(`f"replay is not idempotent: {before} -> {after} rows"`), so CI output says what
broke without needing the surrounding narration.

One behavioural note: `_report_parity` had its failure branch marked
`# pragma: no cover - defensive; should not happen`. It is now reachable and
raises — that branch prints the per-field diff first, so the detail survives.

**`just demos` vs `just demo`.** `demo` stays the narrated two-part tour for a
human reader (it is what `docs/whats-new.md` walks through). `demos` is the
regression gate: all eleven, first failure stops it. Both are green locally;
`just demos` takes ~90s.

**CI.** A separate `demos` job rather than a step on `test`, so a demo failure is
distinguishable at a glance from a unit-test failure. It syncs only
`--extra dev --extra django` (no docs/prod needed). `ruff check` and
`ruff format --check` now include `examples/` — that first sweep reformatted 22
files, which is most of the diff. Three `ARG001` suppressions in
`polyglot/signals.py` are Django-mandated receiver signatures.

**Gate:** 981 passed (unchanged — this PR adds no unit tests), `ruff check` and
`ruff format --check` clean across `src/ tests/ examples/`, `just demos` green.

PR #98, opened 2026-08-14
Pin down the store return type that broke our first consumer, and write an upgrade guide

joshbrooks:
Asking a store for a stream used to hand back the raw database record. It now
hands back a plain snapshot of the stream's metadata instead, which is the
correct thing to do — the server reads those fields from async code, and a live
database record would try to fetch itself at exactly the wrong moment. The
problem is that nothing anywhere said the return type had changed, and the shared
test suite both stores must pass only checked that certain field names existed,
which was true of the old shape too. So the change sailed through, and the first
project using rakaia in production broke on it.

That gap is closed: the return type is now declared, and the shared suite checks
it on both stores. There is also a new UPGRADING.md, because this was not the
only thing waiting to bite anyone bumping their pinned version — the package
itself was renamed on PyPI, which breaks the dependency entry even though the
import name never changed, and there are two database migrations to run. Each one
is written up with the exact change to make.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -289,6 +289,18 @@ is not yet tagged in a release.
   there was executed or linted by CI at all — which is how it drifted far enough
   to need a repair pass. `ruff` now covers `examples/` too.
 
+- **`StreamServerStore.get` is typed, and the contract asserts it.** It was
+  declared `-> Any`, and the conformance suite only checked `hasattr` for the six
+  fields a server reads — which a backend's own row object satisfies too. So when
+  `DjangoStreamStore.get` changed from returning its ORM `Stream` row to a
+  `rakaia.types.Stream` metadata snapshot, nothing noticed, and the first
+  downstream consumer broke on it. The snapshot is the right shape (a protocol
+  server is async and reads these fields outside the `run_sync` bridge, so
+  anything lazy would fail there); the mistake was that nothing said so. `get` is
+  now `-> Stream | None` and two contract cases assert the type and its inertness
+  on both backends. A new [`UPGRADING.md`](UPGRADING.md) documents this and the
+  other breaks since `5e4a6e3`, with the exact edits.
+
 - **A store now refuses an offset it did not issue.** Both stores accepted the
   other's offset format and resolved it to an unrelated position instead of
   failing: `int("0_5")` is `5` in Python, so the durable store read the wrong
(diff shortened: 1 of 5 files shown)


joshbrooks:
### Technical detail

**RED by reverting.** The new cases pass on `main` — the current `get()` is
already correct — so the proof is that they would have *caught* the change.
Temporarily restored the pre-#88 implementation
(`return Stream.objects.filter(stream_id=path).first()`):

```
2 failed, 2 passed, 60 deselected
FAILED test_get_returns_the_shared_stream_type
FAILED test_get_does_not_leak_a_backend_row
E  AttributeError: 'Stream' object has no attribute 'path'
```

The two that stayed green are the pre-existing `test_get_exposes_what_a_server_reads`
(`hasattr` for six fields) and `test_get_is_none_for_a_missing_stream`. That
split is the finding: the ORM row has `content_type`, `ttl_seconds`,
`expires_at`, `last_seq`, `closed` and a `current_offset` property, so it
satisfied the old contract completely. `src/` was restored before committing.

**Two cases, not one.** `test_get_returns_the_shared_stream_type` is the
`isinstance` check. `test_get_does_not_leak_a_backend_row` reads every field a
server touches and asserts that needs no live connection — because *inertness*,
not the class name, is the property the async server actually depends on. A
future backend could satisfy `isinstance` with a lazy subclass and still break.

**Typing.** `-> Any` becomes `-> StreamMetadata | None` (`rakaia.types.Stream`,
aliased at import to avoid colliding with `django_rakaia.models.Stream` in
readers' heads). Pyright is unchanged at 46 pre-existing errors before and after.

**The fix consumers need is not an API change.** `store.get(path)` was never the
right way to reach the ORM row; it was just what happened to come back. The
replacement is `StreamEntry.objects.filter(stream__stream_id=path)` — one query
instead of two, and it does not involve the store. `UPGRADING.md` says so rather
than adding a compatibility shim, because a shim would re-establish the coupling
this PR is pinning shut.

**UPGRADING.md scope.** It documents the full set of breaks between `5e4a6e3`
(the revision the current consumer pins) and `main`:

1. the `rakaia` -> `rakaia-streams` distribution rename, which breaks a
   `[tool.uv.sources]` entry even for a git pin, since uv matches the requirement
   name against the name the source builds;
2. this `get()` change;
3. migrations `0006` and `0007` — `0007` verified to carry a `RunPython`
   seeding `last_activity_at`, so it touches every `Stream` row.

Plus a "not a break despite looking like one" section (named failures still
subclass the builtins they replaced, so `except KeyError` holds) and a
behaviour-change list. **That last list describes PRs #90, #91, #93, #95 and #96,
which are still open** — if any is rejected or reshaped, the corresponding bullet
needs editing before this merges.

It lives at the repo root alongside `CHANGELOG.md` rather than in `docs/`,
matching where the changelog sits; the README links to it from the install
section.

**Gate:** 985 passed (981 + 4 = 2 cases × 2 stores), `ruff check` +
`ruff format --check` clean on `src/ tests/`.

PR #99, opened 2026-08-14
The registry kept the same bookkeeping in three places, keyed by tuple position

joshbrooks:
When you register a handler, rakaia writes a note about it to a log so that a
fresh process can find and reload it later. There are three kinds of thing you
can register, and each one had its own copy of that bookkeeping — creating the
log, reading it back, deciding whether an entry is new. Alongside them sat six
small functions in matched pairs: one to describe a registration, one to
reconstruct that description from what was written down. The two halves of each
pair had to stay in step by hand, and nothing checked that they did.

They were also read back by position rather than by name, so the code that
reloads modules asked for "field five" of one kind of record and "field two" of
another. Two of those functions carried comments telling whoever edited them next
to only ever add fields at the end, or the reloading would break. Adding a single
field to a registration meant touching four functions and re-verifying two
position comments in a fifth.

That bookkeeping now exists once, each kind of record describes itself, and
fields are looked up by name. Nothing behaves differently — but the log can
finally be tested directly, without registering real handlers or importing
anything, which is where the twenty-two new tests come from.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,24 @@ is not yet tagged in a release.
   `create()` that truncated. With it pinned, the `has()`/`create()` dance is
   provably redundant — `registry.py`'s three copies are now a single `create()`.
 
+- **One registration log instead of three.** The meta-stream mechanism —
+  create the stream, read back what is recorded, append if new — was written
+  three times in `registry.py` (handlers, reducers, upcasters), alongside six
+  module-level identity functions in hand-mirrored pairs: one building a tuple
+  from the object, one rebuilding the same tuple from stored JSON, with nothing
+  checking that they agreed. The tuples were then read back **positionally**
+  (`ident[4]` for a handler's dotted path, `ident[2]` for a reducer's), and two
+  of the builders carried comments warning editors to append new fields at the
+  end so those indices stayed valid.
+
+  `rakaia.registration_log.RegistrationLog` owns the mechanism once, and each
+  record type owns its own `identity` / `to_payload` / `identity_from_payload` —
+  so the round trip is a property of the type rather than an agreement between
+  two functions, and `modules()` reads `dotted_path` by name. No behaviour
+  change; `registry.py` loses 89 lines and every positional index lookup, and
+  the log is now testable against a bare in-memory store with no decorators,
+  no source hashing and no import machinery.
+
 - **`PreloadedProjectionReader` — bulk-fetch a verification sweep.**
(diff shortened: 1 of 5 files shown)


joshbrooks:
### Technical detail

**RED first.** `tests/test_rakaia/test_registration_log.py` on `main`:

```
E  ModuleNotFoundError: No module named 'rakaia.registration_log'
```

22 pass after. Suite 981 → 1003. `registry.py` 1180 → 1091 lines; the new module
is 157.

**Behaviour is unchanged** — this is a pure refactor, and the existing
persistence tests (`test_registry_persistence.py`, `test_upcasters.py`,
`test_registry_isolation.py`) pass untouched apart from two that reached into
`reg._persisted_ids` directly and now read `reg._handler_log.known()`.

**Where the depth comes from.** The mechanism (create / load / dedup / append)
moves into `RegistrationLog`; the *knowledge* (what identifies a record, how it
serialises) moves onto the three dataclasses as `identity` / `to_payload` /
`identity_from_payload`. That split is what removes the drift risk: the round
trip is now a property of one type rather than an agreement between
`_identity()` and `_identity_from_payload()` sitting 40 lines apart.
`TestIdentityRoundTrip` asserts it through real JSON — a dict round trip would
not catch the `frozenset` case, which is the one that actually needs the
special handling.

**Positional indices are gone.** `grep -c "ident\[" src/rakaia/registry.py`
returns 0. `modules()` reads `payload["dotted_path"]` by name, so the two
"append new fields at the end so `ident[4]` stays valid" comments are deleted
rather than relocated.

**Two subtleties preserved and now pinned**, both previously only guaranteed by
comments:

- a `frozenset` event match serialises **sorted**, so the meta-stream does not
  change with set iteration order and re-append the same handler on every
  restart (`test_a_set_event_match_serializes_stably`);
- `identity_from_payload` tolerates payloads written before `match_field` and
  `stage` existed — meta-streams already in production have to keep loading
  (`TestBackwardCompatibilityWithOlderPayloads`).

**`store=None` is handled inside the log** rather than at each call site, which
removes four `if self._store is not None:` guards from the registries.

**Not folded in:** the identity functions had non-persistence callers too — the
in-memory dedup in `register()` compared `_identity(existing) == _identity(new)`.
Those now use the `.identity` property, which is the same value; I did not try to
route in-memory dedup through the log as well, since the log is about what has
been *written down* and the in-memory check is about what is currently
registered. Merging them would conflate two different questions.

**Gate:** 1003 passed, `ruff check` + `ruff format --check` clean.

Issue #100, opened 2026-08-15
fold_events defaults its scratch stream path to a consumer-specific name


joshbrooks:
`django_rakaia.envelope.SCRATCH_PATH` is `"produce/submission"` (src/django_rakaia/envelope.py:43), and `fold_events` uses it as the default `scratch_path`.

`django_rakaia` is the generic Django integration, but `produce/submission` is domain language from the first consumer. Two consequences:

- A reader of `fold_events` cannot tell whether the path is meaningful or arbitrary. It is arbitrary — the store is in-memory and thrown away per call — but the name implies a convention that does not exist.
- The path is load-bearing for callers: a registry's `event_match` has to name it for handlers to fire. So a consumer whose streams are not `produce/*` must either pass `scratch_path` at every call site or register handlers against another project's vocabulary.

Options, roughly in order of preference:

1. Rename the constant's value to something neutral and self-describing (`"_scratch/fold"`), and note in the docstring that it is arbitrary but must match the registry's `event_match`.
2. Make `scratch_path` required, so the coupling to `event_match` is explicit at the call site.

Not urgent — it is a default, and overridable. Raised while reviewing #94.

joshbrooks:
Fixed in #102 — `SCRATCH_PATH` is now `"_scratch/fold"`, and the docstrings say the value is arbitrary but must match the registry's `event_match`.

PR #102, opened 2026-08-15
A safety guard was blaming the rebuild for its own query

### LINKED ISSUE #101: assert_no_live_writes must nest outside deny_database_access, and nothing enforces it
`assert_no_live_writes` counts rows on the guarded alias in its `finally` block (src/django_rakaia/hermeticity.py). `deny_database_access` installs an `execute_wrapper` that raises `AmbientDatabaseAccess` on any query to that alias.

So this works:

```python
with assert_no_live_writes(FinanceLine):
    with deny_database_access("default"):
        replay(...)
```

and this fails confusingly — the closing count query trips the read guard, and the `AmbientDatabaseAccess` it raises comes from the guard's own bookkeeping rather than from anything the rebuild did:

```python
with deny_database_access("default"):
    with assert_no_live_writes(FinanceLine):
        replay(...)
```

Every docstring and test in #92 uses the correct nesting, so this is a latent footgun rather than a live bug. Worth either detecting it (have `assert_no_live_writes` take its counts through a path the wrapper ignores, or check for an active deny and raise a clear error) or stating the ordering constraint explicitly in `assert_no_live_writes`'s docstring, where someone writing the call will actually read it.

Raised while reviewing #92.

joshbrooks:
Rakaia ships two guards for proving that rebuilding your tables from the event
log doesn't touch the live database — one watches for reads, one watches for
writes. Used together in one order they work; used in the other order the
write-guard's own row count sets off the read-guard, and the resulting error says
the rebuild touched the live database. It didn't. The guard did, while checking.
For a tool whose entire purpose is to tell you where a leak is, pointing at one
that doesn't exist is about the worst failure it can have — and every example we
shipped happened to use the order that works, so nobody would find out until
they wrote it the other way round.

The order no longer matters. The write guard now takes its counts with rakaia's
own read-guard stood down, on the grounds that counting rows to check for a leak
isn't the rebuild reading live data. A genuine leak is still caught either way,
and if you have your own query logging installed it still sees everything.

Also in here, from the same review: the helper that replays a batch of events was
defaulting to a stream name borrowed from the first project that used rakaia.
It's an arbitrary name, but not an unimportant one — your handlers have to be
registered against it — so every other project was quietly adopting a stranger's
naming. It's now a neutral one, and the docs say what it's for. Closes #101 and
#100.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -205,6 +205,28 @@ is not yet tagged in a release.
 
 ### Fixed
 
+- **The two rebuild guards compose in either nesting order** (#101).
+  `assert_no_live_writes` counts rows on the alias `deny_database_access`
+  forbids reading, so nesting them the "wrong" way round made the closing
+  `COUNT(*)` trip the read guard — and the `AmbientDatabaseAccess` it raised
+  said *"replay touched the 'default' database directly"*, blaming the rebuild
+  for a query the guard itself issued. In a tool whose only job is to say where
+  a leak is, that sends the reader after one that does not exist. The guard now
+  suspends rakaia's own deny wrappers while taking its counts — checking for a
+  leak is bookkeeping, not the rebuild reading live data. A consumer's own
+  `execute_wrapper` is left armed, and a genuine leak is still caught in either
+  order.
+
+- **`fold_events` no longer defaults to another project's vocabulary** (#100).
+  `SCRATCH_PATH` was `"produce/submission"` — domain language from the first
+  consumer, sitting in the generic Django integration. The value is arbitrary
+  (the store is in-memory and discarded per call) but *load-bearing*, because a
+  registry's `event_match` has to name it, so every other consumer was
+  registering handlers against a stranger's stream naming. It is now
+  `"_scratch/fold"`, namespaced so it cannot collide with a consumer's paths,
+  and the docstrings say plainly that the value is arbitrary but must match
+  `event_match`.
+
 - **`RAKAIA_STORE` no longer fails open.** `get_store()` returned the
   *in-memory* store for any backend string that wasn't exactly `"durable"`, so a
   one-character typo (`"durrable"`, a stray space from a `.env` file, `"Durable"`)
--- a/src/django_rakaia/envelope.py
+++ b/src/django_rakaia/envelope.py
@@ -36,11 +36,20 @@ from rakaia.protocols import ProjectionReader, WritableStore
 from rakaia.registry import HandlerRegistry
 from rakaia.replay import replay
 
-#: The in-memory stream path a live fold replays through. A fold is not a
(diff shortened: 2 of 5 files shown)


joshbrooks:
### Technical detail

**RED first.** Both findings reproduce; the #101 one is the interesting output:

```
FAILED ...::TestTheTwoGuardsComposeInEitherOrder::test_read_guard_outside_write_guard
E  AmbientDatabaseAccess: replay touched the 'default' database directly
   (SQL: SELECT COUNT(*) AS "__count" FROM "test_django_rakaia_financeline").
   A hermetic rebuild reads only through the injected reader/executor alias …
FAILED ...::test_an_unrelated_execute_wrapper_is_not_disturbed
```

The message names the guard's own `COUNT(*)` while asserting the *rebuild* did
it. That is the defect — not the raise itself.

**Fix shape (#101).** `deny_database_access`'s blocker is tagged
`_wrap._rakaia_deny_guard = True`, and `_without_deny_guards(alias)` filters
*only* tagged wrappers out of `connection.execute_wrappers` for the duration of
the count, restoring the original list in `finally`.

Chose suspension over the issue's other two options deliberately:

- *Detect the nesting and raise a clearer error* — still fails, just legibly. The
  two orders are equally reasonable to write; making one of them an error is a
  worse outcome than making both work.
- *Document the constraint* — puts the burden on someone reading a docstring at
  the moment they are debugging a rebuild.

**Why filter rather than clear.** Blanket-clearing `execute_wrappers` would also
suspend a consumer's query logger or timing wrapper for those two queries — a
surprise we have no business causing.
`test_an_unrelated_execute_wrapper_is_not_disturbed` installs a spy and asserts
it still sees the `COUNT`.

**The obvious risk with this fix is blinding the guard**, so two tests exist for
exactly that: `test_a_real_leak_is_still_caught_in_the_hard_order` (leak detected
with the read guard outermost) and `test_the_read_guard_still_catches_a_real_ambient_read`
(the deny wrapper is restored and still armed afterwards).

**#100** is a value change: `"produce/submission"` → `"_scratch/fold"`. Taken
option 1 from the issue rather than option 2 (making `scratch_path` required),
because requiring it at every call site would make the common case noisier
without removing the coupling — the caller still has to match `event_match`
either way. Instead the docstrings state that the value is arbitrary but must
match, and recommend both sides name the constant rather than repeat a string.
Three tests pin it, including `test_a_registry_matching_a_different_path_projects_nothing`,
which makes the failure mode explicit: a mismatch is **silent**, projecting
nothing.

No consumer is on `SCRATCH_PATH` yet — it shipped in #94, one day ago, and
partisipa's bump predates it.

**Gate:** 1125 passed, `ruff check` + `ruff format --check` clean on
`src/ tests/ examples/`.

PR #103, opened 2026-08-15
Timestamps looked different on every replay, forever

### LINKED ISSUE #83: canonical_value has no datetime normalizer: payload timestamps never compare equal to a DateTimeField
## Problem

`canonical_value` (`src/django_rakaia/verification.py:189`) is the single seam where a value lifted out of `event.data` meets the column it is being written to. It applies `DEFAULT_NORMALIZERS`, which is:

```python
DEFAULT_NORMALIZERS: tuple[Normalizer, ...] = (normalize_uuid, normalize_decimal)
```
(`verification.py:186`)

Both return `value` untouched unless the field is a `UUIDField` / `DecimalField`. **There is no datetime normalizer**, and no field-type-aware string→datetime coercion.

`StreamEvent.data` is JSON, so a payload timestamp is always a *string*. A `DateTimeField` column always holds a *`datetime`*. So at the compare in `effect_executor.py:151-154`:

```python
changed = {
    k: v
    for k, v in defaults.items()
    if canonical_value(model, k, getattr(row, k))   # column: datetime
    != canonical_value(model, k, v)                 # payload: str
}
```

the comparison reduces to `datetime(...) != "2026-08-09T12:30:00.123Z"` — **always `True`, at any precision**.

`diff_effects_against_rows` uses the same helper via `_diff_row` (`verification.py:389`), and `canonical_value`'s docstring (`verification.py:198-203`) commits to the diff and the write path defining "unchanged" identically, so both callers are affected the same way.

## Impact

Quiet, and not data corruption:

- `skip_unchanged` decides "changed" on **every** replay for that field, issuing a redundant `UPDATE` per event — exactly the churn `_upsert_skip_unchanged` exists to prevent.
- `diff_effects_against_rows` reports a permanent phantom diff on that field that no amount of replaying will clear.

The write itself still succeeds — Django's `DateTimeField.get_prep_value` parses the string on save.

## Reachability: latent today

No shipped code path hits this. Audited on `worktree-issue-80-consumer-findings`:

- Every `DateTimeField` in the repo is Django- or manager-managed, never written from `effect.defaults`: `models.py:39`, `:150`, `:201` (`auto_now_add`), `models.py:365` (`auto_now`), `models.py:328` `Translatable.deleted` (soft-delete marker), `examples/chat/chat/models.py:26,66`, `examples/formkit_submissions/submission_stream/models.py:34`.
- Every projection column that receives a payload timestamp is a `CharField`, where both sides are strings from the same encoder and compare fine: `examples/partisipa_history/history/models.py:26,47`, `examples/partisipa_merge/subproject/models.py:70`, `examples/formkit_submissions/submissions/models.py:33` (copied verbatim by `visit_defaults`, `mapping.py:71`), `tests/test_django_rakaia/models.py:300,311`.

It fires the first time a handler targets a `DateTimeField` — a natural thing to do, with nothing warning against it.

## Reproduction

Add a `DateTimeField` to a projection model, have a handler write an ISO string from the payload into it via `update_or_create` effect `defaults`, and replay twice with `skip_unchanged`. The second pass issues an `UPDATE` it should have skipped.

## Design note: both sides must truncate

The obvious fix — parse the payload string to a `datetime` and compare — **is not sufficient on its own**. `StreamEvent.data` is encoded with `DjangoJSONEncoder`, which truncates (does not round) datetimes to millisecond precision and emits no fractional part at all when `microsecond` is 0. See the table under "Payload types" in `docs/django-integration.md` and the `data` docstring at `models.py:136`.

Django stores microseconds. So a parsed payload value still won't equal the column exactly — the normalizer has to truncate **both** sides to milliseconds to agree.

That decision is load-bearing rather than cosmetic, and it needs to be made deliberately:

- Truncating both sides means a genuine sub-millisecond change is reported as "unchanged" and silently skipped.
- Not truncating means the phantom diff persists for any value with non-zero microseconds — i.e. almost all of them.
- Whatever is chosen must hold for `diff_effects_against_rows` and `_upsert_skip_unchanged` simultaneously, per the `canonical_value` contract.

`DateField` and `TimeField` want the same treatment (`DjangoJSONEncoder` truncates `time` to milliseconds too, via `r[:12]`).

## Acceptance

- A `normalize_datetime` added to `DEFAULT_NORMALIZERS`, resolving the field via `_resolve_field` the way `normalize_decimal` does.
- A payload ISO string compares equal to the `DateTimeField` value it was encoded from; `skip_unchanged` skips the `UPDATE` on a re-replay.
- The truncation semantics are documented on `DEFAULT_NORMALIZERS` and pinned by a test, including the `microsecond == 0` (no fractional part) shape.
- `DateField` / `TimeField` covered or explicitly scoped out.

Found while reviewing #81.

joshbrooks:
Rakaia compares what a replay would write against what your tables already hold,
to check that replaying the log reproduces your data. Any column holding a date
or a time always failed that comparison — not sometimes, always. When an event is
recorded, timestamps get rounded down to the nearest millisecond, but the
database keeps microseconds, so the two never matched. Dates were worse still:
the recorded form is text and the column hands back a date, which can't be equal
whatever the value.

The visible cost is a verification report that always shows differences, which
teaches you to stop reading it. The expensive cost is quieter: the same
comparison decides whether a replay can skip rewriting a row it hasn't changed.
So every replay rewrote every row with a timestamp — bumping "last modified"
fields, re-firing whatever listens for saves, and re-replicating it all, for
values nobody touched.

Both sides are now rounded down to milliseconds before comparing, which is the
only way to make them meet — reading the recorded value back isn't enough on its
own. The trade-off is that a change smaller than a millisecond now counts as no
change; that's deliberate, since a difference that small can't be recorded in the
log to begin with. Stored values keep their full precision. Closes #83.


joshbrooks:
### Technical detail

**RED first.** `tests/test_django_rakaia/test_verification_temporal.py` on `main`:

```
9 failed, 7 passed
FAILED ...::TestCanonicalValueMakesThemAgree::test_an_encoded_datetime_equals_the_column_value
FAILED ...::test_a_whole_second_datetime_equals_the_column_value
FAILED ...::test_an_encoded_date_equals_the_column_value
FAILED ...::test_an_encoded_time_equals_the_column_value
FAILED ...::TestTheEndToEndSymptom::test_skip_unchanged_does_not_rewrite_an_unchanged_row
E  datetime.datetime(2026, 1, 2, 3, 4, 5, 123000, tzinfo=utc) = <Measure>.observed_at
```

The 7 that passed include `TestTheEncoderLosesPrecision`, which characterizes the
premise rather than the bug. Suite 981 → 1133 overall.

**Encoder behaviour, verified rather than assumed:**

```
datetime(…, 123456) -> "2026-01-02T03:04:05.123Z"   # truncated to ms
datetime(…, 0)      -> "2026-01-02T03:04:05Z"       # no fractional part at all
datetime(…, 999999) -> "…05.999Z"                   # truncates, does not round
date(2026,1,2)      -> "2026-01-02"
time(3,4,5,123456)  -> "03:04:05.123"
```

The `999999` case is why the normalizer truncates rather than rounds — rounding
would disagree with the encoder on half of all values, swapping one phantom diff
for another.

**Both sides truncate**, as the issue requires. Parsing the payload alone leaves
`…05.123` against a stored `…05.123456`, so the column value has to come down to
meet it. `_truncate_to_encoder_precision` does integer division by 1000µs.

**Ordering matters and is pinned.** `DateTimeField` subclasses `DateField` in
Django, so a normalizer testing `DateField` first would truncate every timestamp
to its calendar date. `test_a_datetime_field_is_not_treated_as_a_date_field`
catches that.

**Deliberate non-goals, each with a test:**

- an unparseable string is returned unchanged
  (`test_an_unparseable_string_is_left_alone`) — reporting a difference beats
  silently swallowing a value we failed to interpret;
- the normalizer never writes truncated values back
  (`test_skip_unchanged_does_not_rewrite_an_unchanged_row` asserts
  `row.observed_at == AWARE` with full microseconds);
- sub-millisecond changes now read as unchanged — `TestTheAcceptedCost` pins it
  so the trade-off is a recorded decision rather than a discovery.

**Timezones** need no special handling: the encoder emits `Z` for `+00:00` and
`parse_datetime` restores UTC, so both sides derive from the same original under
the same settings. Mixing naive and aware would compare unequal, which is a real
difference worth reporting rather than one to paper over.

**Test fixtures:** `Measure` gains `observed_at` / `observed_on` /
`observed_time` (all nullable) plus migration `0008` — the test app had no
temporal column at all, which is precisely why this went unnoticed. The issue's
survey noted every existing timestamp projection uses `CharField`, where both
sides are strings from the same encoder and compare fine.

**Gate:** 1133 passed, `ruff check` + `ruff format --check` clean on
`src/ tests/ examples/`.

PR #104, opened 2026-08-15
The producer error table was written out three times

joshbrooks:
When a writer sends an event and rakaia refuses it — because it already arrived,
because another writer has taken over, because a number is out of sequence, or
because the stream is closed — each case turns into a particular HTTP status with
particular headers. That table was spelled out three separate times inside the
same function, which is already the longest one in the codebase. Two of the
copies agreed; the third differed, and it took a careful read to work out whether
those differences were intentional or just drift.

We'd already fixed exactly this problem for the other half of the same
question — errors thrown by the storage layer used to be identified by matching
English text and now go through one small lookup. This brings the writer-refusal
half into the same shape. The function it lived in loses about a third of its
length, and the table can now be checked directly rather than by starting a
server, opening a stream, and driving it into each failure in turn.

No behaviour change intended; the existing protocol conformance suite and the
full test suite both pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -203,6 +203,20 @@ is not yet tagged in a release.
   recipe** that runs the scripted demos end-to-end.
   → [`docs/whats-new.md`](docs/whats-new.md).
 
+### Changed
+
+- **One producer response table instead of three.** The five-arm `isinstance`
+  ladder over `ProducerValidationResult` — duplicate → 204, stale epoch → 403,
+  bad epoch opening → 400, sequence gap → 409, closed stream → 409, each with
+  its own headers — was written out three times inside `_handle_append`, the
+  largest function in the largest module. The exception half of the same mapping
+  had already been deepened into a single `_status_for` lookup; the union-typed
+  half never followed. `producer_response(result, *, producer_epoch, offset,
+  stream_closed)` now sits beside `_status_for`, and the three call sites pass
+  what differs rather than restating the table. `_handle_append` drops from 244
+  to 164 lines, and the status/header mapping is testable as a table — no ASGI
+  round trip, no live store.
+
 ### Fixed
 
 - **`RAKAIA_STORE` no longer fails open.** `get_store()` returned the
--- a/src/rakaia/handler.py
+++ b/src/rakaia/handler.py
@@ -55,6 +55,7 @@ from .types import (
     EmptyJsonArray,
     InvalidJson,
     InvalidOffset,
+    ProducerAccepted,
     ProducerDuplicate,
     ProducerInvalidEpochSeq,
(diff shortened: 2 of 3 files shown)


joshbrooks:
### Technical detail

**RED first.** `tests/test_rakaia/test_producer_response.py` on `main`:

```
E  ImportError: cannot import name 'producer_response' from 'rakaia.handler'
```

17 pass after. Suite 1133 → 1134 net (the ladder's own coverage was already
there via the ASGI tests, which is the point — they still pass unchanged).

**Measured, not asserted:** `_handle_append` goes **244 → 164 lines**. Note the
*file* grows by 12 lines, because `producer_response` carries the docstring the
three ladders never had anywhere. The win is locality, not line count, and I'd
rather say that than quote a misleading total.

**What was actually duplicated vs. what differed.** Three of five arms (stale
epoch, invalid epoch opening, sequence gap) were byte-identical between the
close-path and append-result copies. The differences were all in the duplicate
arm — which offset to report (`close_result.final_offset` vs. a fetched
`store.get(path).current_offset` vs. none at all) and whether to set
`Stream-Closed`. Those are now the two keyword arguments, so the variation is
visible at the call site instead of being buried in a re-typed branch.

**One `isinstance(pr, …)` survives, deliberately.** In the close path,
`ProducerStreamClosed` is the only arm needing the stream's current offset, and
fetching it is a round trip. That check decides *whether to spend a query*, not
what the response is — commented as such.

**One reachable-behaviour question, and my answer.** The closed-stream branch
previously fell through to a bare 409 for anything that was not a
`ProducerDuplicate`, including `pr is None`. That `None` case is preserved
explicitly (it synthesises a `ProducerStreamClosed`). A `ProducerStaleEpoch`
arriving with `stream_closed=True` would now get 403 rather than 409 — but that
combination cannot occur: `decide_append` returns `stream_closed=False` for every
fencing refusal, so `stream_closed=True` only arrives with `None` or
`ProducerDuplicate`. Flagged here rather than left for a reviewer to spot.

**Closed-set guard.** `test_no_refusal_type_falls_through` enumerates every
`Producer*` refusal type in `rakaia.types` and asserts each has an arm — so a new
result type fails this test instead of silently 500ing at runtime. That mirrors
what `_status_for`'s closed-set test already does for the exception half.

**Header constants**, not literals: `PRODUCER_EPOCH_HEADER` and friends are
lowercase in `rakaia.types` (`"producer-epoch"`), which my first draft of the
tests got wrong by asserting title case. The tests now import the constants, so
they cannot drift from the implementation.

**Gate:** 1134 passed, `ruff check` + `ruff format --check` clean on
`src/ tests/ examples/`; conformance suite unaffected.

PR #105, opened 2026-08-15
Two ways to replay, one set of staging rules written twice

joshbrooks:
Rakaia can rebuild your tables by replaying one stream in order, or by merging
several streams together and replaying them as one. Those really are different
jobs, and they should be. But everything that happens afterwards — running each
event through your handlers in stages, making sure every event clears stage one
before stage two starts, running the summary steps, counting what was
processed — was written out separately for each of them, in two shapes that had
already stopped resembling each other. Changing how staging works meant making
the same change twice and getting it right both times.

The shared part now lives in one place, and the two entry points just supply the
events. Both functions get noticeably shorter, and the staging rules can finally
be tested against a plain list of events rather than needing a real stream with
real encoded data behind it — which is where the new tests come from.

One thing deliberately left alone: replaying a single stream still decodes events
one at a time rather than all up front. That isn't a speed tweak. It means a
corrupt event halfway through fails after the earlier ones have been applied,
which is how replay has always behaved, and changing it would quietly move where
a failure leaves you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -217,6 +217,26 @@ is not yet tagged in a release.
   to 164 lines, and the status/header mapping is testable as a table — no ASGI
   round trip, no live store.
 
+- **`replay` and `merge_replay` share one staging pipeline.** Reading one stream
+  in order and k-way-merging several by an order key are genuinely different
+  jobs — but everything *after* that was written twice: registry defaulting, a
+  byte-identical `_upcaster_drift` closure, a `_ReplayCtx` with the same seven
+  fields, the `any(stage > 0) or has_reducers()` test, the reader-required
+  check, the stage-pass loop, and `events_processed`. Around sixty lines, in two
+  control-flow shapes that had already drifted apart, so any change to staging
+  semantics had to be made twice and correctly in both.
+
+  `build_pipeline()` / `run_passes()` now own that; the two entry points are
+  event *sources* that hand over `(seq, match_str, event)` tuples. `replay`
+  drops 129 → 102 lines and `merge_replay` 144 → 119. Staging is now testable on
+  a fabricated list of events — no store, no bytes, no upcasters — which is what
+  the new tests use.
+
+  `replay`'s single-stage path deliberately keeps its own streaming loop rather
+  than going through `run_passes`: decoding lazily means a malformed event at
+  offset N fails *after* the first N-1 have been dispatched, which is the
+  partial-progress behaviour replay had before staging existed.
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


joshbrooks:
### Technical detail

**RED first.** `tests/test_rakaia/test_replay_pipeline.py` on `main`:

```
E  ImportError: cannot import name 'build_pipeline' from 'rakaia.replay'
```

10 pass after. Suite 1126 → 1127 net; every existing replay/merge test passes
unchanged, which is the check that matters for a no-behaviour-change refactor.

**Measured:** `replay` 129 → 102 lines, `merge_replay` 144 → 119. The *file*
grows 605 → 662, because `build_pipeline` / `is_staged` / `require_reader` /
`run_passes` carry docstrings the duplicated code never had. Locality is the win,
not line count.

**The seam is `(seq, match_str, event)`.** That triple is exactly what differs
between the two callers and nothing more: `replay` numbers by offset in one
stream, `merge_replay` by position in the merged order. The pipeline never learns
which it is, which is what makes a third entry point (an incremental replay from
a cursor, say) cost an event source rather than a re-implementation.

**One thing deliberately NOT unified.** `replay`'s single-stage path keeps its
own streaming loop instead of calling `run_passes`. The existing comment called
this "bounded memory", which undersells it — `store.read()` has already
materialised the messages, so the real property is *partial progress*: decoding
lazily means a malformed event at offset N fails after the first N-1 have been
dispatched. Materialising first moves that failure earlier and changes what a
half-completed replay leaves behind. The comment now says that.

**Behaviour preserved in two places worth naming:**

- `merge_replay` used `passes = stages if staged else [None]`, i.e. it called
  `_run_stage_reducers(ctx, None)` even unstaged, where `replay` skipped it.
  Equivalent, because `staged` is false only when no reducers exist. `run_passes`
  keeps the single-`None`-pass shape, so the branch disappears rather than
  needing a special case.
- The reader-required message differed between the two ("Replay has…" vs
  "merge_replay has…"). `what=` keeps each caller named, pinned by
  `test_the_message_names_the_caller` — otherwise the guidance would point at the
  wrong function.

**A bug in my own test, worth flagging** because it demonstrates a real
mechanism: I first wrote the reducer as `lambda _reader, _s=stage: …` to dodge
late binding. That second *positional* parameter is precisely how a reducer opts
in to the touched-subjects argument, so it was handed the `TouchedSubject` tuple
instead of the stage. Replaced with a factory closure, and commented so the next
person doesn't repeat it.

**Gate:** 1127 passed, `ruff check` + `ruff format --check` clean on
`src/ tests/ examples/`.

PR #106, opened 2026-08-15
Everyone installing rakaia got a translations table they never asked for

joshbrooks:
Rakaia's Django package shipped a small translations feature — a model for
storing translated strings, an admin page, a web dashboard, and some API
endpoints. It's a nice demo of live updates, but it isn't what the library is
for, and because the model was declared in the very first migration, every single
project that installed rakaia got a translations table created in their database
whether they wanted one or not. The giveaway is that its language options were
hard-coded to Tetum, Portuguese and Indonesian.

It now lives with the polyglot example, which is the only thing that was ever
using it. That example already had a comment explaining why it deliberately
didn't push its own concerns into the library — this finishes the thought.

Nothing in the streams code referenced any of it, so this is straight
subtraction rather than untangling: the models file loses 135 lines, views 213,
and so on. A new migration drops the table, and the upgrade guide has the model
definition ready to paste into your own app if you were relying on it, along with
the command to export your rows first. The stream dashboard and everything else
under `/streams/` are untouched.


joshbrooks:
### Technical detail

**Investigated before touching anything**, since this deletes a documented
public API. The short version: cleanly separable, and less entangled than
expected.

- **Zero inbound dependencies.** No core module imports anything from the
  feature; `src/rakaia/` has no references at all; `apps.py` has none.
- **The one shared helper goes the right way.** `_sanitize_group_name` is
  core-owned and core-consumed (`broadcast_entries`, `stream_events_sse`); the
  translations receiver hard-coded `"translations"` and never called it. The core
  SSE path (`StreamEntry` `post_save` → `broadcast_entries` → `_frame`) is an
  entirely separate mechanism from the removed `StreamEvent` receiver.
- **No test used `Translatable` as a fixture for core behaviour.** That was the
  stated risk and it does not materialise — every core stream test builds its own
  `Stream`/`StreamEvent`/`StreamEntry` fixtures. Confirmed by grep: only
  `test_admin.py` and `test_views.py` import it, and `test_channels.py` never
  does (its three translations tests use a `translatable_id` key in a payload
  dict).
- **Downstream is clean.** `git grep` across `partisipa-import` on both `main`
  and `feat/durable-store` finds no use of `Translatable`, `translatable_id`,
  `django_rakaia.views`, `django_rakaia.urls` or `django_rakaia.admin`. It has
  its own unrelated `translate.Translatable`, which is further evidence the
  library's version served nobody.

**What actually broke, and was fixed:** `examples/polyglot` imported
`django_rakaia.models.Translatable` (views.py:19, signals.py:23) and owned **no**
migrations — it relied entirely on the library's `0001_initial` to create its
table. So the model had to move *into* polyglot with its own `0001_initial`, not
merely be deleted. Verified end to end on a fresh database: migrate, create a
`Translatable`, and the `post_save` fan-out writes 1 `StreamEvent` / 1
`StreamEntry`, with `Translatable.objects.gettext(...)` returning the translation.

**Migration.** `0008` is a real `DeleteModel` — it drops the table. Chosen over
`SeparateDatabaseAndState` because leaving an orphan table nobody manages is
worse than an announced drop, and the one known deployment's table is empty. The
migration carries a docstring saying so, and `UPGRADING.md` has the `dumpdata`
step plus the model definition to paste into your own app. Note polyglot's table
is `polyglot_translatable` — a *different* table — so a dump needs its app label
changed; also documented.

**Two smaller things I'd have missed without looking:**

- `_format_sse_event` in `channels_views.py` was used *only* by the translations
  HTML SSE view, so it went too; `render_to_string` with it.
- Removing the `translations/` route changes routing: a stream literally named
  `translations` now resolves to `stream_detail`, because that path sat *before*
  `<str:stream_id>/`. An improvement, but a behaviour change — called out in
  `UPGRADING.md` and the changelog.

**Deliberately dropped rather than moved:** the HTMX dashboard, its templates and
the `"translations"` channel group. Polyglot has its own UI, its own per-langcode
streams (`translations:{langcode}`) and its own templates — the two were separate
non-overlapping demos, so folding both in would have left polyglot with two
competing UIs to reconcile. `test_stream_event_without_translatable_does_not_broadcast`
was deleted outright rather than kept, since it becomes a vacuous negative
control once the receiver is gone.

**Gate:** 1105 passed (1126 − 21 translations tests), `ruff check` +
`ruff format --check` clean, `zensical build` clean, and all 11 demos pass
including `polyglot`.

PR #107, opened 2026-08-15
Make the demos prove the new guarantees, not just describe them

joshbrooks:
The recent batch of work added a few guarantees — a verification sweep that
refuses to certify itself when it examined nothing, a guard that checks a rebuild
really didn't touch your live data, and a one-line way to record an event with
its author and timestamp. All of it was documented. None of it was demonstrated,
and the examples are the part CI actually runs, so a claim isn't really pinned
down until a demo asserts it.

The projection cookbook now deliberately triggers the "certified something with
no evidence" case — by replaying a stream that exists but is empty, which is what
a renamed stream path actually leaves you with — and checks that it's reported
as such rather than as success. It also wraps its dry-run verification in the
rebuild guard, so "this replay wrote nothing" is checked rather than assumed.

The formkit example switches its own hand-written event-recording helper to the
shipped one. Its audit-history check still identifies the right person for every
event, which is the thing worth proving: switching over is deleting code, not
rewriting it.

One thing not demonstrated, deliberately: bulk writes reaching live browsers.
That only means anything with a browser attached, and a batch command has no
subscriber to observe — faking one would demo the fake. It stays covered by
tests.

### DIFF
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -344,7 +344,68 @@ class ChatRoom(models.Model):
 
 → Deep dive: [Django integration](django-integration.md) · Demos: `just dev`, `just polyglot-dev`
 
-## 11. Protocol conformance
+## 11. Proof that a proof means something
+
+A migration is signed off on a verification sweep: replay the log, collect what
+it *would* write, diff that against the rows you already have. Two ways that
+answer can be worthless, both now guarded — see them in
+`just cookbook-demo`, checks **[3]** and **[4]**.
+
+**A sweep that compared nothing must not report success.** `report.ok` asks "did
+anything disagree?", which is vacuously true on an empty population — and the
+ways a sweep silently examines zero rows are all mundane: a store on the wrong
+backend, a replay over a renamed stream path, a filter that stopped matching, a
+registry that failed to autodiscover. Read `verdict` instead, which separates
+`GREEN` from `VACUOUS`, or `certified`, which is the assertion a proof wants:
+
+```python
+report = diff_effects_against_rows(ex.effects)
+assert report.certified, report      # not `report.ok` — see above
+report.raise_if_diff()               # raises VacuousVerification on an empty run
+```
+
+Failures are checked *before* vacuity, so a real failure is never hidden behind
+"empty".
+
+**A dry run must actually be dry.** `assert_no_live_writes` compares the live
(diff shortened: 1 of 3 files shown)


joshbrooks:
### Technical detail

**Two of my first attempts were wrong, and the corrections are the interesting
part.**

1. To force an empty sweep I first passed `event_match="nope"` to `replay()`.
   The demo still reported `verdict=green` — because the cookbook's handlers use
   `match_field="form_type"`, so **content routing overrides the stream-path
   match** and they fired anyway. Replaced with a stream that exists but holds no
   events, which is both genuinely empty and the realistic cause (a renamed
   path). Worth knowing: `event_match` is not a reliable way to disable handlers
   that route on content.
2. The `_check_a_sweep...` helper originally asserted `report.ok is True` as part
   of the demonstration. That reads as endorsing `ok`, so it is now a commented
   no-op branch showing *why* `ok` is the trap, with the real assertions on
   `verdict` / `certified` / `raise_if_diff`.

**`report.ok` → `report.certified`** in the cookbook's existing check `[2]`,
which also now prints the verdict. That line was the exact shape the vacuous-green
work exists to correct, sitting in the example we point people at.

**The formkit swap is the real test of #94.** `_append_event`'s body becomes one
call to `append_event`. What makes it evidence rather than a tidy-up is that the
demo's `[4] HISTORY` check reconstructs the actor per event from envelope
metadata (`by soares`, `by reviewer:tavares`) — so if the helper's envelope
differed from the longhand in any way that matters, that check fails. It passes.

Note the longhand used `with provenance(user=actor)` while `append_event` sets
`metadata["user"]` explicitly; both land where `envelope_actor` reads, which is
why the swap is invisible downstream. Ruff pruned the now-unused `AppendOptions`
and `provenance` imports, which is a nice signal in itself.

**Deliberately not demonstrated: bulk-append SSE fan-out (#82).** A management
command has no subscriber, so any "demo" would have to stand up a channel layer
and consume from it — at which point it is a test wearing a demo's clothes, and a
worse one than `tests/test_django_rakaia/test_publish_on_append.py` already is.
Said plainly in the commit rather than quietly skipped.

**Gate:** 1117 passed, `ruff check` + `ruff format --check` clean,
`zensical build` clean, and all 11 demos pass — `just cookbook-demo` now prints
five checks, `[3]` and `[4]` being the new ones.

PR #108, opened 2026-08-15
The upgrade guide missed one behaviour change

joshbrooks:
The timestamp fix that just landed changes two things a consumer can be relying on: replays stop rewriting rows that only looked different because of a timestamp, and differences smaller than a millisecond now count as no change at all.

Both belong in the upgrade guide's list of behaviour worth re-checking, and neither was there. This adds them.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -199,3 +199,13 @@ These are not API breaks — nothing to edit — but each changes what your code
 - **Bulk appends now reach SSE subscribers.** `append_many` was invisible to the
   channel layer. If you worked around that by also sending your own frame, you
   will now get two.
+- **Timestamps now compare equal to the column they were encoded from.**
+  `DEFAULT_NORMALIZERS` gained `normalize_temporal`, so a `DateTimeField`,
+  `DateField` or `TimeField` no longer reports a difference on every replay.
+  Two consequences worth checking: a replay with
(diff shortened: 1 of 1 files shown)


joshbrooks:
Detail, per the issue-tracker convention:

`DEFAULT_NORMALIZERS` gained `normalize_temporal` in #103. The two behaviour changes:

- **`skip_unchanged` stops rewriting.** `canonical_value` is shared between `diff_effects_against_rows` and `DjangoExecutor(skip_unchanged=True)`, so rows carrying a `DateTimeField`/`DateField`/`TimeField` used to fail the equality check on every replay and be re-`UPDATE`d. That churn — `auto_now` columns, `post_save` receivers, replication traffic — now stops. Anything a consumer had come to depend on as a side effect of that churn goes away with it.
- **Sub-millisecond differences read as unchanged.** `DjangoJSONEncoder` truncates to milliseconds (`r[:23]`), so both sides are truncated before comparing. Deliberate: sub-millisecond precision cannot survive the log, and the alternative is a phantom diff on essentially every row forever.

Also worth flagging for anyone passing `normalizers=` explicitly: that argument replaces the default set rather than extending it, so a custom set needs `normalize_temporal` added by hand. The same was already true of `normalize_uuid`/`normalize_decimal`, but the temporal one is the first whose absence causes a *write*-path difference rather than only a noisy report.

PR #109, opened 2026-08-15
docs(agents): tell the skills where issues, labels, and domain language live

joshbrooks:
The engineering skills need to know where this project tracks work and what words it uses. This adds three short files under `docs/agents/` recording that issues live on GitHub, which labels drive triage, and that `docs/glossary.md` is the domain vocabulary. `CLAUDE.md` points at them.

It also sets a house rule that issue and PR titles and bodies stay short and plain, with technical detail moved to a comment. Docs only — no code changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,6 +13,22 @@
   then `uv run zensical build`. Without `--extra docs` that step fails with
   `Failed to spawn: zensical`.
 
+## Agent skills
+
+### Issue tracker
+
+Issues live as GitHub issues in `joshbrooks/rakaia`, via the `gh` CLI. Issue and PR
+titles and bodies must be short and jargon-free — two paragraphs at most, with any
+technical detail appended as a comment instead. See `docs/agents/issue-tracker.md`.
+
+### Triage labels
+
+Canonical five-label vocabulary, unmapped. See `docs/agents/triage-labels.md`.
+
+### Domain docs
+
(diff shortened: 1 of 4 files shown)


joshbrooks:
Setup detail, per the new guideline that this goes in a comment rather than the body.

**Files**

- `docs/agents/issue-tracker.md` — `gh` CLI conventions for create/read/list/label/close, the short-body writing rule, and a note that the existing `deferred` label is orthogonal to the triage state machine and must not be stripped during triage.
- `docs/agents/triage-labels.md` — identity mapping for the five canonical triage roles, plus an explicit warning that `question` ("further information is requested" of maintainers) is not a substitute for `needs-info` ("waiting on the reporter").
- `docs/agents/domain.md` — declares the repo single-context and points at `docs/glossary.md` as the domain-language document in place of a `CONTEXT.md`, with the three current ADRs listed. Includes a "don't create a `CONTEXT.md`" instruction so skills don't duplicate the glossary.

**Labels created on the repo**

`needs-triage`, `needs-info`, `ready-for-agent`, `ready-for-human`. `wontfix` already existed and was left alone. These are live already — they were created during setup, independent of this branch.

**Known conflict**

The `to-prd` skill publishes a full PRD as an issue body, which is long by design and collides with the two-paragraph rule. Not resolved here. The fix would be for `to-prd` to post a short body and put the PRD in the first comment, but that changes skill behaviour so it was left out of scope.

PR #110, opened 2026-08-15
Decide which handler types we are not adding, and why

joshbrooks:
The first project to use rakaia in production ended up building five things on
top of it that rakaia has no name for: a way to link two records together, a way
to record a correction that outranks the original value, a way to recompute a
value from somewhere else, a way to record a small status change without
re-sending everything, and the whole write side. The obvious question is whether
rakaia should grow first-class support for any of them.

The answer, after reading their code, is mostly no — four of the five turn out to
be the same three lines of registration, differing only in which events they
match. Everything that makes an edge-link feel different from a correction lives
inside the function body, which is where application logic belongs. So this adds
no new concepts. What it does add is a written record of *why*, with the specific
conditions that would make us change our minds, because the point of writing this
down is to stop the same four proposals coming back around every time someone
reviews the architecture — which is exactly what happened with the translations
demo app that nobody had recorded a decision about.

The one real gap is the write side, and it's narrower than "we need a producer
concept". The helpers worth having already shipped last week; the rest is a rule
we can document but genuinely cannot enforce, and the ADR says so rather than
pretending otherwise.

### DIFF
--- a/docs/adr/0004-handler-types-and-fold-order.md
+++ b/docs/adr/0004-handler-types-and-fold-order.md
@@ -1,0 +1,186 @@
+# ADR 0004 — The content-routed staged handler is the extension point; fold order is a stage
+
+- **Status:** Accepted
+- **Date:** 2026-08-15
+- **Deciders:** rakaia maintainers
+- **Related:** [ADR 0002](./0002-framework-vs-protocol-server-boundary.md) (framework
+  vs protocol-server boundary), [ADR 0003](./0003-handler-hermeticity.md) (handler
+  hermeticity), [`docs/research/handler-types.md`](../research/handler-types.md)
+  (the survey this decision rests on), [`glossary.md`](../glossary.md),
+  [`projection-cookbook.md`](../projection-cookbook.md);
+  `src/rakaia/registry.py`, `src/rakaia/replay.py`, `src/rakaia/effects.py`.
+  Upstream consumer decisions this responds to: partisipa-import ADR-0005
+  (an edge annotates, never materialises), ADR-0008 (no rakaia edge primitive
+  yet), ADR-0016 (corrections are events), ADR-0018 (one fact, one derivation),
+  ADR-0021 (a producer must not consult the projection).
+
+## Context
+
+rakaia names three framework-tier extension points:
+
+| Concept | Shape | Selection |
+|---|---|---|
+| **handler** | `event -> Effect`, or `event, reader -> Effect` at stage > 0 | seq-bracketed version series; glob on the stream path **or** on `event[match_field]` |
+| **reducer** | `reader -> Effect`, or `reader, touched -> Effect` | one current definition per name, last-write-wins, not seq-versioned |
+| **upcaster** | `event -> event` | chained by `from_version` |
+
+The first production consumer has since built five shapes on top of these that
+rakaia does not name: **edge/relation binders** (bind a link between two existing
+subjects, `LINK`/`UNLINK`, last-write-wins by append order), **thin attribute
+folds** (a status stream carrying identity plus one field, folded over the content
+stream), **correction folds** (a later stream whose events override an earlier
+stream's projection), **derivation folds** (a value recomputed from another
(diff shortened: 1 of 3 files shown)


joshbrooks:
### Technical detail

**What this rests on.** `docs/research/handler-types.md` (496 lines, committed
separately on `main` as `814d20a`) — a survey of rakaia's registry/replay/effects
source, the consumer's `streams/` modules and all 21 of its ADRs, plus prior art
read from the official docs of EventStoreDB, Axon, Marten, Eventide and Akka
Persistence.

**The core observation, quoted from four separate consumer modules:**

```python
# relations.py            location_correction.py      project_status.py           status.py
reg.register(             reg.register(               reg.register(               reg.register(
  event_match="project",    event_match="location_…",   event_match="project_st…",  event_match="status_ch…",
  match_field="relation",   match_field="relation",     match_field="relation",     match_field="relation",
  stage=1,                  stage=1,                    stage=1,                    stage=1,
)                         )                           )                           )
```

Four "distinct handler types" that are one registration with a different string.

**The decisive argument on fold order**, because it is the part I'd most want
challenged: correction precedence *cannot* be expressed as
`merge_replay(order_key=ENVELOPE_TS)`. A later re-import carries a later
`event_ts`, so the re-import wins and the correction is lost — which is the exact
failure the consumer's ADR-0016 exists to prevent, and the alternative that ADR
explicitly rejected. Correction precedence is typed authority, not time. `stage=`
is already rakaia's word for "this pass runs after that one regardless of event
order".

Prior art agrees on where the line sits: Axon supports `@Order` **within** one
event processor and states it is "not possible to order event handlers belonging
to different Event Processors", warning that ordering at all is coupling. The
consumer's four sequential `replay()` calls are four processors with no ordering
contract.

**Two vocabulary findings worth knowing** even though nothing renames:

- "projection" means the *opposite* thing in EventStoreDB — there it is a
  server-side, event-**producing** subsystem (`emit()`, `linkTo()`), and read
  models are built by subscriptions. Marten, Akka and Eventide are all on
  rakaia's side, so this is a glossary clause, not a rename.
- "reducer" is rakaia's own coinage; nobody else uses it this way. Marten users
  will look for "multi-stream projection", Axon users for "saga".

**Corrections I applied to the research before committing it.** It recommended
upstreaming `append_event`/`fold_events` — **already shipped in #94**; the
consumer still carries its copy only because it is pinned to PyPI `0.1.0`, which
predates it. That is a release problem, not a design decision. It also deferred
the docs-build result to a summary that did not exist; I ran it — clean, and an
un-navigated page under `docs/` neither breaks nor warns (zensical uses an
explicit nav).

**Where I was skeptical of the consumer's own record**, since a decline should not
rest on unverified claims: ADR-0021 (the producer rule) has its flagship
implementation on a *different branch* — `ranked_priority_ids`/`RANK_STREAM` do
not exist on `feat/durable-store` — the live rank producer still reads the
projection, `bind_project_edge` reads `SeparatedSubmissionProject...exists()` in
its emit/skip control flow (violating the rule), and `location_correction`'s
producer has no caller at all. One recorded decision, zero conforming producers,
one violation. That is why Decision 4 documents the rule rather than building for
it.

**One deferral rather than a rejection**, flagged because the bug it would catch
was expensive: a registry-side "one fact, one derivation" conflict check. The
consumer had one column written by a stage-1 fold and a stage-2 reducer under
*different* rules, each green in a different gate. rakaia can't check it at
registration time — a handler's target column is only knowable from the `Effect`s
it returns — and a runtime warning would fire on multi-owner projections, which
`op="update"` exists to support. Listed as an open question.

**Follow-on work this implies** (not in this PR): a `fold order` glossary entry,
the producer contract beside ADR 0003, and an edge recipe in
`projection-cookbook.md` — including that `Effect(op="update")` replaces the
consumer's hand-rolled `reader.get(...) is None` guard, which they are not
currently using.

**Gate:** `zensical build` clean with the new nav entry, 1156 tests pass,
`ruff check` clean. Docs-only change.

PR #112, opened 2026-08-15
Handlers that needed a dependency were quietly losing change detection

joshbrooks:
Rakaia watches your projection code for edits. If you change a handler that has
already processed events, it tells you — that's the whole point of versioned
handlers, and it's how you avoid silently rewriting history. It turns out that
guarantee switched itself off for any handler that needed something passed in
from outside, and said nothing.

The situation arises whenever a handler needs a dependency — most commonly a
check that has to run against the throwaway database a rebuild is using rather
than the live one. There was no supported way to supply that. The obvious way was
refused with an error message that wasn't true, so people used the other way,
which worked. Except that the change detection then watched the four-line wrapper
instead of the code doing the work: you could rewrite the actual logic completely
and rakaia would report no change at all. The first project using rakaia had four
handlers in exactly this state, along with a comment explaining why they'd picked
the approach that broke it.

The previously-refused approach now works properly and is the documented one.
Where change detection genuinely can't see through what you've written, rakaia
now says so instead of pretending everything is fine — but only when it actually
matters, so it doesn't cry wolf in test suites.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -239,6 +239,22 @@ is not yet tagged in a release.
 
 ### Fixed
 
+- **A handler can take an injected dependency without losing drift detection.**
+  A stage-0 handler is called `fn(event)`, so a dependency had nowhere to go, and
+  both routes out were broken. `functools.partial` was rejected outright by
+  `hash_function_source` — with a message claiming the handler was not in an
+  importable source file, which was untrue — pushing callers to a closure
+  factory, which was accepted and silently harmful: its `dotted_path` carries
+  `<locals>` so `rehydrate()` cannot restore it, and its `source_hash` covers the
+  four-line wrapper rather than the function holding the logic. Rewriting the
+  wrapped function entirely produced an **identical hash**, so drift detection —
+  the feature that exists to catch exactly that edit — was blind, with the
+  library reporting success throughout. `partial` is now unwrapped for both, and
+  a *persisting* registry warns when a registration is not importable. Found in
+  the first production consumer, which had four handlers in this state and a
+  comment explaining why it had chosen the closure.
+
+
 - **The two rebuild guards compose in either nesting order** (#101).
   `assert_no_live_writes` counts rows on the alias `deny_database_access`
   forbids reading, so nesting them the "wrong" way round made the closing
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -122,6 +122,28 @@ A handler registered for a specific **range of sequence numbers**, so old events
 run through the logic that was correct *when they happened*. Fixing a rule going
 forward never rewrites the past. → *[versioned handlers](versioned-handlers.md)*
 
+### Handler dependencies
+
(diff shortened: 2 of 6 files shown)


joshbrooks:
### Technical detail

**RED first.** `tests/test_rakaia/test_handler_dependencies.py` on `main`:
`10 failed, 2 passed`. The 2 that passed are the characterization tests, which is
the point — they document the hazard and keep passing after.

**The defect, demonstrated directly** (this is what made me treat it as severe
rather than cosmetic):

```
wrapper hash before: 38c0a1ee88550eaf
wrapper hash after : 38c0a1ee88550eaf   # after rewriting the wrapped function entirely
drift detected?     False
```

`hash_function_source` hashed `inspect.getsource(closure)` — four lines of
wrapper. Everything it called was outside the hash. So the feature whose entire
purpose is "tell me if this handler changed since it processed those events"
returned "unchanged" for an arbitrary rewrite.

**Why `partial` and not a `deps=` kwarg.** A `deps=` parameter on `register()`
would mean changing `_call_handler`'s dispatch and inventing a second way to
express "this handler takes something extra". `functools.partial(fn, **deps)`
already composes with the existing one-positional-argument dispatch — a bound
keyword doesn't change arity — so the fix is to stop rejecting it. Smaller
change, no new concept, and it makes the consumer's *first instinct* the correct
one.

**Identity semantics, chosen deliberately and pinned:** `partial(fn, dep=a)` and
`partial(fn, dep=b)` hash the **same**, because rebinding a dependency is not a
logic change; editing `fn`'s body drifts both. Pinned by
`test_a_partials_hash_tracks_the_wrapped_function`, which asserts both halves —
a test asserting only "the hashes are equal" would pass on a broken
implementation that returned a constant.

**On the warning's scope.** My first pass warned on every unimportable
registration and produced **69 warnings across our own suite**, because rakaia's
tests register lambdas and closures freely. A library that warns 69 times on its
own tests has taught everyone to filter it. It now fires only when the registry
has a store — i.e. when the path is actually being recorded for later
restoration, which is when an unimportable path is a real defect. A storeless
registry records nothing and restores nothing.
`test_a_closure_in_a_registry_without_a_store_is_quiet` pins that, and the suite
is back to zero warnings.

Note the drift caveat applies either way and the message says so; only the
`rehydrate()` consequence is store-dependent.

**Consumer impact.** The first consumer's four closure-registered handlers
(`generic.build_registry.<locals>._projection` and three in `flip.py`) become a
one-line change to `functools.partial`, after which they get drift detection back
and a `dotted_path` that `rehydrate()` can resolve. Worth noting two of their
`flip.py` builders register the *same* `name="tf611_authoritative"` with
*different* dotted paths today — one logical handler, two identities — which this
also collapses.

**Gate:** 1169 passed (1160 + 9), zero warnings, `ruff check` and
`ruff format --check` clean on `src/ tests/ examples/`, `zensical build` clean.
Glossary gains a **Handler dependencies** entry showing the `partial` form and
saying plainly why not a closure.

PR #113, opened 2026-08-15
Say which parts of rakaia we promise not to break

joshbrooks:
Until now rakaia never told anyone which bits of it were safe to depend on. The
Django half was the worst of it: its documentation listed a "Public API" of four
things, none of which were actually importable from the package. So every project
using it had to reach into internal file names instead, which quietly means that
splitting a file in two breaks them, even though nothing they were promised
changed. The first project using rakaia ended up with 178 imports and 84 direct
database queries against us, and neither side could say which of those were
supported.

This declares the answer in three levels. The top level is the names you can
import straight from `rakaia` and `django_rakaia`, and those won't move without a
major version and an upgrade note. The middle level is the database tables — real,
usable, and deliberately not in the top level, so that depending on them is a
visible decision rather than an accident. The bottom level is everything else,
which can change at any time. It also spells out how to depend on us safely,
which right now our only real user isn't doing.

I audited the example apps against it. They were already clean on everything that
matters — nothing pokes at the database tables, nothing takes apart an offset,
nothing uses a private name. The only gap was import style, so those are updated
to show the supported form. All eleven demos still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,28 @@ is not yet tagged in a release.
 
 ### Added
 
+- **A declared public API, and a contract for it.**
+  [`docs/public-api.md`](docs/public-api.md) sets out three tiers: **Tier 1**
+  (`rakaia.__all__` + the new `django_rakaia.__all__`) which does not change
+  without a major bump and an `UPGRADING.md` entry; **Tier 2**, the ORM models
+  and schema — usable, deliberately *not* exported, and free to change in a
+  minor release; and **Tier 3**, everything else. It also says plainly to depend
+  on rakaia with an **upper bound** (`>=0.2,<0.3`), because on a pre-1.0 library
+  an unbounded `>=` admits every future breaking change and takes it silently on
+  the next lockfile refresh.
+
+  `django_rakaia` previously exported **nothing** — 44 lines of docstring naming
+  a "Public API" that was not importable — so every consumer import had to name
+  an internal module, pinning the module *layout* rather than the surface. It now
+  exports 34 names, resolved lazily (PEP 562) so importing the package still does
+  not pull in the ORM, which is what made eager exports impossible and is now
+  pinned by a test. Both surfaces are pinned name-by-name, so changing one is a
+  deliberate act rather than a diff.
+
+  The examples and docs were the main offender and are converted: **72
+  submodule imports** now use the package root.
+
+
 - **`StreamServerStore` — the protocol-server store surface, named.** `create_app`
   was typed against the concrete in-memory `StreamStore`, so nothing else could
   back the protocol server. The new protocol (exported from `rakaia`) covers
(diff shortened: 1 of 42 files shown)


joshbrooks:
### Technical detail

**Why `django_rakaia` exported nothing, and why lazy fixes it.** The package
docstring already explained the constraint: *"We intentionally do not eager
import models here to avoid AppRegistryNotReady errors during Django's
initialization."* That is correct and is why a plain `__all__` was never an
option. PEP 562 `__getattr__` gets both — a declared surface *and* an import that
costs nothing:

```
package imported with no Django setup: OK
models loaded? False
__all__ available: 34
```

`TestImportingTheDjangoPackageStaysCheap` pins that in a subprocess, because the
property is only observable on a clean interpreter.

**Tier 2 is a deliberate position, not a loophole.** The models stay out of
`__all__` so `from django_rakaia.models import StreamEntry` reads as the weaker
guarantee it is. But they are documented as *supported*, because the store
protocol offers `read(path, offset)` and nothing else — no filter, no limit, no
"latest event matching a predicate". A producer asking the log *"have I already
recorded this?"* (which it must ask of the log, not the projection) has the choice
of O(stream) in Python or an indexed JSON query against the tables. Declaring the
tables off-limits while offering no alternative would just make the only real
consumer non-compliant on paper.

**Audit results — `examples/`, 13 projects:**

| Check | Result |
|---|---|
| Direct ORM access to stream tables | **0** |
| Offset parsing (`int(offset)`) | **0** |
| Private-symbol imports | **0** (one grep hit was `materialize_history as _materialize_history` — a public name aliased to a local, not a private import) |
| Submodule imports of Tier-1 names | 42 → **0** |

So the demos were already clean on every axis that carries a correctness
consequence. The only finding was style, which is the thing the contract now
standardises — and it is our inconsistency, not theirs: `docs/` taught
`from rakaia.replay import replay` three times for a name that has always been in
`__all__`. 72 imports converted across `examples/` and `docs/`, done by script
against the live `__all__` sets so nothing could be converted that isn't actually
exported.

Three needed hand-fixing: two multi-line parenthesised imports the regex skipped,
and one inside `orders/templates/orders/info.html` — a code sample in a template,
which is worth knowing exists.

**zensical validates anchors**, which caught a real error: my link to
`glossary.md#handler-dependencies` pointed at an entry that lives on the #112
branch, not `main`. Changed to a page link, since a bare page reference is valid
either way. Worth noting the docs build fails on broken *anchors*, not just
broken pages.

**On the pinned lists.** `test_public_api.py` writes out all 118 names. That is
deliberately tedious: the failure message is the point — adding a name means
adding it here, removing one means you are making a breaking change and should
notice.

**Not done here:** `UPGRADING.md` still frames everything by git revision and
says *"every current consumer pins to a git revision"*, which stopped being true
when the consumer moved to PyPI. That wants rewriting around versions as part of
cutting 0.2.0, not before.

**Gate:** 1171 passed (+15), `ruff check` + `ruff format --check` clean,
`zensical build` clean, all 11 demos pass.

Issue #114, opened 2026-08-15
Our published version number is still 0.1.0


joshbrooks:
Rakaia still calls itself 0.1.0, and has done for months while the library
changed substantially — including changes that break anyone upgrading. Anyone
installing it gets a number that tells them nothing about what they got, which
is why every current user pins a git revision instead of a version.

Before the next release the number needs to move, and the notes need to say what
moved with it. This is the first of a small set of release-hygiene tasks and
should go first, because the others describe the same release.

joshbrooks:
Concrete shape:

- `pyproject.toml` `version = "0.1.0"` → `0.2.0`. Several changes in the range are breaking (distribution rename to `rakaia-streams`, handler signature and executor changes — see `UPGRADING.md`), so a minor bump is the pre-1.0 equivalent of a major.
- `rakaia.__version__` needs no change: as of #111 it is read from installed package metadata, so it follows `pyproject.toml` automatically. `tests/test_rakaia/test_version.py` asserts the two agree and will catch a partial bump.
- `.github/workflows/publish.yml` already gates the release tag against `pyproject.toml`, so a mismatched tag cannot ship.

Blocks/pairs with the CHANGELOG cut and the `UPGRADING.md` restructure — all three describe the same release and should land together or in immediate sequence.

PR #124, opened 2026-08-15
Remove the helpers nobody uses, before we promise not to

### LINKED ISSUE #123: Remove the helpers nobody uses, before we promise not to
We recently published which parts of rakaia are safe to depend on, and promised
not to change them without a version bump and an upgrade note. A few of the
names on that list have no users at all — not in the library, not in the example
apps, not anywhere except their own tests. One is a duplicate of code we use
elsewhere, and it is the copy missing a safety step the real one has.

None of these has ever appeared in a released version, so removing them now
costs nothing. Once we tag the next release, the same removals need a breaking
change and an upgrade note. This is worth doing first, and it blocks the release
in #114.

joshbrooks:
We recently published which parts of rakaia are safe to depend on, and promised not to change them without a version bump and an upgrade note. Three of the names on that list had no users at all — not in the library, not in the example apps, not anywhere except their own tests. One of them was a duplicate of code used elsewhere, and it was the copy missing a safety step the real one has. None has ever appeared in a released version, so removing them now costs nothing; after the next tag the same removals would need a breaking change and an upgrade note.

This removes all three along with their tests and every mention in the docs, corrects a documentation claim that one of the example apps used one of them (it has always had its own version), and tidies up some leftover notes-to-self in the streaming code. The docs build and the full test suite pass unchanged.

Closes #123. Unblocks #114.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -518,6 +518,27 @@ is not yet tagged in a release.
   change: a stream literally named `translations` now resolves to the stream
   detail page.
 
+- **Three helpers nobody was using are gone.** All three were added after
+  `0.1.0` and never appeared in a tagged release, so **no upgrade note is
+  needed** and no released code can break: there is nothing to migrate from.
+
+  - `send_sse_event` — an internal helper for writing a Server-Sent Event that
+    nothing in the library ever called. It was also the unsafe twin of the code
+    the protocol server actually uses: the real one normalises carriage returns
+    so a payload cannot forge an event boundary, and this copy did not.
+  - `dispatch_external` — routed the "external" effects rakaia deliberately
+    never applies (email, webhooks) to per-kind handlers. Its own documentation
+    described the alternative: a two-line loop in your own code. The
+    `multi_owner` example now writes that loop.
+  - `recover_peak_snapshot` — recovered a record's most complete historical
+    snapshot after a legacy blank save. Once the audit rows exist this is a
+    one-line scan over them, and each application wants a slightly different
+    version, so the `partisipa_history` example keeps its own — as it always
+    did.
+
+  `check_disjoint_defaults`, in the same module as one of these, stays: it runs
+  on every effect apply.
+
 ### Changed
 
 - **BREAKING — `DjangoStreamStore.append` and `.append_many` return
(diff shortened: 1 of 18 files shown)


joshbrooks:
Technical detail.

**`send_sse_event` (`src/rakaia/_asgi.py:140-159`)** — deleted with `tests/test_rakaia/test_asgi_helpers.py::TestSendSseEvent`. Confirmed zero `src/` importers first (`handler.py`'s import block never pulled it, and it was not in `rakaia/__all__`). It split on `\n` only; `handler.py::_encode_sse_data` normalises `\r\n` and stray `\r` per protocol §5.7.

**`dispatch_external` (`src/rakaia/effects.py:387-418`)** — removed from `rakaia/__all__`, `RAKAIA_PUBLIC` in `test_public_api.py`, `TestDispatchExternal` in `test_effects.py`, and the two rows in `docs/examples.md`. `examples/multi_owner/demo.py` section 5 now writes the two-line loop the docstring described and still asserts the same result; the README row and demo docstring were reworded to match. `collections.abc.Mapping` became an unused import in `effects.py` and was dropped. `check_disjoint_defaults` untouched — `effect_executor.py:74` calls it on every apply.

**`recover_peak_snapshot` (`src/rakaia/history.py:113-139`)** — removed from `__all__`, the public-API test, `TestRecoverPeakSnapshot` in `test_history.py`, and the module docstring. Docs: `history-read-model.md`'s "Recovering the peak snapshot" section kept (with its `#recovering-the-peak-snapshot` anchor, which `glossary.md:86` deep-links) but rewritten to present the recovery as a `max(..., key=len)` scan over the audit rows plus the mermaid diagram, rather than a shipped helper. `glossary.md`, `whats-new.md:301` and `pghistory-retirement.md:81,105` reworded the same way.

**The false docs claim** — `docs/examples.md:124` said `partisipa_history` demonstrated the library's `recover_peak_snapshot`. That example hand-rolls its own at `history/stream_history.py:116` (a queryset `max`), alongside a pghistory-side twin at `pghistory_today.py:34`, and the demo asserts the two agree. Both example-local functions are untouched; the docs row now says "hand-rolled over audit rows", and `examples/partisipa_history/README.md:75` qualifies the name as `stream_history.recover_peak_snapshot()` so it can't be read as a library import.

**`handler.py` tidy-up** — deleted the stream-of-consciousness block at `:750-755` ("wait the test does…", "Actually the spec says…"), replacing it with a three-line note stating the §5.7 rule. The four hand-built control frames (`:912`, `:934`, `:949`, `:959`) now route through `_build_sse_control`. The first three are the same terminal frame and share a new `_closed_control(offset)` wrapper — passing `control_offset == tail_offset` and `stream_is_closed=True` makes the helper emit `{offset, closed: true}` exactly. The keep-alive frame calls `_build_sse_control` directly with `up_to_date=True`, which regenerates the same cursor `keep_cursor` held, so that local went away. Key insertion order matches the old literals in both cases, so `json.dumps` produces byte-identical frames; the SSE tests pass unchanged.

**Verified**: `ruff check`, `ruff format --check`, `pytest` (1178 passed), `zensical build` (no issues). `grep -rn` for each name across `src/ tests/ docs/` returns nothing; the only remaining hits in `examples/` are `partisipa_history`'s own two functions.

PR #125, opened 2026-08-15
The dashboard no longer shows the wrong events for a resumed view

### LINKED ISSUE #122: Two pages accept stream positions we never handed out
A stream position is meant to be an opaque token — you hand it back to us to
resume reading, and only the backend that issued it can say what it means. Two
of our Django pages instead convert whatever they are given straight to a
number. Because Python quietly ignores underscores inside numbers, a position
issued by our other backend is accepted and read as an entirely different one.

Nothing errors. The page returns a window of the stream that has nothing to do
with what was asked for, or in the live-updates case silently replays the whole
thing from the beginning. We already have a strict check for this that both
backends use; these two pages just do not go through it.

joshbrooks:
The dashboard's event list and its live feed both accepted a position marker from the browser and trusted it without checking it belonged to them. A marker issued by the other kind of stream store looked close enough to a valid one to be read as a completely different position, so the page could quietly show the wrong slice of a stream and still report success. In one case an unrecognised marker was treated as "start from the beginning", so a page that thought it was picking up where it left off was handed the whole history again.

Both entry points now check the marker is one they issued and reject it with a clear error if it isn't. Being told the request is bad is far easier to notice than being shown plausible but wrong events.

Closes #122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -261,6 +261,26 @@ is not yet tagged in a release.
 
 ### Fixed
 
+- **The dashboard views refuse an offset the durable store never issued** (#122).
+  Two Django views parsed a client-supplied offset with bare `int()`, outside
+  any store's ownership check: `after_offset` on the JSON events endpoint, and
+  the `Last-Event-ID` header the browser sends when an `EventSource` reconnects.
+  Python reads `_` as a digit separator, so the in-memory store's compound
+  `{seq}_{byte}` offset — `int("0_5")` is `5` — resolved to an unrelated
+  position: the JSON API returned the wrong window with a `200`, and the SSE
+  endpoint resumed in the wrong place. An offset that did not parse at all fell
+  into `except ValueError: 0` and silently replayed the entire stream to a
+  client that believed it was resuming. Both now parse through the durable
+  store's own strict check and answer `400`; failing a connection is visible,
+  and neither wrong answer was.
+
+  That check has moved to `django_rakaia.offsets` alongside `format_offset`,
+  which also settles a layering inversion: `Stream.current_offset` used to
+  import `format_offset` back out of the store module for its own property, and
+  duplicated `DjangoStreamStore.get_current_offset`'s watermark query verbatim.
+  The store now delegates to the model property; all it adds is the expiry
+  check and the absent-stream `None`.
+
 - **A handler can take an injected dependency without losing drift detection.**
   A stage-0 handler is called `fn(event)`, so a dependency had nowhere to go, and
   both routes out were broken. `functools.partial` was rejected outright by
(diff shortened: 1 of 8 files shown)


joshbrooks:
## Technical detail

**The bug.** Two views parsed a client-supplied offset with bare `int()`, outside any store's ownership check:

- `views.py` `stream_events_api` — `int(after_offset)` from the query string, then `filter(offset__gt=...)`.
- `channels_views.py` `stream_events_sse` — `int(last_event_id)` from the `Last-Event-ID` header, with `except ValueError: after_offset = 0`.

Python treats `_` as a digit separator, so `int("0_5") == 5`. The in-memory store issues compound `{seq}_{byte}` offsets, so one handed to these views resolves to an unrelated durable position.

**RED, before the fix.** `test_foreign_after_offset_is_refused` got a `200` with `{"count": 1, ... "offset": 6}` — the view had resolved `"0_5"` to `5` and filtered `offset__gt=5` over six entries. `test_sse_foreign_last_event_id_is_refused` got a `200` `StreamingHttpResponse`. Both mirror `tests/server_store_contract.py::test_reading_from_a_foreign_offset_is_refused`.

**GREEN.** `DjangoStreamStore._parse_offset` was promoted to a module-level `parse_offset`, and it plus `format_offset` now live in a new `django_rakaia/offsets.py` rather than in `django_store` — the views would otherwise import the whole store module to validate a query parameter. `DjangoStreamStore._parse_offset` remains as a one-line delegate, so there is no second copy.

The JSON view returns `400 {"error": "after_offset must be an integer"}` — the message and status of the existing invalid-`after_offset` branch, so no new response shape.

**SSE choice: `400 HttpResponseBadRequest`, not a silent reset.** Refusing is the only option that is visible. Resuming at a mis-parsed position hands the client a wrong window under a `200`; the old `except ValueError: 0` fallback replays the entire stream to a client that believes it is resuming, which for an at-least-once consumer means silent duplicate delivery. A `400` fails the `EventSource` connection, which is loud and debuggable. Only a *foreign* `Last-Event-ID` is refused — an absent header still means "fresh connect, replay all", unchanged.

**Step 3 (optional cleanup), done.** `models.Stream.current_offset` duplicated `DjangoStreamStore.get_current_offset`'s `max(entries, watermark)` query verbatim and imported `format_offset` back out of the store module, so the model layer depended on the store layer for its own property. With `format_offset` in `offsets.py` that inversion is gone, and `get_current_offset` now delegates to `stream.current_offset`; all it adds is the expiry check and the absent-stream `None`. Nothing else came with it.

**Gate.** `ruff check`, `ruff format --check`, `pytest` (1190 passed), `zensical build` all clean.

PR #126, opened 2026-08-15
A closed stream now says so, whichever way you append

### LINKED ISSUE #119: Appending to a closed stream and closing it twice disagree
Rakaia decides whether to accept a write in one place, on purpose — that module
exists because our two storage backends had already drifted apart once on this
exact question. Seven other places decide it themselves anyway, and they do not
all apply the rules in the same order. The result is that the same rejected
write can be reported differently depending on which door it arrived through.

Nothing caught this because our shared backend test suite never tries the one
combination that exposes it: a fenced producer writing to a stream that is
already closed. Adding that case first, watching it fail on one backend and pass
on the other, is the cleanest way in.

joshbrooks:
Writing to a stream that someone has already closed gave two different answers depending on which store was underneath. On the database-backed store, a writer whose numbering had also slipped was told to fix its numbering — and retrying got the same reply forever — instead of being told the stream was closed and there was nothing left to do. A writer re-sending the very message that closed the stream, after losing the reply, was likewise not recognised as repeating itself.

Both stores now run the one shared set of admission rules, in one order, for every way of appending, and the shared test suite covers the case that had been missed. What a client sees over the wire is unchanged.

Closes #119

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -280,6 +280,21 @@ is not yet tagged in a release.
   duplicated `DjangoStreamStore.get_current_offset`'s watermark query verbatim.
   The store now delegates to the model property; all it adds is the expiry
   check and the absent-stream `None`.
+- **A closed stream answers the same way on both stores, whichever append door
+  you use** (#119). `append_with_producer` on the durable store ran its own
+  admission sequence — producer fencing *before* the closed check, and never
+  reading who closed the stream. A producer whose sequence had also drifted was
+  told about the gap instead of being told the stream was closed, and a producer
+  retrying its own closing append was never recognised as a duplicate; the
+  in-memory store got both right. Both doors on both stores now ask
+  `rakaia.append_decision`, the one shared admission sequence, and the shared
+  store contract drives the combination that was missing, which is why the drift
+  had gone unseen. A refusal on a closed stream now always carries a producer
+  result — `ProducerDuplicate` for a retry of the closing tuple, otherwise
+  `ProducerStreamClosed` — which is the same 409 (or 204) on the wire as before,
+  since the server already synthesised the closed result when the store gave it
+  nothing.
+
 
 - **A handler can take an injected dependency without losing drift detection.**
   A stage-0 handler is called `fn(event)`, so a dependency had nowhere to go, and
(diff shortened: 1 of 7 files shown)


joshbrooks:
**RED first.** `tests/server_store_contract.py` gained `test_append_with_producer_to_a_closed_stream_reports_closed`: append `p/0/0`, close the stream, then `append_with_producer` with `p/0/9`. It failed on `TestDjangoStreamStoreServerContract` (leaked `ProducerSequenceGap`) and passed on the in-memory binding, exactly as the issue predicted. A second case pins the retried closing tuple through the same door.

**`decide_append`.** A closed stream with a non-matching producer tuple now returns `ProducerStreamClosed()` rather than `producer_result=None`. Verified no wire change before relying on it: `producer_response`'s `ProducerStreamClosed` arm ignores the `stream_closed` flag and emits the same `409 Stream is closed` with the same `Stream-Closed`/`Stream-Next-Offset` headers the old `handler.py:1175-1179` fallback produced. That fallback is deleted; the `assert decided is not None` stays with a comment saying why it now holds.

**Collapsed.**
- `DjangoStreamStore._append_with_producer_sync` builds a `StreamFacts` from the locked `Stream` row and calls `decide_append` — this is what turns the RED test green.
- Both close paths (`StreamStore.close_stream_with_producer`, `DjangoStreamStore._close_with_producer_sync`) call `decide_append` with a body-less `AppendOptions`; `verdict.stream_closed` maps to `already_closed`.
- Deleted `django_store._check_content_type`, `_check_seq`, `DjangoStreamStore._validate_producer` and `StreamStore._validate_producer` — all unreferenced after the above (grep-confirmed). `append_many`'s pre-flight keeps its content-type check inline, matching `StreamStore.append_many`, so the batch path is behaviourally untouched.
- `TestOrdering` gained `test_closed_is_checked_before_producer_fencing`. Two assertions that pinned the old `producer_result is None` artifact were updated (`server_store_contract.py`, `test_append_decision.py::TestClosed`).

**Out of scope, deliberately:** producer fencing on either `append_many`. Neither fences today; routing them through `decide_append` would add fencing, and a per-item producer-state lookup would break `test_append_many_bounded_queries`.

Full gate green: `ruff check`, `ruff format --check`, `pytest` (1193 passed), `zensical build`.

PR #127, opened 2026-08-15
Putting events into a stream is one call now

### LINKED ISSUE #120: Putting events into a stream takes four lines everywhere
Getting a handful of events into a stream is the first thing anyone does with
rakaia, and we have never provided a way to do it. So everyone writes the same
four lines: create the stream, loop, turn each event into JSON, encode it to
bytes. Our own test suite does this in five places and our example apps in
twelve, with small inconsistencies between them.

The cost lands hardest on tests, where the setup can run to sixty-eight lines
before the first thing is actually checked. A reader cannot see what the test is
about. We already built this affordance once for the Django side and it worked;
it just never reached the core library or our own suite.

joshbrooks:
Getting a handful of events into a stream used to take four lines, and rakaia
never shipped a way to do it — so everyone wrote the four lines. Our test suite
had six copies of them and our examples twelve more, with small differences
between them. The cost landed hardest on tests, where the setup could run to
sixty-eight lines before the first thing was actually checked.

There is now one call that does it, and every copy has been replaced with it.
The worst test setup drops from sixty-eight lines to thirty-five, and the Django
helper that already did this for its own users is now built on the same code, so
there is one rule for turning an event into bytes rather than two that can drift
apart.

Closes #120

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,27 @@ is not yet tagged in a release.
 
 ### Added
 
+- **`rakaia.seed_stream` — putting events into a stream is one call.** Getting a
+  handful of events into a stream took four lines (create, loop, `json.dumps`,
+  `.encode()`) and rakaia never shipped a way to do it, so the test suite carried
+  six hand-rolled copies in three different shapes and the examples twelve more,
+  four of them wrapped in a locally-defined `_append`. Setup in the worst test ran
+  to sixty-eight lines before the first assertion.
+
+  `seed_stream(path, events, store=..., encoder=...)` builds an in-memory
+  `StreamStore` when no store is given, accepts any `WritableStore` when one is,
+  and returns it either way. Payloads may be dicts or pre-encoded bytes; the
+  label/metadata/`event_ts` envelope is **per event** — pair a payload with an
+  `AppendOptions` — because a batch-level label is not what the callers needed.
+  Creation is idempotent and non-destructive, so seeding an existing path appends
+  and no caller needs a `has()` guard.
+
+  The `encoder` parameter is the point: `django_rakaia.envelope.append_event` and
+  `fold_events` are now built on it, passing `DjangoJSONEncoder` in, so there is
+  exactly one `json.dumps` rule in the codebase rather than the drifting second
+  copy `envelope.py` warns about. `append_event`'s output is unchanged, byte for
+  byte, and a test pins that.
+
 - **A declared public API, and a contract for it.**
   [`docs/public-api.md`](docs/public-api.md) sets out three tiers: **Tier 1**
   (`rakaia.__all__` + the new `django_rakaia.__all__`) which does not change
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -404,6 +404,29 @@ Its sibling `fold_events` runs a batch of events through your handlers *now*,
 via a throwaway in-memory stream, so write-time projection and rebuild-time
 projection run the same code rather than two implementations that can disagree.
 
+Underneath both is `rakaia.seed_stream`, which is the same idea one tier down:
+getting a handful of events into a stream, without Django in the picture.
+
+```python
+from rakaia import AppendOptions, seed_stream
+
+store = seed_stream("submissions", [
+    ({"key": "s1", "a": 1}, AppendOptions(label="insert")),
+    ({"key": "s1", "a": 2}, AppendOptions(label="update")),
+])
+```
+
+Omit `store=` and you get a fresh in-memory one back; pass any `WritableStore`
+— the durable Django store, the `get_store()` singleton — and it is used and
+returned. Payloads may be dicts or already-encoded bytes, the envelope is per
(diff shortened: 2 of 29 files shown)


joshbrooks:
## The interface

`rakaia.seed_stream`, in the new `src/rakaia/seed.py`, exported from `rakaia.__all__`:

```python
def seed_stream(
    path: str,
    events: Iterable[SeedEvent] = (),
    *,
    store: WritableStore | None = None,
    encoder: type[json.JSONEncoder] | None = None,
) -> WritableStore
```

Each decision is forced by a real caller from the issue comment:

- **Optional store, always returned.** `store=None` builds a fresh in-memory
  `StreamStore`; anything satisfying `WritableStore` is used as-is. Overloads make
  the return type exact, so `_seed() -> DjangoStreamStore` still type-checks.
- **Per-event envelope, not batch-level.** An event is a payload, or a
  `(payload, AppendOptions)` pair. This reuses the vocabulary `append` already
  speaks instead of inventing a parallel `label=`/`metadata=`/`event_ts=` triple —
  one concept, and it covers `test_history_materializer` (different labels,
  metadata on the first event only) and the backfill test in `test_replay`
  (different `event_ts` per event) without a special case for either.
- **Dicts or bytes.** Bytes are appended untouched and the encoder never sees them.
- **Unconditional `create()`.** Idempotent and non-truncating by the store
  contract, so `test_replay_command`'s `has()` guard is gone.
- **`encoder` is a `json.JSONEncoder` subclass**, not a `Callable[[Any], bytes]`.
  The callable form would have let `django_rakaia` pass
  `lambda p: json.dumps(p, cls=DjangoJSONEncoder).encode()` — which is the drifting
  second copy of the rule that `envelope.py:15-19` warns about. Taking the class
  and passing it to a single `json.dumps` means there is exactly one such call in
  the repo. `append_event` and `fold_events` are both rewritten onto `seed_stream`;
  `test_the_message_is_byte_identical` already existed and still passes unchanged.

## What was replaced

Eight helper definitions deleted outright and two more gutted to a single
delegating line:

| site | was |
|---|---|
| `tests/test_rakaia/test_replay.py:49` | `_seed_stream(store, path, events)` |
| `tests/test_rakaia/test_replay.py:1305` | closure with per-event `event_ts` |
| `tests/test_rakaia/test_alerts.py:69` | byte-identical twin of the above |
| `tests/test_django_rakaia/test_alerts.py:388` | builds its own store, returns it |
| `tests/test_django_rakaia/test_alerts.py:583` | inline copy inside `_run` |
| `tests/test_django_rakaia/test_history_materializer.py:28` | bytes + per-event label/metadata |
| `tests/test_django_rakaia/test_replay_command.py:20` | `get_store()` + `has()` guard |
| `examples/.../demo_staged.py:51` | `_append` |
| `examples/.../demo_merge.py:57` | `_append` |
| `examples/.../demo_repeaters.py:92` | `_append` (a method) |
| `examples/.../demo_close.py:53` | `_append` |

Plus the inline copies in `test_projection_reader`, `test_replay_helper`,
`test_hermeticity`, `test_using_seam`, `test_verification`, `test_django_store`,
`test_subscription`, `test_executors`, `demo_cookbook`, `demo_orders`,
`demo_history`, `orders/live.py` and `stream_history.py`. In total 31 raw
`json.dumps(...).encode(...)` lines removed; 91 call sites now go through
`seed_stream`.

`tests/test_django_rakaia/test_alerts.py:198` is untouched, as flagged — despite
the name it creates an ORM row.

`examples/.../demo_submissions.py:49` is also untouched: its `_append_event` is a
one-line delegation to `django_rakaia.envelope.append_event` whose docstring
exists to explain that helper, so it is documentation rather than a copy.
`examples/protocol_streams/demo.py` keeps its raw appends — the raw store API is
what that demo is about.

## Before / after

`tests/test_django_rakaia/test_projection_reader.py::test_merge_two_finance_streams_then_reduce`
— lines from the `def` to the first assertion:

- before: 68
- after: 35

The seeding block itself went from 42 lines to 6, via a small `_line(key, suku, delta, hour)`
builder shared with the existing `_FINANCE_EVENTS`.

## Tests

`tests/test_rakaia/test_seed.py` — 16 tests, written first and confirmed red
(`ImportError: cannot import name 'seed_stream'`) before any implementation.
Covers store construction/passthrough/return, an arbitrary `WritableStore`,
non-truncating re-seed, empty-events-still-creates, dict/bytes/mixed payloads,
list order, per-event `AppendOptions` (label, metadata, `event_ts`, mixed with
bare events), and the encoder hook including that it is never applied to bytes.

## Verification

`ruff check`, `ruff format --check`, `pytest` (1201 passed), `zensical build`, and
`just demos` — all twelve demos green, which is the real proof here since #107.

## Docs

Added to the "One envelope, written once" section of `docs/whats-new.md`, which is
where `append_event` and `fold_events` already live — `seed_stream` is the same
idea one tier down and the `encoder` hook is what connects them. Plus a CHANGELOG
entry. `docs/public-api.md` needed no change: it describes the tiers rather than
enumerating names, and `tests/test_rakaia/test_public_api.py` is what pins the set.

PR #128, opened 2026-08-15
Make our two ways of reading projections prove they agree

### LINKED ISSUE #121: Nothing checks that our two ways of reading projections agree
Rakaia has two ways to read already-built projections and two ways to write
them, plus a couple more that live in tests and example apps. Verifying a
rebuild rests entirely on two of those giving identical answers — that is the
whole point of the feature — and nothing anywhere checks that they do.

We already solved this shape of problem for storage backends: one shared suite
of behaviours, and each backend proves it passes. Doing the same here is mostly
following a pattern we have used twice before. It should also let three example
apps and one test file delete their own hand-written versions.

joshbrooks:
Rakaia has several ways to read already-built projections and several ways to write them. Checking that a rebuild worked rests entirely on two of those giving identical answers, and nothing anywhere checked that they did — a disagreement would not have failed loudly, it would have quietly reported a rebuild as verified when it was not. This does for those two seams what we already did twice for storage backends: one shared suite of behaviours, and each implementation proves it passes.

It also lets us delete four hand-written copies. One test file and three example apps each carried their own cut-down version of a reader or a writer; they are replaced by the real ones, and a single in-memory implementation now covers everything the test and example copies did between them. All twelve demos and the full suite pass.

Closes #121


joshbrooks:
Technical detail.

**Blocker cleared first.** `Executor` (`src/rakaia/effects.py`) was not `@runtime_checkable`, so `isinstance(x, Executor)` raised `TypeError` and the house contract idiom's opening assertion was impossible. Added, with a docstring note that it checks method presence only — which is why both suites go on to pin behaviour.

**`tests/executor_contract.py`** — `ExecutorContract`, one `seam` fixture returning an `ExecutorSeam(executor, reader, model)`. The executor seam is only observable through a reader, so the seam carries both plus the `model_label` of a projection with the columns the suite writes (the Django binding points it at the `Alert` test model). What it pins:

- **Three ordered passes.** Three tests emit a batch whose in-order answer is the *wrong* one — `[delete(scope), upsert(row-in-scope)]` must end with zero rows, not one; likewise write-before-retire and delete-before-retire — plus a reconcile emitted both ways landing on the same rows.
- **`produces=` recorded before any effect that Refs it**, including a delete emitted *first* whose `lookup` carries a `Ref` to a later upsert: it resolves rather than raising.
- **One `RefResolver` per `apply()`** — a `Ref` to a previous batch's `produces` raises `UnresolvedRefError`.
- **`check_disjoint_defaults` before the first write** — a colliding batch raises with nothing applied (true for both backends because the check precedes the write, not because of the transaction).
- **`retire_flips` only for retires with `transition_kind`**, with the identity dicts ordered by `transition_key_fields`.

Return type is not asserted, per `effects.py`'s `ApplyReport | None`: the report is inspected only `if report is not None`. Also bound `DjangoExecutor(skip_unchanged=True)` — a different write path, so it runs the whole suite too.

**`tests/projection_reader_contract.py`** — `ProjectionReaderContract`, one `seam` fixture returning a `ReaderSeam(apply, make_reader, model)`. Two callables rather than a bare reader because `PreloadedProjectionReader` takes the batch at construction: `apply` puts rows there, then `make_reader` is handed the same effects. Pins `get` returning the match or `None` (never raising on absence), `filter`/`query` returning re-iterable iterables, and **positional-only `model_label`** on all three methods.

**Did the two Django readers agree?** Yes — `DjangoProjectionReader` and `PreloadedProjectionReader` both passed the reader contract unmodified, cache-hit path and live-fallback path. No behavioural disagreement found, so nothing was papered over.

**Proof the contracts bite.** Before deleting them, the pre-existing copies were bound to the new suites on a scratch branch: **27 failures**, including every ordering test, every `Ref` test and `test_colliding_batch_raises_with_nothing_applied` against `DictProjections`; `test_model_label_is_positional_only` against its reader half — exactly the hole `protocols.py` warned about. `test_satisfies_the_executor_protocol` passed for all of them, which is the point about `runtime_checkable`.

**The two divergences, settled.**

1. *Synthetic pk minting* — kept, from `DictExecutor`. Every row is minted with an integer `pk`, so `Ref("x")` (whose default field is the primary key, the FK case) resolves in memory as it does on Django rather than being the one thing an in-memory run cannot rehearse. The old `DictProjections` rows had no `pk` at all. Additive for the replay tests: all 50 pass unchanged.
2. *`None` vs `ApplyReport`* — the shipped one returns `ApplyReport`, matching `DjangoExecutor`, because that is the only way it can report `retire_flips` and be held to that clause. The *contract* still permits `None`, since `effects.py` does and the orchestrator reads it as empty.

The union also takes `DictExecutor`'s `__in`/`__isnull` matcher and applies it to the reader half (the old reader was exact-match only), and mirrors `update_or_create`'s create path by dropping operator lookups from the minted row.

**Deletions.** `DictProjections` out of `test_replay.py` (now imports `rakaia.executors`); `examples/multi_owner/executor.py` deleted, demo uses `rakaia.DictProjections`; the three example `Refs` classes replaced by `DjangoProjectionReader` at four call sites. The `partisipa_close` and `partisipa_merge` ones were byte-identical and all three lacked `using=`, so none could have participated in rebuild verification.

**Surface.** `DictProjections` added to `rakaia.__all__` and to `test_public_api.py`. Documented in `docs/dry-run-and-executors.md`; the reader/executor agreement is now a fourth named Tier-1 behavioural contract in `docs/public-api.md`.

**Verified.** `ruff check`, `ruff format --check`, `pytest` (1306 passed), `zensical build`, `just demos` (all twelve).

Issue #61, opened 2026-07-26
Conformance: implement the stream-forking family (Stream-Forked-From / Stream-Fork-Offset / Stream-Fork-Sub-Offset)


joshbrooks:
## Summary

rakaia passes the full [Durable Streams protocol](../docs/protocol.md) conformance surface **except the stream-forking family** — currently **56 tests**, the entire `Fork - *` group of `@durable-streams/server-conformance-tests`. Forking is not yet implemented.

This is the accepted gap baselined in [`conformance/expected-failures.txt`](../blob/main/conformance/expected-failures.txt). The conformance CI job (`.github/workflows/conformance.yml`) diffs results against that baseline so a real regression in a currently-passing behavior is distinguishable from this known gap. This issue is the tracking issue referenced by `conformance/README.md`.

## Scope — protocol features not yet implemented

The fork family exercises these headers / behaviors:

- `Stream-Forked-From` — create a stream that inherits a source stream's data up to a point.
- `Stream-Fork-Offset` — fork at a specific message offset (incl. head = inherit all).
- `Stream-Fork-Sub-Offset` — fork *within* a message (JSON flattened-message boundary / binary byte boundary), with prefix materialization of the partial message.
- **Producer/idempotency state** must NOT cross the fork boundary.
- **Content-type inheritance** when the header is omitted (incl. the sub-offset case).
- **Reading** across the fork boundary; fork must not see source appends after the fork point; source stays independent.
- **Recursive forks** — chained forks, sub-offset composition across levels, independent appends per level.
- **Deletion & lifecycle** — soft-delete semantics (410s), re-creation 409s, cascade GC (incl. three-level chains and last-fork-deleted).
- **TTL & expiry** — TTL inheritance from source; source expiry with living forks.
- **Live modes** — SSE + long-poll handover at the fork offset; inherited data returned immediately.
- **Validation** — 400s for malformed / overshooting / dependency-missing sub-offset & offset combinations; 404 for nonexistent source; 409 for mismatched re-creation.

## Baselined tests (56)

Grouped by conformance `ancestorTitle` (see `conformance/expected-failures.txt` for exact `fullName`s):

| Category | Count |
| --- | --- |
| Fork - Creation | 24 |
| Fork - Deletion and Lifecycle | 11 |
| Fork - Reading | 4 |
| Fork - Recursive | 4 |
| Fork - Appending | 3 |
| Fork - Edge Cases | 3 |
| Fork - Live Modes | 3 |
[...]

PR #129, opened 2026-08-15
Give rakaia a version number that means something

### LINKED ISSUE #114: Our published version number is still 0.1.0

### LINKED ISSUE #115: Everything we have built is filed under 'unreleased'
The changelog has a single "Unreleased" section holding every change since the
first version, forty-seven entries deep. A reader cannot tell which of those they
already have, and the publish process reads the changelog to describe a release,
so right now it has nothing to describe.

Cutting that section into a dated release entry is mechanical, but it has to
happen alongside the version bump rather than after it, since the two have to
agree on what the release contains.

### LINKED ISSUE #116: Our upgrade notes assume the version never changes
The upgrade guide is organised by git revision rather than by release, and says
so in its opening lines — on the stated grounds that the version number has not
moved. That was true and is about to stop being true.

It also matters more than it used to: we now publicly promise that anything we
break comes with an upgrade note, so the guide is part of a commitment rather
than a convenience. Re-heading the existing sections under the release they land
in is mostly mechanical, and no content needs rewriting.

joshbrooks:
Rakaia has called itself 0.1.0 since the beginning, through about twenty-five
commits and several changes that break things for anyone upgrading. Because the
number never moved, every project using us pins a git revision instead — which
works, but means nobody can say what they are running or what changed between
two installs.

This makes 0.2.0 real: the number moves, the fifty-odd accumulated changelog
entries become a dated release, and the upgrade guide is reorganised around
releases rather than git revisions now that there is a release to organise it
around. The notes also state plainly, where a user will actually read it, the one
part of the protocol we have not implemented.

Closes #114
Closes #115
Closes #116

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ conformance/node_modules/
 conformance/conformance-results.json
 conformance-results.log
 conformance-server.log
+
+# Agent worktrees — transient checkouts, never part of the tree
+.claude/worktrees/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,25 @@ runnable demo for each.
 
 ## [Unreleased]
 
-Everything below has landed on `main` since the initial `0.1.0` groundwork and
-is not yet tagged in a release.
+Nothing yet.
+
+## [0.2.0] - 2026-08-15
+
+The first release that describes the library. `0.1.0` was the initial
+groundwork; everything below landed on `main` afterwards and, until now, was
+reachable only by pinning a git revision — which is what every consumer was
+doing, because the version number could not tell them what they had.
+
+Several entries are breaking. See [UPGRADING.md](UPGRADING.md) for what to do
+about each; it is organised by release from this version on.
+
+**Known limitation.** Rakaia passes the Durable Streams conformance suite except
+the stream-forking family — `Stream-Forked-From`, `Stream-Fork-Offset` and
+`Stream-Fork-Sub-Offset`, 56 tests. Forking is not implemented. The gap is
+baselined in `conformance/expected-failures.txt` so a real regression stays
+distinguishable from it, and tracked in
+[#61](https://github.com/joshbrooks/rakaia/issues/61). Everything else in the
+protocol is covered by both stores.
 
 ### Added
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -4,15 +4,22 @@ icon: lucide/arrow-up-circle
 
 # Upgrading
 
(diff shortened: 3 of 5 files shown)


joshbrooks:
**What changed, precisely:**

- `pyproject.toml` `version` → `0.2.0`. No change needed to `rakaia.__version__` — since #111 it reads from installed package metadata, and `tests/test_rakaia/test_version.py` asserts the two agree. Confirmed after a re-sync: `rakaia.__version__ == "0.2.0"`.
- `CHANGELOG.md`: the `[Unreleased]` block (~50 entries) becomes `## [0.2.0] - 2026-08-15`, with a fresh empty `[Unreleased]` above it. No entry text was rewritten.
- Added a **Known limitation** note to the 0.2.0 section: the stream-forking family is unimplemented (56 baselined conformance tests, #61). It was visible in CI and in `conformance/README.md`, but nowhere a user reads.
- `UPGRADING.md`: preamble no longer claims the version number never moves; content is now under a `# 0.2.0` heading, with each section keeping its revision as a sub-heading so anyone pinned to a SHA can still locate themselves.
- Two behaviour shifts from this wave added to "Behaviour worth re-checking" — they are not API breaks, which is what that section is for:
  - `AppendResult.producer_result` is now `ProducerStreamClosed` rather than `None` for a refused append to a closed stream (#126). HTTP response unchanged; only affects direct `store.append()` callers testing the field.
  - The dashboard JSON endpoint and SSE `Last-Event-ID` now answer 400 for a foreign-format offset instead of resolving it to an unrelated position (#125).

**No upgrade note for the retirements in #124.** `dispatch_external`, `recover_peak_snapshot` and `send_sse_event` were all added after `0.1.0`, so nobody upgrading from a tagged release ever had them. The changelog records the removal and says why no note is needed.

**Also here:** `.claude/worktrees/` added to `.gitignore`. Agent worktrees are transient checkouts that were not ignored, and a `git add -A` picked them up as embedded repositories.

**Verified locally, the full CI gate:** `ruff check`, `ruff format --check`, `pytest` (1306 passed), `zensical build` (no issues).

After merge, tag `v0.2.0` on the merge commit — `.github/workflows/publish.yml` gates the tag against `pyproject.toml`, so a mismatched tag cannot ship.

Issue #80, opened 2026-08-09
Consumer findings from formkit-ninja: skip raw saves, JSON-encode StreamEvent.data, soft-delete-aware decorator


joshbrooks:
Integrating rakaia as the audit log in formkit-ninja (catalpainternational/formkit-ninja#51, pinned at a7f0864) surfaced three issues in `@stream_model` / `create_stream_event` that every Django consumer will hit. Ordered by severity:

## 1. `handle_post_save` ignores the `raw` kwarg — fixtures corrupt streams

`django_rakaia/decorators.py` `handle_post_save` accepts `**kwargs` but never checks `raw`. Django passes `raw=True` for deserialized saves (`manage.py loaddata`, `serialized_rollback=True` test restores). Today every fixture row appends a phantom `create`/`update` event with a fresh offset — restoring a dump of N submissions writes N bogus audit events, and repeated test-DB restores multiply them. Raw instances can also reference FK rows that aren't loaded yet, so the handler can crash mid-`loaddata`.

Standard one-line guard at the top of the handler:

```python
if kwargs.get("raw"):
    return
```

## 2. `StreamEvent.data` is a plain `JSONField` with no encoder

`models.py` defines `data = models.JSONField()`. Any payload containing a `UUID`, `datetime`, or `Decimal` raises `TypeError` at insert time — from inside the consumer's `post_save`, i.e. it crashes the production save it was auditing. formkit-ninja currently works around this with a `json.dumps(..., cls=DjangoJSONEncoder)` → `json.loads` round-trip in every transformer, which means each payload is serialized three times per write and every future transformer must remember to pre-stringify.

```python
from django.core.serializers.json import DjangoJSONEncoder
data = models.JSONField(encoder=DjangoJSONEncoder)
```

removes the whole failure class for all consumers. (No schema migration impact — encoder is Python-side.)

## 3. No way to customize/opt out of the `post_delete` event — wrong for soft-delete models

`stream_model` unconditionally registers a `post_delete` handler that emits `event_type="delete"`. For a model with `pgtrigger.SoftDelete` (DELETE becomes `UPDATE is_active=false`, row survives), Django still fires `post_delete`, so the stream records a hard delete that never happened — with a payload snapshotting the pre-delete in-memory state (`is_active=true`) — while the real `is_active` flip is never streamed. formkit-ninja had to drop the decorator for `FormKitSchemaNode` and hand-wire both signals.

[...]

PR #133, opened 2026-08-15
Write an event's envelope in one place

### LINKED ISSUE #131: The envelope is written twice and the two copies disagree
There are two places in the Django integration that write an event to the
database, and they were written separately. They disagree about small things:
what to record when no change type is given, and whether to stamp the event with
a time. Neither is wrong exactly, but an event written through one door is not
quite the same shape as one written through the other.

The module that owns the envelope already carries a comment warning that a
second copy of this logic would drift. It has. Routing both doors through one
write would settle it, and the recently-shared append rules are the precedent.

joshbrooks:
There were two implementations of "write an event into the database": the durable store's append, and the `@stream_model` decorator that most Django apps actually use. They had drifted. The store records an event with no label under a stable marker that its reader knows how to undo again; the decorator wrote the label straight through, so an unlabelled event written by a model save could never be read back as unlabelled. The two also allocated offsets by different routes and worked out the metadata differently.

Both doors now go through a single writer that owns all of it — the marker, the metadata merge, the timestamp and the offset allocation. The writer takes a list of streams, so a save that fans out to several streams still records one event shared by all of them, which is what it always did. There is a new test that sends the same event in through both doors and compares what lands in the database; it failed before this change.

Closes #131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -292,6 +292,29 @@ protocol is covered by both stores.
 
 ### Changed
 
+- **One writer for the enveloped event** (#131). "Write an event into the Django
+  models" was implemented twice — once in `DjangoStreamStore._write`, once in
+  `create_stream_event`, the `@stream_model` decorator's door — and the two
+  copies had drifted. The store records a labelless append under a stable
+  `"append"` sentinel that its reader inverts back to "no label"; the decorator
+  wrote the label through raw, so an event from that door could never be
+  recognised as labelless. Each also called its own offset helper and resolved
+  `metadata` its own way. `django_rakaia.envelope`'s module docstring already
+  warned that a second copy of the envelope "produces events that replay
+  differently from every other event in the same stream, and no test anywhere is
+  looking at the difference" — this was that copy.
+
+  Both doors now write through `write_enveloped_event`, which owns the sentinel,
+  the ambient-`provenance()` merge, `{}` rather than NULL metadata, `event_ts`
+  passthrough, and locking offset allocation. It takes a *list* of streams, so
+  `@stream_model`'s fan-out is still one event with one envelope shared by N
+  entries. `tests/test_django_rakaia/test_envelope_writer.py` drives both doors
+  with the same envelope and compares the rows.
+
+  Behaviour is unchanged except that a `@stream_model` event with an empty
+  action string is now stored as `"append"` rather than `""` — the same shape
+  the store has always written, and the same thing both read back as.
+
 - **One producer response table instead of three.** The five-arm `isinstance`
   ladder over `ProducerValidationResult` — duplicate → 204, stale epoch → 403,
   bad epoch opening → 400, sequence gap → 409, closed stream → 409, each with
--- a/src/django_rakaia/decorators.py
+++ b/src/django_rakaia/decorators.py
@@ -16,8 +16,8 @@ from django.db import models, transaction
 from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 
-from django_rakaia.models import Stream, StreamEntry, StreamEvent
-from rakaia.context import merge_provenance
+from django_rakaia.django_store import write_enveloped_event
+from django_rakaia.models import Stream, StreamEvent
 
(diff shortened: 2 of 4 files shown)


joshbrooks:
## What the pinning test found

`tests/test_django_rakaia/test_envelope_writer.py` drives both doors — `DjangoStreamStore.append` and `create_stream_event` — with the same envelope and compares the `StreamEvent` rows and the read-back `StreamMessage`s. Of the three divergences #131 lists, exactly one was observable in the row:

- **The `"append"` sentinel — real, and now fixed.** `store._write` wrote `event_type = label or _APPEND_EVENT_TYPE`; `create_stream_event` wrote `event_type = action` raw. A labelless event from the decorator door landed as `""`, so `_read_since`'s inversion (`event_type == "append" → label = ""`) never matched it. `test_the_decorator_door_writes_the_sentinel` was the one red assertion.
- **`event_ts` — not a writer divergence.** `_write` passes `opts.event_ts` through (may be `None`); `create_stream_event` always stamps `time.time()`. That is a difference of *input*, chosen by the caller, not of writer behaviour: a raw protocol append genuinely has no logical time unless the producer set one, while `@stream_model` deliberately stamps one (ADR 0002 item 5, pinned by `test_decorators_provenance.TestEnvelopeTimestamp`). Forcing the store to stamp too would have changed protocol read-back semantics and diverged the durable store from the in-memory one under `tests/server_store_contract.py`. So the shared writer takes `event_ts` as an argument and passes it through, NULL included, and the test file says explicitly what it is *not* pinning.
- **Offset allocation — already unified.** `decorators._get_next_offset` was a one-line wrapper delegating to `Stream.get_next_offset`, the same method `_write` called. Two names, one implementation. The wrapper is deleted.
- **`metadata` — already agreed.** Both called `merge_provenance` and coalesced `None → {}`; the decorator just had no way to supply an explicit block. The merge now lives in the writer.

## The change

`write_enveloped_event(streams, data, *, label, metadata, event_ts, payload_encoding)` in `django_store.py`, next to the sentinel constant and its inversion. It owns the sentinel, the ambient-`provenance()` merge, `{}`-not-NULL, `event_ts` passthrough, and `Stream.get_next_offset`. It takes an *iterable* of streams and returns `(event, entries)`.

- `_write` calls it with `[stream]` per stored payload (JSON-mode array appends still write one event per element) and keeps only the row → `StreamMessage` mapping. It now reads `event.metadata` back off the written row rather than keeping its own merged copy.
- `create_stream_event` calls it once with all resolved streams, so fan-out stays **one** `StreamEvent` with N `StreamEntry` rows.

## Publish-on-append came free — because it already was

The SSE fan-out is a `post_save` receiver on `StreamEntry` (`channels_signals.handle_stream_entry_created`), which both paths already tripped, since both wrote entries with `.objects.create()`. `append_many` is the only path that doesn't (it uses `bulk_create` and calls `broadcast_entries` itself, #82). The shared writer still uses `.create()`, so the partition is unchanged and there is no doubled frame. `test_publish_on_append.py` and `test_channels.py` pass untouched.

## Preserved, non-obvious

- **The `raw=` guard** lives in the signal receiver, above `create_stream_event`, so it is unaffected. The *second* raw guard — in the `StreamEntry` `post_save` receiver, for fixture-restored entries — also still fires.
- **Transaction.** `create_stream_event` keeps its own `transaction.atomic()`, which is a savepoint when the caller already has one open, so the event stays inside the save that produced it. `get_next_offset` needs the enclosing transaction for its `select_for_update` on the high-water row. The `_get_or_create_stream` calls are evaluated inside the block (they are the writer's first argument).
- **Soft delete / `on_delete` / `delete_to_dataclass`** are entirely above the writer — untouched.
- **Payload encoding.** `create_stream_event` still round-trips the payload through `DjangoJSONEncoder` *before* handing it over, because the field encoder only covers the INSERT and the SSE receiver broadcasts the in-memory object (#80). The writer takes an already-encoded value.

No existing test was edited. Full suite 1316 passed; `ruff check`, `ruff format --check`, `zensical build` and all twelve `just demos` are green.

PR #135, opened 2026-08-15
Turn off the testing endpoint that let anyone break a stream

### LINKED ISSUE #130: Anyone can turn our server's errors on from outside
Our protocol server answers a special address that exists so the conformance
test suite can make requests fail on purpose. It is always switched on, it needs
no password, and it is checked before anything else the server does. A single
request from anyone who can reach the server makes any stream return errors for
as long as they like.

This has been harmless while the server was something you ran locally. We are
about to publish rakaia for other people to deploy, which is the point at which
it stops being harmless. It should be off unless you deliberately turn it on.

joshbrooks:
The server had a testing endpoint that anyone could call, with no password and no
check of any kind. One request could make any stream start failing, and keep it
failing, for everybody. That was harmless while the server only ever ran on
someone's laptop, but the package is being published now and people will deploy
it somewhere reachable.

The endpoint is now switched off unless you deliberately turn it on with an
environment variable. With it off the server answers as though the endpoint
simply does not exist, rather than saying "not allowed" — there is no reason to
tell a stranger it is there. Nothing we run today needs it turned on, and the
conformance results are unchanged.

Closes #130

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -326,6 +326,31 @@ protocol is covered by both stores.
 
 ### Fixed
 
+- **The fault-injection endpoint is no longer reachable on a deployed server**
+  (#130). `/_test/inject-error` was matched before any stream routing, with no
+  gate of any kind. An unauthenticated `POST` of
+  `{"path": "orders", "status": 500, "count": 999999}` made every subsequent
+  request to that path fail, indefinitely, for every client — a denial of
+  service any anonymous caller could aim at any stream. Harmless while the
+  server only ever ran on a laptop; not once the package is published and
+  someone deploys it.
+
+  The endpoint is now off unless `RAKAIA_ENABLE_FAULT_INJECTION` is set to a
+  truthy value, via a new `ServerOptions.enable_fault_injection` that reads it
+  the same way `long_poll_timeout` reads `LONG_POLL_TIMEOUT`. The environment
+  variable is what matters: `uvicorn rakaia:app` builds the module-level app
+  with default `ServerOptions()`, so a constructor argument alone could never
+  reach it. While off the path is not routed at all and answers `404` like any
+  other unknown path — deliberately not `403`, which would confirm that the
+  endpoint exists. Nothing needs the flag today: the upstream conformance suite
+  never calls the endpoint, so `conformance/run.sh` does not set it and the
+  baseline in `conformance/expected-failures.txt` is unchanged.
+
+  Six of `InjectedFault`'s ten fields were also stored and never read —
+  `delayMs`, `dropConnection`, `truncateBodyBytes`, `corruptBody`, `jitterMs`
+  and `injectSseEvent` — so a caller asking for a 200ms delay got no delay and
+  no error either. They are gone; `count`, `status`, `retryAfter` and `method`
+  are the four that ever did anything.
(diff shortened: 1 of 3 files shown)


joshbrooks:
Technical detail.

**What drove the endpoint: nothing.** The issue flagged this as worth resolving
first, since it decides whether `conformance/run.sh` needs the flag. I unpacked
`@durable-streams/server-conformance-tests@0.3.6` (the pinned version in
`conformance/package.json`) and grepped both `src/index.ts` and the built
`dist/`. The suite never references `/_test/`, `inject-error`, `retryAfter`, or
any fault-control path — its only "inject" hits are SSE CRLF-injection tests,
which are about payload escaping and do not touch the endpoint. Nothing in
`conformance/`, `tests/` or `examples/` calls it either. So `run.sh` does **not**
export the variable, and the endpoint ships dead unless a human turns it on.

**The gate.** `ServerOptions.enable_fault_injection` is a
`field(default_factory=_fault_injection_enabled_by_env)`, reading
`RAKAIA_ENABLE_FAULT_INJECTION` (`1`/`true`/`yes`/`on`, case- and
space-insensitive). A `default_factory` rather than `long_poll_timeout`'s
evaluated-at-import default, so the value is read per instantiation and is
testable with `monkeypatch.setenv`. The variable is namespaced deliberately: a
bare `ENABLE_FAULT_INJECTION` in a deployer's environment is exactly the
accidental enable this change exists to prevent.

The gate is on the routing condition itself (`handler.py:319`), so with the flag
off the request falls through to normal stream routing and 404s from
`_handle_append`/`_handle_delete` like any other unknown path — indistinguishable
from a path that was never implemented. Not a 403.

**Dead fields.** Verified by grep before removing, not taken on trust — the issue
listed five, but there were six. Removed: `delay_ms`, `drop_connection`,
`truncate_body_bytes`, `corrupt_body`, `jitter_ms`, and `inject_sse_event`, which
the issue's list missed; it was also written and never read. `_consume_fault` and
the fault branch read only `count`, `status`, `retry_after` and `method`. The
JSON keys for the removed fields are now silently ignored rather than stored,
which is the same observable behaviour as before — they never did anything.

**Conformance: the baseline did not move.** `269 passed, 63 failed, 6 skipped (of
338)`, `56 expected (known gap)` — byte-identical to a control run I did on a
clean `origin/main` tree in the same worktree. `expected-failures.txt` is
untouched.

Worth flagging separately: `check-regressions.mjs` reports **7 NEW failures on
`origin/main` as it stands**, all Stream-Seq ordering and idempotent-producer
tests (`should enforce lexicographic seq ordering`, `out-of-order seq values are
rejected`, and five more). Those are present with and without this branch, so
they are not from this change and I have left them alone — but the conformance
baseline is currently stale on `main` and someone should look. Happy to open an
issue if it isn't already tracked.

**Tests.** `TestFaultInjectionGate` asserts the 404 on a default app for both
POST and DELETE, that a POST to the disabled endpoint cannot fault a real
stream, and that with the flag on a fault still injects, still carries
`retry-after`, is still consumed after `count`, and is still cleared by DELETE.
`TestFaultInjectionEnvFlag` covers the env-var parsing on both sides. The
enabled-path assertions are the discriminator: they fail if the gate is wired
backwards.

PR #136, opened 2026-08-15
One rule for routing registrations, and a fix for handlers registered away from where they are defined

### LINKED ISSUE #117: Restoring registrations can import the wrong file
When rakaia restores a saved set of projection registrations, it works out which
files to re-read from the location of each registered function. That is only the
same place the registration was written if the two happen to sit in the same
file. Register a function that lives elsewhere — which is now the documented way
to attach a dependency to a handler — and the restore reads the file holding the
function, never the one holding the registration, so the registration silently
does not come back.

Nothing is lost or corrupted; the restored setup is just quietly missing
handlers, which is the kind of gap that shows up much later as a projection that
was never rebuilt. It needs a test either way, since nothing currently covers it.

### LINKED ISSUE #132: Two registries keep the same rule in step by hand
Handlers and upcasters both decide which events they apply to using the same
rule. That rule is written out twice, once in each registry, and a comment in
the second copy says outright that it mirrors the first. Resetting and restoring
are mirrored the same way, and the format registrations are saved in is defined
once but enforced separately in three places.

Nothing is broken today. The cost is that a change to how matching works has to
be made twice, correctly, by someone who notices the second copy exists. There
is a precedent for fixing exactly this in the same file, and it is documented as
having worked.

joshbrooks:
Handlers and upcasters were each answering the same question — what text does this registration get matched against? — in their own copy of the same code, with a comment promising the two were kept in step. Three other pieces of the registration record were duplicated the same way. Each of those now exists once, and the behaviour is unchanged: tests putting identical inputs through both sides went in first, so any silent change in what matches what would have failed them.

The second fix is a real bug. Restoring registrations after a restart works by re-importing modules and letting them register themselves again, and the module to import was worked out from where the handler function was written. That is the wrong module whenever a handler is wired up somewhere other than where it is defined — which is what the recommended way of giving a handler a dependency does. The handler was quietly missing after a restart, with nothing said about it. Registrations now record where they were registered, and that is what gets re-imported.

Closes #132
Closes #117

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -303,17 +303,33 @@ protocol is covered by both stores.
   warned that a second copy of the envelope "produces events that replay
   differently from every other event in the same stream, and no test anywhere is
   looking at the difference" — this was that copy.
-
   Both doors now write through `write_enveloped_event`, which owns the sentinel,
   the ambient-`provenance()` merge, `{}` rather than NULL metadata, `event_ts`
   passthrough, and locking offset allocation. It takes a *list* of streams, so
   `@stream_model`'s fan-out is still one event with one envelope shared by N
   entries. `tests/test_django_rakaia/test_envelope_writer.py` drives both doors
   with the same envelope and compares the rows.
-
   Behaviour is unchanged except that a `@stream_model` event with an empty
   action string is now stored as `"append"` rather than `""` — the same shape
   the store has always written, and the same thing both read back as.
+
+- **One content-routing rule, one registration record** (#132). Handlers and
+  upcasters each answered the same question — *what string does this
+  registration's glob get tested against?* — in their own method, the second
+  carrying a comment saying it mirrored the first. `reset()` and `rehydrate()`
+  were written out twice; the "a frozenset serialises as a sorted list" rule
+  three times; and `identity` / `to_payload` / `identity_from_payload` once per
+  record type, nine methods that had to agree pairwise with nothing checking
+  that they did. Content routing is now one function both registries call, the
+  meta-stream mechanics are one base class both registries inherit, and each
+  record type declares its persisted fields once and derives the round trip from
+  the declaration. `rakaia.registration_log` made the same move one level down;
+  its module docstring says why. Routing behaviour is unchanged, and pinned
+  first by tests that put identical inputs through both registries — including
+  the two edges that carry the weight: a missing key routes as `""` and does not
+  match, and a content-routed registration matched with no event raises.
+  Content routing for upcasters (ADR 0002 item 3) is untouched; this is where
+  that decision lands rather than a walk-back of it.
 
 - **One producer response table instead of three.** The five-arm `isinstance`
   ladder over `ProducerValidationResult` — duplicate → 204, stale epoch → 403,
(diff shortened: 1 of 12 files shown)


joshbrooks:
## Detail

### #117 — RED first

`tests/test_rakaia/test_registration_site.py` drives a real two-module split: `registration_sites/logic.py` holds the functions, `registration_sites/wiring.py` is the only module that calls a decorator, and it registers `functools.partial(logic.project_room, prefix="p")`. Against `main` the restore test failed exactly as #117 describes — `assert [] == ["room"]`, no error, no warning. The method case failed louder: `ModuleNotFoundError: No module named ...logic.Projector; ...logic is not a package`, from `rsplit(".", 1)[0]` on `pkg.mod.Class.method`. 8 of 13 red before the fix.

### #117 — the fix

`registered_in` on all three record types, taken from `_registration_site()` — walks out of rakaia's own frames so it lands on the caller whether they used `@register_handler`, `@register_simple`, or `registry.register()` directly. `RegistrationLog.modules()` reads it, falling back to the old `dotted_path.rsplit(".", 1)[0]` for payloads written before the field existed. The fallback is what keeps an existing meta-stream from re-appending on the first restart after upgrade: a same-module registration reconstructs the same value the fallback produces, so the identity still dedups. `dotted_path` is unchanged and still means "where the logic is", for drift reporting.

One knock-on: the `<locals>` warning's claim that `rehydrate()` "cannot restore" a closure is no longer accurate — restoration now depends on whether the registering module re-registers on import, which a module-scope closure does. Warning and glossary reworded to lead with the durable reason (the hash covers only the wrapper) and state the restore condition correctly. Still warns, still on `<locals>`, still only when persisting.

### #132 — what was consolidated

| was | now |
| --- | --- |
| `HandlerRegistry._match_subject` + `UpcasterRegistry._subject` | `_routing_subject()`, called by both. Callers pass only the wording of the no-event error, so each still names its own entry point. |
| `reset()` ×2, `rehydrate()` ×2, log load ×2 | `_LogBackedRegistry` base with `_logs`, `_load_logs`, `_reset_logs`, `rehydrate`. `UpcasterRegistry.rehydrate` gains a docstring it never had. |
| frozenset rule stated in `registration_log` and enforced in two spots in `registry` | `_event_match_payload` encodes; `_canonical_event_match` — already the argument normaliser — decodes. Reading a payload back is the same normalisation as accepting an argument. |
| `identity`/`to_payload`/`identity_from_payload` ×3 types = 9 methods | `_PAYLOAD_FIELDS: tuple[_PayloadField, ...]` per type; `_MetaStreamRecord` derives all three. Adding a field is one line. |

`identity_from_payload` became a `classmethod` (the protocol declares it as one now); `RegistrationLog` calls it off the type either way.

Two tests indexed identity tuples positionally (`ident[-1]`, `[6]`, `[7]`) — exactly the fragility `registration_log`'s docstring complains about. They now look the position up from the declared field list.

Pinning tests are `tests/test_rakaia/test_content_routing.py`, 31 cases including a parametrised block asserting the two registries match or don't match together on the same inputs. ADR 0002 item 3 is untouched.

### Judged and kept

- **`register_simple`** — kept. It is used by `examples/projection_cookbook` and referenced twice in `docs/examples.md`, and the docstring earns its length by saying when *not* to reach for it ("only when you actually need version brackets"). 13 docstring lines, not 30.
- **`upcast`** — kept. The pass-through is the point: it saves the caller from fetching the default registry to call `upcast_to_current` on it. That is the whole ergonomic difference between `upcast(event, "submissions")` and `get_default_upcaster_registry().upcast_to_current(event, "submissions")`.

Removing either would have broken docs and examples to save a dozen lines, and the tag not being pushed makes removal cheap, not free.

- **Empty section-header blocks** at the old `:549-555`, `:842-844`, `:880-883` — removed. Every remaining `# ===` header labels real content.

### Verified

`ruff check`, `ruff format --check`, `pytest` (1353 passed), `zensical build`, `just demos` — all clean.

PR #139, opened 2026-08-15
Compare Stream-Seq the way the protocol says

### LINKED ISSUE #137: Seven conformance tests fail and our check cannot tell us
The conformance suite reports seven failures against our server that are not in
the list of known gaps. They are all about the ordering rule for writer sequence
numbers, and they have been failing for a while — they predate all of this
month's work.

The reason nobody noticed is that the check reporting them is configured never
to fail. It runs, prints what it finds, and reports success either way, so a
green tick next to "conformance" on a pull request means the job finished, not
that the suite agrees with us. That needs deciding on two levels: whether we are
right about the ordering rule, and whether a check that cannot fail is worth
having in that shape.

joshbrooks:
The Durable Streams spec says the `Stream-Seq` header holds an opaque string, compared byte by byte. We stopped doing that in August, after a writer that sent 9 and then 10 was turned away. Comparing byte by byte, "10" comes before "9", so the refusal was right — but we changed rakaia to read the header as a number instead, and we also edited our copy of the spec to say numbers, crediting a revision that was never written. This puts both back.

If you send `Stream-Seq` as plain counting numbers, pad them: send "09" then "10", to whatever width you will reach, or use a ULID. Anything already padded, or not sending the header at all, is unaffected. Seven conformance tests were failing on this and now pass, taking us from 269 to 276.

Closes #137

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -636,12 +636,22 @@ protocol is covered by both stores.
   broadcast, after the write had landed. Anything outside the channel layer's
   allowed set is now replaced, and over-long names are truncated.
 
-- **Stream-Seq was compared as text, not as a number.** The header reached the
-  store unparsed, so the conflict check `opts.seq <= stream.last_seq` compared
-  strings: sending `Stream-Seq: 10` after `9` was rejected with a 409, because
-  `"10" < "9"` lexicographically. Every producer broke on reaching double
-  digits. The header is now parsed strictly (400 on a non-integer, matching the
-  producer headers) and both fields are typed `int | None`.
+- **`Stream-Seq` is an opaque string, compared byte-wise**
+  ([#137](https://github.com/joshbrooks/rakaia/issues/137)). Rakaia spent part
+  of this cycle parsing the header as a decimal integer, comparing it
+  numerically, and returning `400 Bad Request` for anything that was not one.
+  The protocol specifies none of that: `Stream-Seq` values are opaque strings
+  that **MUST** compare using simple byte-wise lexicographic ordering, so any
+  value is well-formed and a ULID is as valid as a digit string.
+
+  **If you send unpadded decimals, pad them.** Byte-wise, `"10"` sorts below
+  `"9"`, so `Stream-Seq: 10` after `9` is a `409 Conflict` — correctly. Send
+  `"09"` then `"10"`, at whatever fixed width your writer will reach; rakaia's
+  own offsets zero-pad for exactly this reason. Nothing else in the header's
+  behaviour changed, and a writer already padding or using a ULID is unaffected.
+
+  Seven conformance tests were failing on the numeric comparison. They pass now,
+  leaving the stream-forking family as the only known gap.
 
 - **`@stream_model` no longer appends phantom events for fixture loads** (#80).
   `handle_post_save` now honours Django's `raw` kwarg, so `manage.py loaddata`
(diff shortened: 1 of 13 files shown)


joshbrooks:
Technical detail.

**The comparison.** `AppendOptions.seq` and `Stream.last_seq` are `str | None` again; `append_decision.py` compares with `str <= str`, which is Python's byte-wise comparison and exactly the rule the spec states. Both stores read those fields through `StreamFacts` and neither assumed an int in its logic, so no store code changed apart from the durable column: `django_rakaia.Stream.last_seq` was `IntegerField` and is now `CharField(max_length=255)`, widened by migration `0009`. The column was only ever written by the protocol server and only ever held decimal digits, which survive the widening unchanged.

**The header.** `handler.py` was matching `Stream-Seq` against `STRICT_INTEGER_PATTERN` and returning `400 Bad Request` on anything else, which made a conforming value such as a ULID unusable. There is nothing to validate in an opaque string, so the header now passes through untouched. `Producer-Id`/`Producer-Epoch`/`Producer-Seq` still parse strictly — those are genuinely integers, they are a different mechanism (idempotent producers, section 5.2.1), and they are not in scope here.

**The vendored spec.** `docs/protocol.md` is a verbatim copy of upstream `PROTOCOL.md`, headed *Author: ElectricSQL*. Commit `9c07438` changed three places in it: the header type (`<integer>` for `<string>`), the layering summary (`numeric` for `lexicographic`), and the IANA header list (`decimal integer` for `opaque string`) — plus a parenthetical asserting that "an earlier revision of this document specified opaque strings under byte-wise lexicographic ordering". Grepping the current upstream document for `earlier revision`, `previous revision` and `compare numerically` returns nothing. All four are reverted to upstream's exact wording; nothing new was written.

**The internal inconsistency this resolves.** The argument for numeric comparison was that lexicographic ordering rejects `10` after `9`. That is true only of unpadded decimal. Offsets are the same kind of token under this protocol — opaque, lexicographically sortable — and rakaia already zero-pads them for precisely this reason: `src/rakaia/store.py:645` to 16 digits, `django_rakaia/offsets.py:21` to 20, both with comments citing the protocol's ordering requirement. The two positions could not both be right.

**Tests.** `test_seq_is_compared_as_a_number` becomes `test_seq_is_compared_lexicographically`, asserting both halves: `"10"` after `"9"` conflicts, and padded `"09"` then `"10"` is accepted. Its counterparts in the shared store contract (`tests/server_store_contract.py`, which runs against both stores) and in the Django store's own suite move the same way, and `test_non_numeric_seq_is_400` becomes `test_a_non_numeric_seq_is_accepted` with a ULID. Several tests were already passing `AppendOptions(seq="5")`, left over from the pre-#87 typing — those are correct again rather than merely tolerated.

**Conformance.** Before, on a clean tree: `269 passed, 63 failed, 6 skipped (of 338)` — `7 NEW failure(s) · 56 expected (known gap)`. After: `276 passed, 56 failed, 6 skipped` — `0 NEW failure(s) · 0 newly passing · 56 expected`. The seven that retired are the two lexicographic ordering cases, the case-sensitive ordering case, the two idempotent-producer cases that send `Stream-Seq` alongside producer headers, and the two property-based sequence-ordering cases. `conformance/expected-failures.txt` is unchanged — the seven were never in it, and nothing was baselined.

**Not addressed here.** The other half of #137 is that `conformance.yml:65` sets `continue-on-error: true` and never sets `CONFORMANCE_FAIL_ON_REGRESSION=1`, so the check reported pass throughout. With the baseline honest again, turning that on is now possible; it is a separate change.

PR #140, opened 2026-08-15
A failing conformance test should fail the build

joshbrooks:
Our conformance job was set up never to fail. It ran the protocol suite,
reported what it found, and then reported success regardless — so a green tick
next to it meant the job had finished, not that we passed. Seven genuine
failures sat there for weeks while every pull request showed conformance as
green.

Those seven are fixed now and the only remaining gap is one we have written
down deliberately, so the check can start meaning what it appears to mean. If it
goes red, either the change broke something or the divergence is intentional and
belongs in the list of known gaps with a reason.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -1,16 +1,19 @@
 name: Conformance
 
 # Runs the upstream Durable Streams server-conformance-tests suite against a
-# live rakaia server. It is currently INFORMATIONAL (non-blocking): the suite
-# step uses `continue-on-error` so failures are reported without blocking
-# merges while rakaia's protocol coverage matures (stream forking is not yet
-# implemented).
+# live rakaia server. It BLOCKS on a regression.
 #
 # Results are diffed against conformance/expected-failures.txt so a genuine
-# regression is distinguished from the known gap: NEW failures surface as
-# `::error::` annotations and a step summary, while expected fork failures stay
-# quiet. To make regressions block merges later, either flip `continue-on-error`
-# off below or set CONFORMANCE_FAIL_ON_REGRESSION=1 for the run step.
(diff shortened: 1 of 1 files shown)


joshbrooks:
**What changed:** removed `continue-on-error: true` from the suite step, set `CONFORMANCE_FAIL_ON_REGRESSION=1`, and added `shell: bash`. Also rewrote the file header, which still described the job as informational.

**`shell: bash` is load-bearing, not tidiness.** The step pipes through `tee`, and a bash pipeline reports the exit status of its *last* command. GitHub's default shell is `bash -e {0}` with no `pipefail`, so `./conformance/run.sh | tee` would return `tee`'s status — always 0 — and the job would keep passing no matter what the script exited with. That is the same silent-green failure this PR exists to remove, wearing a different hat. `shell: bash` runs with `-eo pipefail`.

**Verified in both directions**, because a gate that cannot fail is what we are fixing:

- Against the real baseline: `276 passed, 56 failed, 6 skipped (of 338)` — `0 NEW failure(s) · 56 expected (known gap)`, exit code **0**. Captured without a pipe so the code is the script's own.
- Against a baseline with one line removed from `expected-failures.txt`: `1 NEW failure(s)`, `Exiting non-zero: 1 regression(s) and CONFORMANCE_FAIL_ON_REGRESSION=1`, exit code **1**.

**Why now.** #139 retired the seven Stream-Seq failures, taking the baseline from `7 NEW` to `0 NEW`. Before that, switching this on would simply have painted CI red. The remaining 56 failures are the stream-forking family (#61), listed in `expected-failures.txt` and correctly reported as expected rather than new.

**Follow-on:** `conformance` is not currently a required status check on the branch protection rule, so this makes it red but not merge-blocking. Promoting it to required is a repo-settings change and deliberately not in this PR.

PR #142, opened 2026-08-15
One effect type per operation, instead of one type with an operation name

### LINKED ISSUE #141: One instruction shape per thing a projection does
When a projection wants to change the database, it fills in one form with
thirteen boxes. Which boxes are allowed depends on which of five things it is
doing, and nothing stops you filling the wrong ones until the code runs and
rejects it. We solve this correctly elsewhere in the same library — producer
outcomes are five small separate shapes, each carrying only what it needs — so
this is bringing one part up to the standard of another.

One of the five is not really the same kind of thing at all: it tells the
application to do something, not the database, and it currently travels to the
database writer mixed in with real writes. It should be handed back to the
caller instead. This changes something people have written code against, so it
needs an upgrade note, and it is much cheaper to do before the next release than
after.

joshbrooks:
Writing an effect used to mean picking one of five operations by name and then filling in whichever of thirteen fields that operation happened to use. The other ten did nothing, and five hand-written checks existed only to catch the combinations that made no sense. There is now one small type per operation, each carrying only its own fields, so those combinations can no longer be written down and every one of the checks is gone.

Sending an email is no longer treated as a kind of database write. It never was one — nothing ever applied it — so a replay now simply hands the emails back to whoever asked for the replay, to send or ignore. This is a breaking change for anyone writing handlers; UPGRADING.md has a before-and-after for each of the five old operations, and a consumer using only the three common ones has a purely mechanical rename.

Closes #141


joshbrooks:
## The target shape

```python
@dataclass(frozen=True)
class RowEffect:            # shared row identity, both now required
    model_label: str
    lookup: dict[str, Any]

class Upsert(RowEffect):    defaults: dict | None; produces: str | None
class Update(RowEffect):    defaults: dict | None          # never mints a row
class Delete(RowEffect):    spare: Exclude | SpareKeys | None
class Retire(RowEffect):    patch: dict; spare: SpareKeys | None; transition: Transition | None

Effect = Upsert | Update | Delete | Retire      # database
class ExternalEffect:       kind: str; payload: dict     # application
```

`Effect` stays exported as the union, so a `-> list[Effect]` annotation is
unchanged. `RowEffect`, `AnyEffect`, `Exclude`, `SpareKeys` and `Transition`
are new exports; `EffectOp` is gone.

## Cross-field validation reached zero

`Effect.__post_init__` enforced five rules. Where each went:

| Rule | Where it went |
|---|---|
| `produces` only on `update_or_create` | Gone — only `Upsert` has the field |
| `exclude` only on `delete` | Gone — only `Delete` can hold an `Exclude` |
| `transition_*` only on `retire` | Gone — only `Retire` has `transition` |
| `exclude` XOR `spare_keys` | Modelled away — one `spare` field, two shapes |
| `transition_kind` and `_key_fields` set together | Modelled away — one `Transition` |
| `transition_key_fields` non-empty | **Survives**, inside `Transition.__post_init__` |

The file now contains exactly one `__post_init__`, on `Transition`. The
executors also lost their `model_label is None or lookup is None` guards, since
`RowEffect` makes both required.

## Tests deleted, not rewritten

Seven tests asserting `ValueError` on a now-untypeable construction were
removed. Each one constructed an effect that no longer compiles, so there is
nothing left to assert:

From `tests/test_rakaia/test_effects.py::TestEffectValidation` (the whole class):

1. `test_exclude_and_spare_keys_together_rejected`
2. `test_exclude_on_retire_rejected`
3. `test_transition_kind_without_key_fields_rejected`
4. `test_key_fields_without_transition_kind_rejected`
5. `test_transition_kind_on_non_retire_rejected`
6. `test_exclude_on_update_rejected`
7. `test_transition_kind_on_update_rejected`

`test_empty_transition_key_fields_rejected` was kept and moved to a new
`TestTransition` class — it is the one rule that still exists.

In their place, `TestVariantFields` asserts positively that each variant carries
only its own fields (`assert not hasattr(eff, "patch")` and friends), so a field
creeping back onto the wrong variant fails loudly. `test_refs.py`'s
`test_produces_rejected_on_other_ops` became `test_produces_exists_only_on_upsert`
on the same principle.

One caveat worth recording: the issue says the guarantee "moves from a test to
the type checker", and it does — but `pyright` is `continue-on-error: true` in
CI and excludes `tests/`, so it is currently informational. That is a
pre-existing project decision, not something this PR changed. Pyright error
count is unchanged at 34 (all pre-existing, in `views.py` / `store.py` / a
`reducers_for_stage` stage arg).

## `external` left the Effect family

`ReplayResult.external_effects_skipped: int` → `ReplayResult.external:
list[ExternalEffect]`, carrying the effects themselves. `_apply_effects` now
splits the batch by *type* — externals to the result, the rest to the executor
— so no executor needs an external branch and `include_external` has nothing
left to mean. It is removed from `replay()`, `merge_replay()`,
`django_rakaia.replay.replay_stream`, and as `--include-external` from both
`manage.py replay` and the orders demo command.

`_synth_transitions` still manufactures one transition per retire flip; it now
appends to the same list rather than recursing through `_apply_effects` with a
gate. The `examples/multi_owner` external section keeps its dispatch loop —
that is the *application* routing them, which is the shape this change is
arguing for; the issue's `examples/multi_owner/executor.py` does not exist on
`main` (the branch it named was in `demo.py`).

## Other API changes

- `diff_effects_against_rows(ops=("update_or_create", ...))` → `kinds=(Upsert,
  Update)`, filtering by `isinstance` rather than string. No caller passed it.
- `types.Stream.messages` is gone. It was read in four places, all in
  `src/rakaia/store.py`; the in-memory store now keeps a `_messages: dict[str,
  list[StreamMessage]]` side map keyed by path, dropped on `delete()`/`clear()`.
  The durable store already documented the field as permanently empty.

## Migration mechanics

251 construction sites (16 `src`, 194 tests, 41 examples) were migrated with an
AST-based codemod — locate every `Effect(` call, read its literal `op=`, rebuild
the call from the source segments of the remaining keywords — then `ruff format`.
Eleven sites the codemod flagged (positional args, a non-literal `op`, calls
containing comments) were done by hand and reviewed individually.

## Verification

- `ruff check` / `ruff format --check` over `src/ tests/ examples/`: clean
- `pyright src/`: 34 errors, same as `main`
- `pytest`: 1372 passed
- `zensical build`: no issues
- `just demos`: all twelve pass
- `conformance/run.sh`: **0 NEW failure(s)**, 0 newly passing, 56 expected (the
  known stream-forking gap)

## Docs

`docs/glossary.md` gains `Upsert`, `Update`, `Delete`, `Retire`,
`ExternalEffect`, `Transition` and `Exclude`/`SpareKeys`, and its `Effect`,
`Idempotent` and `External effect` entries were rewritten. `UPGRADING.md` has a
new section under `# 0.2.0` with a before/after diff for all five old ops, the
`external` change and the `Stream.messages` change. CHANGELOG under
`## [0.2.0] - 2026-08-15`. Fifteen further docs and READMEs, the `okf/`
knowledge bundle, and the orders example's HTML templates were swept for the
old `op=` shape.

PR #144, opened 2026-08-15
Make the type checker fail the build again

### LINKED ISSUE #143: Our type checker reports problems and the build ignores them
We run a type checker in continuous integration, and it is configured never to
fail the build. It currently reports several dozen problems, and has done for a
while. It also does not look at our tests at all.

This matters more than it did last week. We have just reshaped the instructions
projections give the database so that invalid combinations cannot be written
down — and deliberately deleted the tests that used to catch them, on the
grounds that the type checker now does. That reasoning only holds if the type
checker is actually consulted.

joshbrooks:
The type checker was set to report its complaints and let the build pass anyway. Thirty-four of them had piled up that way, unnoticed, because nothing ever failed. That was fine when they were all noise about Django, but it stopped being fine once the library started relying on the type system to make invalid values impossible to write down — a check nobody is required to pass is not a guarantee.

All thirty-four are cleared and the step now fails the build. Most were Django attributes that a checker cannot see on its own, and those are now written down where they are used. Two were real problems worth fixing on their own: a function that could be handed a value it said it would never accept, and a check that let a wrong kind of value through to fail later with a confusing message.

Closes #143

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,10 +37,11 @@ jobs:
       - name: Format check with ruff
         run: uv run ruff format --check src/ tests/ examples/
 
-      - name: Type check with pyright (informational)
-        # Pre-existing Django ORM dynamism complaints; tracked separately.
-        # Migrate to django-stubs and flip this back to a hard failure.
-        continue-on-error: true
+      - name: Type check with pyright
+        # A hard gate since #143. Django's synthesised attributes (reverse
+        # accessors, implicit `id`, `annotate()` aliases) are declared on the
+        # models and at the query sites rather than waved through, so a new
+        # type error here is a real one. Keep it failing.
         run: uv run pyright src/
 
       - name: Run tests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -409,6 +409,31 @@ protocol is covered by both stores.
   offset N fails *after* the first N-1 have been dispatched, which is the
   partial-progress behaviour replay had before staging existed.
 
+- **Type checking is a CI gate, not a report** (#143). `pyright src/` ran with
+  `continue-on-error: true`, so its 34 errors were printed and discarded on
+  every green build. That mattered more after the `Effect` split: the guarantee
+  that an invalid effect is *unconstructible* is enforced by the type system and
+  by nothing else, so a checker CI ignores is a guarantee CI does not hold. The
+  step now fails the build.
+
+  Getting to zero needed no new dependency and no loosened settings. Django's
+  synthesised attributes — reverse accessors, the implicit `id` primary key —
+  are declared on `Stream`, `StreamEvent` and `StreamEntry` under
+  `if TYPE_CHECKING`, using the `RelatedManager` stub `django-types` already
+  ships; `annotate()` aliases, which no stub can know about because the query
+  invents them, are named in a `Protocol` beside the query that produces them.
+  Neither construction runs. The old admin idiom of assigning
+  `.short_description` onto a method is now `@admin.display(description=…)`,
+  which is what Django has documented since 3.2.
+
+  Two of the errors were worth having: `HandlerRegistry.reducers_for_stage` took
+  `stage: int` while replay's non-staged pass passed `None`, working only
+  because the comparison inside it never matched — the parameter is now
+  `int | None` with the rule stated; and `@stream_model`'s check that
+  `to_dataclass` returned a dataclass accepted a dataclass *class*, which then
+  failed obscurely inside `asdict` instead of with the clear message every other
+  wrong return already got.
+
 ### Fixed
 
 - **The fault-injection endpoint is no longer reachable on a deployed server**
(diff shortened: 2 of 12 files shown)


joshbrooks:
Technical detail.

**Baseline → result.** `PYRIGHT_PYTHON_FORCE_VERSION=latest uv run pyright src/`: 34 errors, 1 warning → **0 errors, 1 warning**. `continue-on-error: true` is off the pyright step in `.github/workflows/ci.yml`.

The surviving warning is `reportUnsupportedDunderAll` on `django_rakaia/__init__.py:104`, where `__all__ = sorted(_EXPORTS)` is computed from the lazy-import table. Pyright cannot evaluate it. Warnings do not affect pyright's exit code, so the gate is unaffected; the alternative is duplicating ~40 names as a literal list next to the dict that already holds them, which is a worse trade. Left alone deliberately.

**django-types vs django-stubs.** `django-types>=0.23.0` was already a dev dependency and was being picked up correctly — the CI comment's "migrate to django-stubs" diagnosis was wrong. Neither package can help here: reverse accessors and `annotate()` aliases are synthesised by `ModelBase` at class-construction time, and django-stubs recovers them only through its **mypy plugin**, which pyright cannot load. `django-stubs` under pyright would have produced exactly the same errors plus a dependency. What django-types *does* ship is `django.db.models.manager.RelatedManager`, stub-only, which exists precisely so the per-model declaration can be written by hand. No dependency change in this PR.

**Group A — 13 errors, `reportFunctionMemberAccess`, all `admin.py`.** `fn.short_description = "..."` → `@admin.display(description="...")`, available since Django 3.2 and we require >= 4.2. Checked for the sibling attributes: `admin_order_field` and `boolean` are not used anywhere in the repo, so `short_description` was the only one.

**Group B — 15 errors, `reportAttributeAccessIssue`.** Two kinds.

*Model-level* (10): `Stream.entries`, `Stream.producers`, `StreamEvent.entries`, and the implicit `id` on `Stream`/`StreamEvent`/`StreamEntry`. Declared in `if TYPE_CHECKING:` blocks in `models.py` using `RelatedManager["StreamEntry"]` etc. The block does not execute, so Django still builds the real descriptors and the real `AutoField` — nothing is shadowed and no migration is implied. This also fixed the errors in `views.py`, `streams.py`, `django_store.py` and both `channels_*.py` files at once, since they were all downstream of the same missing declarations.

*Query-level* (5, all `views.py:255-258`): `event_count` / `min_offset` / `max_offset` / `last_event` come from `.annotate()`, so they exist on the row and not on the class, and no stub can ever know them. Added an `_AnnotatedStream` `Protocol` naming the five aliases and their types, and `cast(Iterable[_AnnotatedStream], streams)` at the comprehension. `cast` is identity at runtime. `values()` would also have satisfied the checker but changes what the queryset yields, so it was rejected under the no-behaviour-change rule.

**Group C — 6 errors.**

1. `registry.py:636` `reducers_for_stage(self, stage: int)` → `int | None`. Real signature defect: `replay._run_stage_reducers` passes `None` on the non-staged path deliberately, and it only worked because `ReducerVersion.stage` is a non-negative `int` so `r.stage == None` never matched. The rule is now in the docstring — `None` is the single non-staged pass and correctly yields no reducers — so it is intentional rather than emergent.
2. `store.py:462` and `django_store.py:413`: `verdict.producer_result` is `| None` in general, but `close_stream_with_producer` declares its producer parameters as required, so `decide_append` always fences and always produces one. Added `assert ... is not None` with the invariant written above it, matching the existing `assert` style in `handler.py` / `channels_views.py`.
3. `decorators.py:96`: `dataclasses.is_dataclass()` narrows to `DataclassInstance | type[DataclassInstance]`, and `asdict` takes only the instance. Added `or isinstance(dc_instance, type)` to the guard. **Small behaviour change, confined to an error path**: a `to_dataclass` returning a dataclass *class* previously passed the guard and then died inside `asdict` with an opaque message; it now raises the same clear `TypeError` every other wrong return already got. Message reworded to "must return a dataclass instance"; nothing asserts on that string.
4. `hermeticity.py:113`: a genuine **stub bug**. `BaseDatabaseWrapper.execute_wrapper` is a `@contextmanager` at runtime, but the django-types stub annotates the *undecorated generator* — `-> Iterator[None]` — so `ExitStack.enter_context` rejects it. Fixed with `cast(AbstractContextManager[None], ...)` and a comment naming the stub defect, rather than an ignore.
5. `django_rakaia/replay.py:52`: `replay_stream(store: object | None)` was widened past the protocol it forwards to. Now `ReadableStore | None`, which is what `rakaia.replay.replay` requires. Runtime unchanged; the two in-repo callers already pass a real store or nothing.

**No blanket ignores, no new `# type: ignore` of any kind, and `typeCheckingMode = "standard"` / `exclude = ["tests"]` are untouched** (dropping the tests exclusion is issue #143's step 4 and stays out of scope).

**Verification, all green:** `pyright src/` 0 errors · `ruff check` · `ruff format --check` · `pytest` 1372 passed · `zensical build` · `just demos` all passed · `conformance/run.sh` 276 passed / 56 failed / 6 skipped, **0 NEW failure(s)**, 0 newly passing.

PR #145, opened 2026-08-16
Rewrite the docs around what the reader is trying to do

joshbrooks:
The front page was the most jargon-heavy page in the docs — denser than the glossary, whose job is jargon — and there was no tutorial at all. This regroups the navigation by what the reader is trying to do, adds the missing tutorial and a plain-language page on why the project exists, and generates the API reference so it cannot go stale.

It also fixes what the audit turned up along the way: docs naming a model that was deleted in migration 0008, a page still claiming a feature had not landed when it had, two links pointing at a repository that 404s, the strongest feature missing from the page that should describe it, and a stale exclusion in the local quality gate.

Technical detail is in a comment below.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,6 +18,25 @@
   then `uv run zensical build`. Without `--extra docs` that step fails with
   `Failed to spawn: zensical`.
 
+## Documentation
+
+- **`docs/api-reference.md` is generated. Do not hand-edit it.** It is produced
+  from `rakaia.__all__` / `django_rakaia.__all__` by
+  `scripts/gen_api_reference.py`. Run `just api-reference` and commit the result;
+  `just api-reference-check` fails if it has drifted. To move a name into a
+  different section, edit `GROUPS` in that script, not the Markdown.
+- **The nav in `zensical.toml` is grouped by reader intent** — *Start here*,
+  *How do I…*, *How it works*, *Look it up*, *Experiments* — not by subsystem.
+  The four Diátaxis modes were used to work out what each page is *for*; the
+  section names are the plain-English version. When adding a page, decide which
+  question the reader is asking and put it there.
+- **Main pages stay plain-language; technical detail goes in a trailing
+  `## Appendix` section on the same page.** Keeping the detail next to what it
+  qualifies is deliberate — it is what stops the caveats rotting away from the
+  claims they modify.
+- Research notes live in `docs/research/` and are deliberately **not** in the
+  nav. They are dated and are not decisions; decisions go in `docs/adr/`.
+
 ## Agent skills
 
 ### Issue tracker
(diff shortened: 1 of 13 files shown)


joshbrooks:
### What changed

**New pages**
- `docs/tutorial.md` — built as a real Django project in a clean venv against published `rakaia-streams` 0.2.0 and Django 6.1, and run end to end. Every command and every block of output in the page is the real thing. Ends on a deliberate bug in a handler, fixed by editing one line and replaying.
- `docs/why-rakaia.md` — the pghistory parity proof promoted out of "(spike)" and rewritten for a reader who does not have the vocabulary yet.
- `docs/api-reference.md` — 131 exported names, generated from `rakaia.__all__` / `django_rakaia.__all__` by `scripts/gen_api_reference.py`. `just api-reference-check` fails on drift. Output is deterministic; memory addresses are stripped from default-value reprs so the committed file does not churn between runs.

**Navigation** — grouped by reader intent rather than subsystem. Diátaxis was used to diagnose what each page was failing to be, not as a set of folders to sort into; its own guidance is against imposing the four sections top-down. Four of the five "(spike)" pages moved to `Experiments`.

**Corrections**
- `django-integration.md` documented an admin registration for `Translatable`, removed from the library by migration `0008`.
- `pghistory-retirement.md` said the envelope was "not yet part of rakaia core". All three of its "what making it first-class requires" items have shipped — `provenance()`/`label_marker()`/`append_if_changed()`, `StreamEvent.metadata` and `StreamEvent.event_ts` (`src/django_rakaia/models.py:262,265`), and `materialize_history()`/`history_effects()`.
- `zensical.toml` `site_url` and the social link both pointed at `joshbrooks/durable-streams`, which does not exist. Nothing is served at the new `site_url` yet either — publishing to Pages is still a separate step.
- `dry-run-and-executors.md` never mentioned `deny_database_access` or `assert_no_live_writes`. Both are now documented, including that the first *prevents* the query while the second *detects* afterwards, so outside a transaction the offending row has already landed when it raises.
- `justfile` excluded `typecheck` from `check` citing pre-existing pyright warnings, which #143 eliminated.

### Measurements

- `index.md`: 9.2 → 0.7 jargon terms per 100 words. New pages score 0.9 and 1.3.
- Every relative link in every page resolves — 0 broken, checked mechanically.
- Every documented `from rakaia …` / `from django_rakaia …` import resolves against the published 0.2.0.
- `ruff check`, `ruff format --check`, `pyright src/`, `pytest` (1372 passed), `zensical build` (no issues) all green.

### Also included

Two research notes under `docs/research/`, deliberately not in the nav, matching the convention set by `handler-types.md`. `demonstrating-value.md` carries a correction block at the top: it claims PyPI serves only 0.1.0, which is wrong — 0.2.0 was published 2026-08-15T15:58Z, and the note was written against a stale read of the index.

Issue #148, opened 2026-08-16
Test the row locking properly, with about nine tests rather than converting hundreds


joshbrooks:
The test suite runs in Django's fast mode, which wraps each test in a transaction and undoes it afterwards. Row locking does not work in that mode, and it does not work on SQLite either, so the code that stops two writers colliding is currently never exercised by anything. A Postgres test run has been spiked and the whole suite passes on it unchanged, so nothing was hiding — but the locking still is not being checked.

The obvious response is to move every test that touches a lock into the slower mode. That would be about 227 tests, and it is not worth it: only three lines in the codebase actually take a lock, and almost all of those tests merely pass through one on their way to doing something else. Around nine deliberate tests cover the same ground without making every future run pay for it.

The measurements and the proposed tests are in a comment below.

joshbrooks:
### Measurements

Taken by instrumenting `QuerySet.select_for_update` across a full run on `main` and recording the caller and the test's `django_db` marker. Independent of, and agreeing with, the spike's numbers.

**290 tests take a row lock; 227 run under plain `django_db`.**

Split by who asks for the lock:

| Origin | Tests reaching it |
|---|---|
| `src/django_rakaia/models.py:161` — offset watermark allocation | 199 |
| Django's own `QuerySet.update_or_create` | 123 |
| `src/django_rakaia/django_store.py:250` — stream row lock | 116 |
| `src/django_rakaia/effect_executor.py:211` — retire/transition | 8 |

160 non-transactional tests reach a lock rakaia asked for explicitly, but there are only **three** such call sites. The rest reach one incidentally — an append allocates an offset, and allocating locks.

### Why the fast mode hides two different problems

Per Django's own docs on `select_for_update()`:

> Evaluating a queryset with `select_for_update()` in autocommit mode on backends which support `SELECT ... FOR UPDATE` is a `TransactionManagementError` … If allowed, this would facilitate data corruption and could easily be caused by calling code that expects to be run in a transaction outside of one.

> Although `select_for_update()` normally fails in autocommit mode, since `TestCase` automatically wraps each test in a transaction, calling `select_for_update()` in a `TestCase` even outside an `atomic()` block will (perhaps unexpectedly) pass without raising a `TransactionManagementError`. To properly test `select_for_update()` you should use `TransactionTestCase`.

[...]

Issue #147, opened 2026-08-16
A rebuild safety check can quietly report a clean run it didn't verify


joshbrooks:
Before rebuilding tables from the log, you can switch on a check that makes any use of the live database an error, so a rehearsal that reports "nothing touched" has actually been verified rather than assumed. That check only watches the thread it was switched on in. The durable store hands its database work to a different thread, so anything done through that path slips past unnoticed and the run still reports clean.

The companion check that counts rows is not affected and still catches a leaked write from any thread, so following the documented advice to use both leaves you covered against the damaging case. What escapes is a rebuild secretly reading live data, which quietly undermines the claim that it was rebuilt from the log alone. Nothing in the codebase currently switches the check on around the threaded path, so this is a sharp edge rather than a live fault — but it fails silently, which is the shape worth fixing.

Detail, evidence and the options are in a comment below.

joshbrooks:
### Measured behaviour

`deny_database_access` installs a Django `execute_wrapper` on `connections[alias]` (`src/django_rakaia/hermeticity.py:79-123`). Django stores connections in a thread-local (`asgiref.local.Local(thread_critical=True)`), and `DjangoStreamStore.run_sync` dispatches ORM work via `sync_to_async(..., thread_sensitive=True)` (`src/django_rakaia/django_store.py:959-967`). The worker thread resolves the alias to a different connection object with an empty `execute_wrappers`.

Verified against published 0.2.0 + Django 6.1, one write pushed through each path:

| guard | same thread | via `run_sync` |
|---|---|---|
| `deny_database_access` | caught | **not caught** |
| `assert_no_live_writes` | caught | **caught** |

`assert_no_live_writes` is immune by construction: it compares `COUNT(*)` before and after (`hermeticity.py:200-215`), which measures the database rather than intercepting statements. The module docstring already directs callers to nest both, and describes the row-count guard as the one you can wrap a whole rebuild in — so the documented pattern still holds against leaked writes from any thread.

**Residual exposure:** a cross-thread *read* (determinism / hermeticity, not corruption), plus a cross-thread in-place `UPDATE` of existing rows — though the latter is already a documented limit of the row-count guard regardless of threading.

### Options considered

[...]

PR #149, opened 2026-08-16
Run the test suite against Postgres, so the row locking is actually exercised

joshbrooks:
The suite has only ever run on SQLite, which has no row locking. Everything the code does to stop two writers colliding was therefore never exercised — five comments in the source already say a Postgres run is what would catch it. This adds one, as a separate opt-in job so the existing fast run is untouched.

The good news is that nothing was hiding: the whole suite passes on Postgres with no changes to the library. It also adds two tests that genuinely race two writers against one stream, which pass on Postgres and are skipped with a stated reason on SQLite, and it records one real finding as a known failure — a rebuild safety check does not cover work the store hands to another thread, tracked in #147.

Detail in a comment below. The follow-up test plan is #148.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,65 @@ jobs:
       - name: Build docs
         run: uv run zensical build
 
+  test-postgres:
+    # A second run of the same suite against a real Postgres. It exists because
+    # SQLite makes `select_for_update()` a no-op: Django's SQL compiler emits
+    # `FOR UPDATE` only when `connection.features.has_select_for_update` is
+    # true, and the SQLite backend leaves it false. So on the `test` job above,
+    # every row lock in django_rakaia is silently skipped, and the
+    # `TransactionManagementError` that enforces "select_for_update must be
+    # inside a transaction" can never fire either.
+    #
+    # Deliberately a separate job, not extra steps on `test`: it needs a
+    # service container and takes longer, and the SQLite run is the one that has
+    # to stay fast and green.
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    services:
+      postgres:
+        # Pinned to a major on purpose. `postgres:latest` would move the suite
+        # onto a new major the day one is released, turning an unrelated PR red
+        # for reasons no one changed. 16 is the major this leg was developed
+        # and verified against, and it is in community support until late 2028,
+        # so the pin will not need touching soon. Bump it deliberately.
+        image: postgres:16-alpine
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+        ports:
+          - 5432:5432
+        # Without a health check the job can start psql-ing before the server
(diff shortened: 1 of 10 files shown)


joshbrooks:
### What it adds

- `tests/test_django_rakaia/settings.py` — `RAKAIA_TEST_DB=postgres` switches both aliases (`default`, `overlay`) to Postgres from standard `PG*` env vars; anything else keeps the existing in-memory SQLite. Each alias gets an explicit `TEST["NAME"]`, without which both derive the same test database name and the second clobbers the first.
- `pyproject.toml` — a `postgres` extra (`psycopg[binary]`), kept out of `dev` so the default run still needs no compiled driver.
- `.github/workflows/ci.yml` — a separate `test-postgres` job with a health-checked `services:` container. The existing `test` job is untouched.
- `justfile` — `pg-up` / `pg-down` / `test-pg` for the same run locally.
- `tests/test_django_rakaia/test_concurrent_appends.py` — two tests racing two writers on one stream.
- Conversions of `test_django_store.py` (all 5 markers) and two `test_alerts.py` classes to `django_db(transaction=True)`, since Django's docs are explicit that `select_for_update()` inside a plain `TestCase` "will (perhaps unexpectedly) pass without raising a `TransactionManagementError`".

**Postgres pin: `postgres:16-alpine`.** Pinned to a major so a new major shipping cannot turn an unrelated PR red; 16 is in community support until late 2028.

### Verified independently on this branch

| | result |
|---|---|
| `ruff check` / `ruff format --check` | clean |
| `pyright src/` | 0 errors (1 pre-existing `reportUnsupportedDunderAll` warning) |
| `pytest` (SQLite) | 1373 passed, 2 skipped, 1 xfailed — 32s |
| `pytest` (Postgres 16) | 1373 passed, 2 skipped, 1 xfailed — 76s |
| `test_concurrent_appends.py` on Postgres | both PASSED (not skipped) |

The 2 skips differ per backend and both are correct: on SQLite it is the two racing tests (`has_select_for_update` is false); on Postgres it is two `test_decorators_consumer_contract.py` tests whose emulated soft-delete trigger uses SQLite-specific `RAISE(IGNORE)`.

**No library changes were needed** — nothing in the existing suite was relying on SQLite's lack of locks.

### The one real finding

`test_hermeticity.py` gains `test_deny_database_access_should_block_a_write_through_run_sync`, `xfail(strict=True)` so it converts to a failure the day the guard is fixed, plus a control test proving the same-thread path holds. Full analysis and options in #147.

### Known rough edges

- Postgres teardown emits `PytestWarning: database "test_rakaia" is being accessed by other users` — a thread connection not closed at session end. Present before this branch too; harmless, worth tidying separately.
- 76s vs 32s. `--reuse-db` / `--create-db` wiring would close much of that gap and is deliberately out of scope here.
- Merges cleanly with #145 (verified with `git merge-tree`), though both touch `CLAUDE.md` and the `justfile`.

Follow-up test plan, including why converting the remaining ~227 lock-touching tests is the wrong trade, is #148.

PR #150, opened 2026-08-16
Make the rebuild safety check work across the store's worker thread

### LINKED ISSUE #147: A rebuild safety check can quietly report a clean run it didn't verify

joshbrooks:
Switching on the check that makes a rebuild rehearsal prove it touched nothing only ever watched one thread, because Django keeps a separate database connection per thread. The durable store deliberately hands its work to a different thread, so anything going through that route slipped past and the rehearsal still reported a clean run. It gave no warning when it happened, which is the part worth fixing — a check that quietly stops checking is worse than no check.

The store now asks what the caller is guarding and switches the same guard on inside its worker, so it means the same thing on both sides. A store reading its log from a guarded database will now raise instead of passing quietly, which is what the guard always said it would do. The companion row-counting check was never affected; the documentation claimed it was, and that is corrected here.

Closes #147. Detail in a comment below.

### DIFF
--- a/docs/dry-run-and-executors.md
+++ b/docs/dry-run-and-executors.md
@@ -149,10 +149,12 @@ with assert_no_live_writes(OrderSummary):
     `transaction.atomic()` if you need the write undone as well as reported, and
     prefer `deny_database_access` when you want prevention.
 
-    Both install themselves on the calling thread's connection. A write issued on
-    another thread — including ORM work handed off by the durable store's
-    `run_sync` — is **not** caught. See
-    [ADR 0003](adr/0003-handler-hermeticity.md).
+    Both install themselves on the calling thread's connection, and Django keeps
+    one connection per thread. `assert_no_live_writes` is unaffected — it
+    compares row counts, so it sees a write from any thread. `deny_database_access`
+    would miss one, so the durable store carries the guard across the one thread
+    hop it makes (`run_sync`); a write issued on a thread rakaia does not control
+    is still not caught. See [ADR 0003](adr/0003-handler-hermeticity.md).
 
 ### A rebuild that changes nothing is not a pass
 
--- a/src/django_rakaia/django_store.py
+++ b/src/django_rakaia/django_store.py
@@ -66,6 +66,7 @@ from rakaia.types import (
 )
 from rakaia.types import Stream as ProtocolStream
 
+from .hermeticity import armed_deny_aliases, deny_database_access
 from .models import (
     Stream,
     StreamEntry,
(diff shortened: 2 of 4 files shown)


joshbrooks:
### The mechanism

`deny_database_access` installs a Django `execute_wrapper` on `connections[alias]`. `ConnectionHandler` stores connections in `asgiref.local.Local(thread_critical=True)`, so the worker thread `DjangoStreamStore.run_sync` dispatches to (`sync_to_async(..., thread_sensitive=True)`) resolves the same alias to a *different* connection object with an empty `execute_wrappers` list. The query ran, committed, and the guard never fired.

### The fix

New `hermeticity.armed_deny_aliases()` reports which aliases the **calling thread** is currently guarding, by scanning `execute_wrappers` for the existing `_rakaia_deny_guard` tag. `run_sync` re-arms exactly those aliases inside the worker:

```python
denied = armed_deny_aliases()
if not denied:
    return await sync_to_async(fn, thread_sensitive=True)(*args, **kwargs)

def _guarded(*inner_args, **inner_kwargs):
    with deny_database_access(*denied):
        return fn(*inner_args, **inner_kwargs)

return await sync_to_async(_guarded, thread_sensitive=True)(*args, **kwargs)
```

Reading `execute_wrappers` does not open a connection — `connections[alias]` returns the wrapper object and the socket is opened lazily on first query — so the no-guard path costs one tuple comprehension over configured aliases.

### Why this shape, and not the alternatives from #147

Option **A** (patching `CursorWrapper._execute_with_wrappers` at class level) was prototyped and does work across all threads, but it patches a private Django API — a guard that breaks silently on a Django upgrade is the exact failure being fixed — and is process-wide, so it would block unrelated concurrent database work in the same process.

Option **B** (refuse to arm) was the recommendation in #147. Propagation turns out to be strictly better and no harder: it covers the same hop, and the loud failure B was after still happens naturally when the store's log lives on a guarded alias.

This covers the one thread hop rakaia controls. A hop made by consumer code is still outside the guard, and the docs now say so rather than claiming blanket coverage.

### Tests

- `test_deny_database_access_blocks_a_write_through_run_sync` — was `xfail(strict=True)` on the spike branch and is now a plain passing test. It also asserts nothing landed, so the guard is shown to *prevent* rather than merely report.
- `test_run_sync_guard_does_not_block_an_unguarded_alias` — **new**, the over-reach check. `DjangoStreamStore` has no `using=` and always works on the default alias, so the case proved is the mirror: guarding `overlay` must leave the store's default-alias work running normally across the same hop. Without it, a fix that simply froze the worker would pass the test above and be unusable in practice.
- `test_deny_database_access_does_hold_on_the_calling_thread` — retained as the control.

### Verified

| | result |
|---|---|
| `pytest` (SQLite) | 1382 passed, 2 skipped — no xfail remaining |
| `pytest` (Postgres 16) | 1382 passed, 2 skipped |
| `ruff check` / `format --check` | clean |
| `pyright src/` | 0 errors |
| `zensical build` | no issues |

The `docs/dry-run-and-executors.md` warning added in #145 said a write on another thread is "**not** caught" by either guard. That was wrong about `assert_no_live_writes` — it compares row counts, so it always saw cross-thread writes — and is now stale about `deny_database_access`. Corrected in this PR.

PR #151, opened 2026-08-16
Test the row locking deliberately instead of incidentally

joshbrooks:
Three places in the package take a row lock, and until now nothing put any of them under contention. Around 290 tests touch a lock, but nearly all of them reach one in passing on the way to something else, so none would notice if the locking quietly stopped working. Converting all of those to the slower test mode would cost every future run and still not test the lock; nine deliberate tests do.

Every test here was checked by deleting the lock it covers and confirming it fails. That check earned its keep: it found that the existing watermark test passed with its lock removed, because it raced the very first allocation where a unique key does the serialising instead, and that two tests written for this change also passed either way because the two database calls they were trying to interleave happen back to back. Both are fixed rather than papered over.

Refs #148. Detail in a comment below.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,6 +19,17 @@
   `django_db(transaction=True)`; a plain `django_db` test runs inside a
   transaction pytest-django rolls back, so a lock taken outside the code's own
   `atomic()` still looks fine and a second connection can never see the rows.
+- **Row locking is covered deliberately, not incidentally.** There are three
+  `select_for_update()` sites (`models.py` offset watermark, `django_store.py`
+  stream row, `effect_executor.py` retire capture). Around 290 tests touch one,
+  but nearly all reach it in passing — an append allocates an offset, and
+  allocating locks. The cover is `test_concurrent_appends.py` and
+  `test_locking.py`; converting the incidental ~227 to `transaction=True` was
+  considered and rejected (#148), because it buys nothing those files do not
+  and makes every run pay truncation teardown. **A new test that exercises a
+  lock belongs in one of those two files, marked `transaction=True`, and must
+  be shown to fail with the lock removed** — two earlier attempts passed with
+  and without it, which is the failure mode these tests exist to avoid.
 - **pyright is a hard gate** and `src/` is expected at zero errors. Run it as
   `PYRIGHT_PYTHON_FORCE_VERSION=latest uv run pyright src/` locally — plain
   `uv run pyright` can object to its own pinned version. Django's synthesised
--- a/tests/test_django_rakaia/test_concurrent_appends.py
+++ b/tests/test_django_rakaia/test_concurrent_appends.py
@@ -87,8 +87,20 @@ def test_a_second_writer_blocks_on_the_offset_watermark() -> None:
     it is not, B reads the pre-A value and hands out the offset A already
     took -- two events at the same offset in the same stream, which the
     unique constraint would reject at insert time in production.
+
+    **The first append is load-bearing.** Without it this test passes even with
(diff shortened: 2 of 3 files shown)


joshbrooks:
### The three lock sites

| Site | What it protects |
|---|---|
| `models.py` — `Stream.get_next_offset_block` | two writers never receive the same offset |
| `django_store.py` — `_live_stream(..., for_update=True)` | a writer's closed / content-type / producer checks cannot be invalidated between reading them and acting on them |
| `effect_executor.py` — `DjangoExecutor._retire` | the flip set a retire reports is the set its `UPDATE` actually flipped |

Measured on `main` by instrumenting `QuerySet.select_for_update`: 290 tests take a lock, 227 under plain `django_db`. Split by caller — 199 via the watermark, 123 via Django's own `update_or_create`, 116 via the stream row, 8 via retire. Three lines, ~290 tests: the coverage was incidental, and mass conversion would have bought none of what is below.

### What was added

**`test_locking.py`** (new, 7 tests):
- `TestOpensItsOwnTransaction` — three tests, one per site. With `transaction=True` there is no ambient transaction to inherit, so removing the library's own `atomic()` surfaces as `TransactionManagementError`. This is the cover the ~160 incidental non-transactional tests were notionally providing, and it collapses to three.
- `TestStreamRowLock` — an append must not overtake a committed close; two concurrent closes settle on exactly one winner.
- `TestRetireLock` — a concurrent resolve cannot split the reported flip set; two competing retires cannot both claim the same rows.

**`test_concurrent_appends.py`** — the existing forced-interleave test now seeds one append before racing (see below).

**`CLAUDE.md`** — records the policy and requires new lock tests to be shown failing with the lock removed.

### Mutation results

Each mutation was confirmed to have applied before the run — an earlier pass reported two false "not caught" results that turned out to be `str.replace(..., 1)` hitting the wrong method.

| Mutation | Caught by | Result |
|---|---|---|
| watermark `select_for_update()` removed | `test_concurrent_appends.py` | 1 failed |
| stream row `select_for_update()` removed | `TestStreamRowLock` | 2 failed |
| retire `select_for_update()` removed | `TestRetireLock` | 2 failed |
| `atomic()` removed from `append` | `TestOpensItsOwnTransaction` | 2 failed |
| `atomic()` removed from `apply` | `TestOpensItsOwnTransaction` | 1 failed |

### The two holes the mutation check found

**The existing watermark test passed with its lock removed.** It raced the first ever allocation of a stream, where the watermark row does not exist yet — so the second writer races the first one's *insert*, blocks on the unique key, retries, and reads the committed value. The constraint was doing the serialising, not the lock. Seeding one append first makes the row exist, and the test now fails when the lock goes.

**Two retire tests written for this PR also passed either way.** `apply()` issues its capturing `SELECT` and its `UPDATE` back to back, so two plainly-scheduled threads never interleaved between them. They now patch `QuerySet.update` to pause on the retiring thread only, holding the window open long enough for the competing writer to reach it. A race test that cannot lose is not a test.

### Verified

| | result |
|---|---|
| `pytest` (SQLite) | 1380 passed, 9 skipped, 1 xfailed — 17s |
| `pytest` (Postgres 16) | 1387 passed, 2 skipped, 1 xfailed — 76s |
| 5 consecutive runs of both locking files | 9 passed each time, no flakes |

All 7 new tests skip on SQLite with the reason stated — the backend reports `has_select_for_update = False`, so `select_for_update()` compiles to nothing and, per Django's docs, *"an error isn't raised"* either. Neither failure mode is observable there.

The remaining `xfail` is the guard leak, fixed separately in #150.

Issue #152, opened 2026-08-16
Several rules in the codebase are written in more than one place


joshbrooks:
An architecture review looked at the parts of the code that change most often. The good news is that the overall structure holds up: the split between the core package and the Django one is clean, the shared test suites genuinely cover more than one implementation, and almost nothing in the test suite reaches around a module to poke at its insides.

What it did find is one pattern repeated eight times: a rule that has more than one implementation, and the copies have drifted apart. Deciding whether a write is allowed happens in three places and is skipped in a fourth. Turning a stored row back into an event is written six times, and the three copies outside the store are wrong. Each gap is small on its own; together they account for four defects a user could hit today. Every one of them now has a failing test in the repository proving it is real, so none of this has to be taken on trust.

Sub-issues are listed below. Technical detail is in a comment.

joshbrooks:
### Sub-issues

Ordered by what I would tackle first.

- [x] #153 — one row-to-event translation *(Strong · 2 failing tests · start here)*
- [x] #156 — hash a rule's source once per rebuild, not once per event *(Strong · cheapest win, ~7× faster)*
- [x] #154 — one append pipeline, so the batch door cannot skip admission *(Strong · data integrity)*
- [x] #155 — keep wire framing out of the stored event *(Strong · rebuild is broken for content-typed streams)*
- [x] #157 — publish from the store, after commit *(Strong shape; the re-timing half is Worth exploring)*
- [x] #159 — thread the database alias through the write path *(Strong)*
- [x] #160 — move the value-equality rule where both sides can reach it *(Worth exploring)*
- [x] #158 — make a read return a decided answer rather than sending it *(Worth exploring · no defect, shape only)*
- [ ] #161 — the smaller items *(Speculative)*

### Method

Two independent read-only surveys, then every runnable claim re-verified against this checkout. That verification mattered — three findings came back **worse** than first reported, and one was **wrong**:

- Rebuilding cannot read *any* content-typed stream, not just batched ones. A single append is enough.
- Re-hashing is ~86% of rebuild time, not the ~72% first estimated.
- The batch door does not consult the admission rule *at all* — it was first described as having a partial copy.
- The claim that live updates mangle `text/plain` payloads was **wrong**; that case round-trips correctly. Only payloads stored encoded (non-UTF-8) diverge. The test was rewritten to target the case that actually breaks.

### The proofs

[...]

Issue #161, opened 2026-08-16
Smaller items from the same review


joshbrooks:
Four minor findings that are real but not worth an issue each: importing the rebuild module gives you a function instead of the module, which silently misdirects debugging tools; the code that substitutes references guesses at field names by string, so a renamed field would fail quietly rather than loudly; the shared store lookup is a module-level value with no way to replace it, which is why one entry point has no test coverage at all; and the notification payload for a resolved item is invented in one place while its shape is described in another.

Also worth noting separately: the 92 lines that decide which of two competing writers wins have no tests of their own and are only exercised indirectly. That is the cheapest thing in the package to pin down properly and among the most important to get right.

Part of #152. Detail in a comment.

joshbrooks:
Four small items plus one coverage gap. None individually worth an issue; grouped so they are not lost.

**1. `rakaia.replay` the function shadows `rakaia.replay` the module.** `__init__.py:87` does `from .replay import … replay`, rebinding the package attribute. Verified:

```python
import rakaia.replay as rp
rp.build_pipeline   # AttributeError: 'function' object has no attribute 'build_pipeline'
```

`from rakaia.replay import X` and `sys.modules["rakaia.replay"]` both still work. It cost me a wrong measurement while verifying #156 — a monkeypatch landed on the function object and silently reported zero calls. Cheapest fix is a note in the module docstring; renaming is breaking. **Test:** `test_importing_the_replay_submodule_yields_the_module`.

**2. `RefResolver` guesses fields by string.** `effects.py:296`, `:336-372` walk effects with `_MAPPING_FIELDS = ("lookup", "defaults", "patch")` plus `hasattr`/`getattr`, reintroducing the "which fields does this variant have?" reasoning that one-type-per-operation removed — and the comment at `:295` concedes it. A renamed field or a new variant fails *silently*: `hasattr` returns False and refs go unresolved. Declaring ref-bearing fields per effect type makes it a pyright error instead.

**3. `get_store()` is a module global with no reset.** `store.py:19-67` memoises into `_stores`. `test_store_selection.py:31` mutates that private dict and `:95` asserts on it — the only place in the suite that reaches past an interface by design. Other tests call `get_store().clear()`, a method only the in-memory store has, quietly pinning them to that backend. And because `replay.py:52` and `integration.py:29` resolve the store internally rather than accepting one, `integration.py` cannot be driven at all — which is why it has no coverage.

[...]

Issue #160, opened 2026-08-16
Two parts of the system can disagree about whether a value changed


joshbrooks:
Deciding whether a stored value is really different from a new one is a single question, and the answer is used in two places: when skipping writes that would change nothing, and when checking a rebuild matches what is already there. The rule lives inside the checking code, and only the checking side can be handed a custom version of it.

So a rebuild verified as identical under one set of rules can be rewritten under another. Both sides already have comments pointing at this, and neither can fix it from where it sits. Moving the rule somewhere both can reach lets them be given the same one.

Part of #152. Detail in a comment.

joshbrooks:
**Files** — `effect_executor.py:49` (`from .verification import canonical_value`), `:176-181` · `verification.py:230-400`

**The layering.** `canonical_value` and `DEFAULT_NORMALIZERS` answer a domain question — what the log's representation of a column *means* — and both the write path (`skip_unchanged`) and the verify path depend on the answer. Because they live in `verification.py`, the executor's dependency arrow points at the verification module.

**The consequence.** `diff_effects_against_rows` accepts `normalizers=`; `DjangoExecutor.__init__` takes only `skip_unchanged` and `using`. So there is no way to construct an executor that agrees with a diff run under a custom normalizer set. Both docstrings already flag the divergence (`verification.py:373-374`, `effect_executor.py:176-181`) and neither can close it from where it sits.

The three normalizers are also never named directly by any test — they are only reached through `canonical_value`.

**Deletion test** — moves rather than concentrates. The concentration is already achieved; this changes *where* it sits. Reported because the misplacement is what blocks closing the divergence.

**Shape.** Before: `effect_executor` → `verification` → `projection_reader`, the write path importing the verify path. After: `effect_executor` → `canonicalisation` ← `verification`; both depend on the rule, neither on the other.

**Test note.** `test_the_skip_rule_can_be_given_the_normalizers_a_diff_uses` is the one spike in this epic that asserts a *desired* interface (a `normalizers` parameter on the executor) rather than an existing claim, because the divergence cannot be demonstrated without one. Worth knowing when reading it.

Issue #159, opened 2026-08-16
Saving to a second database splits the write across two


joshbrooks:
A project can tell Django which database a save should go to. When that happens, the row goes where it was told and the event recorded alongside it goes to the default database instead, because the instruction is discarded on the way through.

The same gap blocks something the documentation offers. Rebuilding tables safely is supposed to be possible against a scratch copy of the database, and the tooling can already read and write projections that way — but nothing on the recording side accepts the instruction, so that option cannot actually be used. The existing safety check even names this exact situation as the leak it exists to catch, while the package ships no way to prevent it. Failing test included.

Part of #152. Detail in a comment.

joshbrooks:
**Files** — `decorators.py:236-250` (receiver discards `kwargs["using"]`), `:133` (bare `transaction.atomic()`) · `django_store.py:133-188` and every `Model.objects` in that file · `models.py:160-172` (offset allocation) · `hermeticity.py:44-47`, `:238-242`

**The seam stops half-way.** `DjangoExecutor` (`effect_executor.py:73`) and `DjangoProjectionReader` (`projection_reader.py:35`) both take `using`. The write half — `@stream_model`, `create_stream_event`, `write_enveloped_event`, the offset allocator, `DjangoStreamStore` — has no `using` anywhere. `DjangoStreamStore` has no constructor at all.

**Two concrete consequences.**

1. A save routed to another alias writes its model row there and its stream event to `default`. The failing test asserts the event count on `default` is unchanged after `area.save(using="overlay")`; it is not.
2. `hermeticity.py:44-47` tells the reader a hermetic rebuild may read its log from "a store on another alias" — and no such store can be constructed. `test_hermeticity.py:174-176` currently pins that absence as intentional.

**The pointed bit.** `assert_no_live_writes` names the exact leak it exists to catch — *"a receiver that saves without a `using=`"* (`hermeticity.py:238-242`) — and the receiver in question is this package's own `@stream_model`. The package ships the detector and not the seam that would prevent it.

⚠ Adjacent to **ADR 0003** but does not contradict it: the ADR offers both an in-memory store and a store on another alias, and only the first exists. Flagging because the docs promise the option in prose while a test pins it closed.

**Deletion test** — concentrates. Without it every adopter re-derives rebuild-time signal suppression in their own tree, which is precisely the history ADR 0003's 2026-08-14 update records for the guards themselves.

[...]

PR #162, opened 2026-08-16
Pin the architecture review's findings as failing tests

joshbrooks:
An architecture review found eight places where a single rule has more than one implementation and the copies have drifted apart. Rather than leave that as a written claim that ages badly, this adds one test per finding that fails today for exactly the reason stated.

Each is marked as a strictly-expected failure, so the suite stays green while the work is outstanding, and the marker flips into a real failure the moment someone fixes the underlying problem. No behaviour changes here and no source file is touched — this is the evidence, not the fix.

Refs #152. Detail in a comment.

### DIFF
--- a/tests/test_django_rakaia/test_architecture_spikes.py
+++ b/tests/test_django_rakaia/test_architecture_spikes.py
@@ -1,0 +1,274 @@
+"""Executable proof for the django_rakaia findings of the 2026-08-17 architecture review.
+
+Same contract as `tests/test_rakaia/test_architecture_spikes.py`: every test here
+**fails today**, is marked `xfail(strict=True)`, and asserts a property the code
+already claims somewhere — in a docstring, in `docs/glossary.md`, or in a sibling
+implementation. When a finding is fixed its test reports XPASS(strict), which is
+a failure telling you to delete the marker.
+
+Epic: #152.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import pytest
+from django.db import transaction
+
+from django_rakaia.django_store import DjangoStreamStore
+from django_rakaia.models import StreamEntry, StreamEvent
+from rakaia.types import AppendOptions
+
+pytestmark = pytest.mark.django_db
+
+
+class _RecordingChannelLayer:
+    """A channel layer that records what was published, and when.
+
+    `broadcast_entries` is documented as a no-op when the channel layer is
+    absent, so `get_channel_layer` is already a seam — this is its second
+    adapter.
+    """
+
+    def __init__(self) -> None:
+        self.sent: list[tuple[str, dict[str, Any]]] = []
+
+    async def group_send(self, group: str, message: dict[str, Any]) -> None:
+        self.sent.append((group, message))
+
+
+@pytest.fixture
+def recording_layer(monkeypatch: pytest.MonkeyPatch) -> _RecordingChannelLayer:
+    layer = _RecordingChannelLayer()
+    import django_rakaia.channels_signals as cs
+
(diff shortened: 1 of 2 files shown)


joshbrooks:
### What is here

Two files, no source changes:

| File | Tests | Findings |
|---|---|---|
| `tests/test_rakaia/test_architecture_spikes.py` | 4 | #155, #156, #161 |
| `tests/test_django_rakaia/test_architecture_spikes.py` | 6 | #153, #154, #157, #159, #160 |

`1382 passed, 9 skipped, 10 xfailed` · ruff and ruff format clean.

### Why `xfail(strict=True)` rather than a skip or a TODO

A skipped test proves nothing and a comment rots. A strict expected-failure runs on every commit, so if the underlying defect is fixed — deliberately or by accident — CI says so immediately with an XPASS. It is the same pattern this repo already used for the hermeticity gap in #149/#150, and that one did its job.

### Each test asserts an existing claim, not a wishlist

Deliberate constraint. Examples:

- `read()` inverts the `"append"` sentinel and *documents* that it does so to match the in-memory store — so the SSE frame not doing it is a divergence from a stated property, not a matter of taste.
- `write_enveloped_event`'s docstring says it is "the only writer of one".
- `server_store_contract.py:674` is *named* `test_a_json_array_append_flattens_into_separate_messages`.
- `hermeticity.py:238` names "a receiver that saves without a `using=`" as the leak it detects.

The one exception is #160's, which asserts a `normalizers` parameter the executor does not have — the divergence cannot be demonstrated without one. Called out in that issue.

### Writing the tests changed the findings

This is the part worth reading. Three findings were understated and one was wrong:

**#155 is worse.** Reported as a JSON-array problem. Actually `process_json_append` adds a trailing comma to *every* append on a content-typed stream, so one ordinary append is enough:

```
ValueError: Cannot decode event at seq=0 in stream='t' as JSON: Extra data: line 1 column 8
```

**#156 is worse.** ~86% of rebuild time re-hashing, not ~72%. 2000 events: 146.0 ms → 18.2 ms memoized.

**#154 is worse.** Reported as a partial copy of the admission rule. `append_many` does not call `decide_append` at all — the returned `AppendResult` carries `producer_result=None`, so the fencing decision was never made.

**#153 was partly wrong.** The claim that `text/plain` payloads reach subscribers mangled is **false** — that path round-trips correctly, because `encode_payload` stores text verbatim and `decode_payload` reverses it symmetrically. My first version of the test XPASSed, which is exactly what strict mode is for. Rewritten against the base64 path (bytes that are not valid UTF-8), where the divergence is real:

```
frame data = b'//4gYmluYXJ5...'   vs   read() data = b'\xff\xfe binary, not utf-8'
```

**#158 has no test at all.** The survey suggested the ETag rule had drifted between GET and HEAD. It has not — HEAD legitimately uses `-1` and the current offset because it has no request offset. I could not construct a failing case, so none is claimed and the issue says so. Same for the `a and b or c and d` predicate, which Python parses as intended.

### Verification of the verification

Every test was run with `--runxfail` to confirm it fails for the *stated* reason and not incidentally. That caught one broken test of my own — the fencing spike initially failed with `TypeError: AppendOptions.__init__() got an unexpected keyword argument 'epoch'`, i.e. my API guess, not the defect. The real fields are `producer_epoch` / `producer_seq`.

### Not included

The `_RecordingChannelLayer` in the Django file is a second adapter for the `get_channel_layer` seam, which `broadcast_entries` already treats as one ("a no-op when the channel layer is absent"). It is deliberately local to the spike file rather than promoted to a shared fixture — one use is not yet a shared fixture's worth of justification.

PR #163, opened 2026-08-17
Give a stored row as an event one definition

### LINKED ISSUE #153: Subscribers are told a different story than readers
Turning a saved row back into an event — working out what actually happened, and what the payload was — is written six times across the codebase. The three copies inside the store agree with each other. The three outside it, used for live updates and the dashboard, do not.

Two consequences a user can see. A live subscriber receives an internal marker word as the change label, where anyone reading the same event back from the store correctly sees no label at all. And a payload that had to be stored encoded arrives at the subscriber still encoded, with nothing to say it should be decoded, while a reader gets the original bytes. Both are proven by failing tests in `tests/test_django_rakaia/test_architecture_spikes.py`.

joshbrooks:
Working out what a saved row actually represents means undoing three things storage does to it. That work was written six times across the codebase: three copies inside the store, which agreed with each other, and three outside it, which did not.

Live subscribers were the ones affected. They were sent an internal placeholder as the change label, where anyone reading the same event back correctly saw no label, and they were sent an encoded payload with nothing to say it was encoded. All of it now goes through one function, and the two failing tests that proved the problem now pass with their markers removed.

Closes #153. Stacked on #162, which holds the evidence. Detail in a comment.

### DIFF
--- a/src/django_rakaia/channels_signals.py
+++ b/src/django_rakaia/channels_signals.py
@@ -5,13 +5,17 @@ Replaces the custom SSEManager/Unix socket/Redis broadcasting with
 Django Channels' channel layer for cross-process event distribution.
 """
 
+import base64
+import json
 from collections.abc import Sequence
+from typing import Any
 
 from asgiref.sync import async_to_sync
 from channels.layers import get_channel_layer
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
+from .event_message import message_of
 from .models import StreamEntry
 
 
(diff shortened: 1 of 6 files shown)


joshbrooks:
### The shape

New `src/django_rakaia/event_message.py` holds the single translation plus the payload helpers it needs (`encode_payload` / `decode_payload` / `APPEND_EVENT_TYPE` moved out of `django_store`, where nothing else imported them and they were not exported). It sits **below** both `django_store` and `channels_signals` deliberately — `_publish` already imports `broadcast_entries` lazily to dodge a cycle, so the translation could not live in either without one.

`message_of(entry) -> StreamMessage` is now the only thing that reads those columns.

| Site | Before | After |
|---|---|---|
| `django_store._write` | inline construction | `message_of(entry)` |
| `django_store.append_many` | inline construction | `message_of(entry)` |
| `django_store._read_since` | inline construction | `message_of(entry)` |
| `channels_signals._frame` | read columns raw ✗ | `frame_event(entry)` → `message_of` |
| `channels_views` | own copy, same defects ✗ | shares `frame_event` |
| `views` dashboard API ×2 | raw `event_type` ✗ | `event_label(...)` |
| `admin` | raw columns | **unchanged, deliberately** |

**Why the admin is left alone.** It is an inspector for the stored rows themselves. Showing the actual column value is the correct thing there, not a divergence. Changing all six sites mechanically would have made it worse, so it did not happen.

**Why the dashboard only gets half.** `views.py` reads through `.values()` queries rather than model instances, so `message_of` does not apply without changing the query shape. The label inversion is a pure function of the column and transfers cleanly; the payload stays as stored. Noted rather than quietly skipped.

### The payload, which was not a simple share

The frame is JSON on the wire, so bytes that are not valid UTF-8 **cannot** ride in it. That is the actual root of #153 — the old frame published the stored column and left the subscriber no way to tell base64 from text that happens to look like base64.

Silently re-encoding would have repeated the bug in a new place, so the encoding travels with the value:

```python
event["data"] = <json value | text | base64>
event["payload_encoding"] = None (absent) | "utf-8" | "base64"
```

**The key is omitted entirely when the payload is an ordinary JSON value.** That keeps the common case byte-identical, so no existing subscriber sees a new key. This also matters because `test_channels.py:107` asserts an exact key set on the frame — it broke on my first attempt, and rather than loosen a real contract test I made the change genuinely additive. (`public-api.md` does list the SSE frame wire format as unpromised, so loosening would have been defensible; not needing to is better.)

### Tests

The two #153 spikes lost their `xfail` markers and pass. Two more added:

- `test_a_subscriber_can_recover_the_payload_a_reader_sees` — replaces the original payload spike. The original asserted the frame's bytes equal the reader's bytes, which is **impossible** over a JSON wire; it now asserts the subscriber can *reconstruct* them, which is the property that actually matters.
- `test_a_json_subscriber_still_receives_the_plain_json_value` — pins the no-regression case: ordinary JSON stays unencoded, with no new key.

### Verified

| | result |
|---|---|
| `pytest` (SQLite) | 1385 passed, 9 skipped, 8 xfailed |
| `pytest` (Postgres 16) | 1392 passed, 2 skipped, 8 xfailed |
| `ruff` / `ruff format` / `pyright src/` | clean |
| coverage | `event_message.py` 94%, `channels_views.py` 100%, total 95% |

xfail count drops 10 → 8: the two #153 proofs are now real tests.

PR #164, opened 2026-08-17
Read a rule's source once per rebuild, not once per event

### LINKED ISSUE #156: Rebuilding a table spends most of its time re-reading its own source code
Before applying each rule, the code checks whether that rule's source has changed since it was registered, by reading the source off disk and hashing it. This is worth doing once. It currently happens once per event.

Measured on this checkout, rebuilding from 2000 events spends around 86% of its total time re-hashing a handful of functions that cannot change while it runs. Doing it once at the start makes rebuilding roughly seven times faster and changes nothing a caller can observe, because the resulting warning is already reported only once. A failing test pins the shape rather than the timing.

joshbrooks:
Before applying each rule, the code checked whether that rule's source had changed since it was registered, by reading it off disk and hashing it. Worth doing once; it was happening once per event, on functions that cannot change while their own rebuild is running.

Rebuilding 2000 events now takes 18ms instead of 146ms, and reads the source once rather than 2000 times. Nothing a caller can observe changes, except that a rule which has drifted is reported once rather than once per event. The memo lasts for one rebuild only, so a long-running process that reloads code between rebuilds can still detect the change that reload introduced.

Closes #156. Stacked on #162. Detail in a comment.

### DIFF
--- a/src/rakaia/registry.py
+++ b/src/rakaia/registry.py
@@ -921,6 +921,7 @@ class UpcasterRegistry(_LogBackedRegistry):
         target_version: int,
         *,
         drift_callback: Callable[[UpcasterVersion, str], None] | None = None,
+        hasher: Callable[[Any], str] | None = None,
     ) -> dict[str, Any]:
         """
         Upcast `event` from its current schema_version up to `target_version`.
@@ -940,6 +941,12 @@ class UpcasterRegistry(_LogBackedRegistry):
         If `drift_callback` is provided, it is called once per step with
         `(version, current_hash)` if the upcaster's source hash has changed
         since registration. The callback decides whether to raise or warn.
+
+        `hasher` overrides how a step's live source hash is computed. A replay
+        passes one memoised for its own duration: an upcaster's source cannot
+        change mid-replay, and re-reading it per event per step was a measurable
+        share of replay time (#156). Defaults to hashing afresh, so a caller
+        that keeps no such ledger is unaffected.
         """
         current = int(event.get("schema_version", 1))
         if current > target_version:
@@ -969,7 +976,7 @@ class UpcasterRegistry(_LogBackedRegistry):
                 )
             step = matching[0]
             if drift_callback is not None:
-                live_hash = hash_function_source(step.fn)
+                live_hash = (hasher or hash_function_source)(step.fn)
                 if live_hash != step.source_hash:
                     drift_callback(step, live_hash)
             event = step.fn(event)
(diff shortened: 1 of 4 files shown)


joshbrooks:
### Measured

```
before: 2000 events → 146.0 ms | hash_function_source called 2000 times
after:  2000 events →  18.0 ms | hash_function_source called    1 time
```

~8× on this checkout. The 2000 calls were `inspect.getsource` + `textwrap.dedent` + SHA-256 on one unchanged function.

### The shape

`_ReplayCtx` gains two fields:

- `source_hashes: dict[Any, str]` — the memo, keyed by callable.
- `drift_reported: set` — so a drifted registration is reported once.

`live_source_hash(ctx, fn)` is the memoising hasher. `_check_handler_drift` and `_check_reducer_drift` now take `ctx` instead of `(on_drift, result)` — fewer arguments *and* access to the ledger.

**The comparison stays per registration.** Only the hash computation is memoised. Two registrations may legitimately share a function while storing different hashes, so collapsing the comparison would have been wrong; collapsing the I/O is not.

**Upcasters had the same problem** in `registry.apply_chain`, which hashes per step per event. It now takes an optional `hasher`, defaulting to `hash_function_source`, and replay passes `partial(live_source_hash, ctx)`. A caller that keeps no ledger is unaffected.

### Why per-replay and not a global cache

A module-level `lru_cache` would be smaller and faster still. It would also mean that a long-running process which reloads code between rebuilds could never detect the drift that reload introduced — the cache would answer with the pre-reload hash forever.

Caching away the signal inside the check whose entire job is to find it is a bad trade for a few microseconds, so the memo is scoped to one replay. That is also the semantically correct unit: the claim being relied on is "source cannot change *mid-replay*", which is true.

### Behaviour change, stated plainly

A drifted rule now produces **one** warning per replay instead of one per event. `result.drift_detected` already deduplicated by name; `result.warnings` and the log did not, so a drifted handler over a 10k-event stream emitted 10k identical warnings. `on_drift="raise"` is unaffected — it raised on the first event before and still does.

### Tests

- `test_a_handler_source_is_hashed_once_per_replay` — the spike from #162, marker removed, now passing.
- `test_handler_drift_is_reported_once_not_once_per_event` — **new**, in `TestDrift` beside the existing drift tests. Verified it catches a regression: with the dedupe removed it fails with 10 warnings for 10 events.

The five existing drift tests pass unchanged, which is the property that mattered most here — this touches a safety check, and detection had to stay exactly as sensitive.

### Verified

| | result |
|---|---|
| `pytest` | 1384 passed, 9 skipped, 9 xfailed |
| `just demos` | all pass (they exercise upcasters and drift end to end) |
| `ruff` / `ruff format` / `pyright src/` | clean |

xfail count 10 → 9.

**Note:** during this work I ran `git checkout -- src/rakaia/replay.py` to undo a temporary probe and reverted the entire fix with it. Reapplied and re-verified from scratch; the numbers above are from the reapplied version.

PR #167, opened 2026-08-17
Decide a batch write with the same rule as every other write

### LINKED ISSUE #154: A batch write skips the check that decides whether a write is allowed
There are four ways to write to a stream. Three of them ask the shared rule whether the write should be allowed. The fourth, the one that writes many events at once, does not — it checks only two of the conditions itself and ignores the rest.

The practical effect is that a writer that has been superseded by a newer one is correctly refused when it writes events one at a time, and silently accepted when it writes them as a batch. That is the protection against two writers corrupting a stream, and it does not cover the batch path. There is a failing test demonstrating exactly this.

joshbrooks:
There are four ways to write to a stream. Three of them ask the shared rule whether the write is allowed. The fourth, the one that writes many events at once, checked two of the conditions itself and skipped the rest — so a writer that had been superseded by a newer one was correctly refused when writing one at a time, and silently accepted when writing a batch.

It also never recorded the result, which had a second consequence: a producer that legitimately carried on after its own batch was rejected for a gap it had not created. Both are now fixed by asking the same rule per item, with an item the rule refuses keeping its slot in the results rather than disappearing.

Closes #154. Builds on #165, which is not yet merged. Detail in a comment.

### DIFF
--- a/src/django_rakaia/django_store.py
+++ b/src/django_rakaia/django_store.py
@@ -42,7 +42,7 @@ from typing import Any
 from asgiref.sync import sync_to_async
 from django.db import transaction
 
-from rakaia.append_decision import StreamFacts, decide_append
+from rakaia.append_decision import AppendVerdict, StreamFacts, decide_append
 from rakaia.context import merge_provenance
 from rakaia.json_mode import (
     format_json_response,
@@ -54,11 +54,9 @@ from rakaia.types import (
     AppendOptions,
     AppendResult,
     CloseResult,
-    ContentTypeMismatch,
     ProducerAccepted,
     ProducerState,
     ProducerValidationResult,
-    SequenceConflict,
     StreamConfigConflict,
     StreamMessage,
     StreamNotFound,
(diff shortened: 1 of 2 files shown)


joshbrooks:
### What was actually wrong

`decide_append` was called at three sites. `append_many` called it nowhere — its own loop checked per-item content type and `Stream-Seq` and stopped there. So a batch got:

- **no producer fencing** — a displaced epoch wrote freely;
- **no `closed_by` handling** — a producer retrying its own closing append was not recognised as a duplicate;
- **no producer state recorded** — even had it been fenced, the next writer would be judged against the pre-batch row.

### The fix

The validation loop now asks `decide_append` per item, threading facts that advance across the batch exactly as they would across a loop of `append`: `last_seq` moves with each accepted item, and a producer's state moves with each accepted fenced write, so a second item from the same producer is fenced against the first rather than against the pre-batch row.

Conflicts (`ContentTypeMismatch`, `SequenceConflict`) still raise and abandon the whole batch — they are caller errors, unchanged. Fencing outcomes are *returned*, so a refused item becomes a refusal in its own slot instead of aborting its siblings.

Two consequences that needed handling and did not exist before:

1. **All items refused** — nothing to write, and `get_next_offset_block(0)` raises `ValueError: count must be >= 1`. Guarded, with each item still getting its own answer.
2. **Some items refused** — the result assembly built results from the written items plus the trailing refusals, which would have silently re-paired every later result with the wrong input. Now built per input index.

### Tests, and the mutation check that made them worth having

Three mutations, each confirmed to be caught:

| Mutation | Caught by |
|---|---|
| verdict forced to `write=True` (restore the bypass) | `test_a_batch_from_a_fenced_producer_is_refused` |
| refused items dropped from results | `test_a_partly_fenced_batch_keeps_one_result_per_item` |
| `_commit_producer` removed | `test_a_batch_advances_producer_state_for_the_next_writer` |

The third one **did not pass first time**. My initial version asserted only that a replayed sequence was refused — which is true with *or* without the fix, because an uncommitted producer is refused for a sequence gap instead of recognised as a duplicate. Same outcome, wrong reason. It now asserts the recorded row, that the replay reports `duplicate` rather than a gap, and that the producer can continue at the next sequence — the last being the real user-facing consequence, since without the commit a legitimate continuation is rejected.

### A correction to the original spike

The #162 spike for this finding set up its fence with `producer_seq=1`. A producer's first sequence is **0**, so that setup append was itself refused for a sequence gap — the epoch was never established, and the test proved "the batch does not fence" rather than the intended "the batch admits a *stale-epoch* write". Same conclusion, weaker evidence. The setup now starts at 0 and asserts the precondition landed.

### Verified

`pytest` 1395 passed / 9 skipped / 7 xfailed · `ruff`, `ruff format`, `pyright src/` clean. Postgres and `just demos` running locally alongside this; CI covers both regardless.

xfail count 8 → 7. Net `django_store.py` change across this branch is −8 lines.

**Note on base:** this branch includes #165 (not yet merged) because both touch `django_store.py` heavily. If #165 lands first this rebases to nothing; the diff to review here is the `append_many` change.

joshbrooks:
Rebased onto `main` now that #165 and #166 have landed. The #165 commit this branch carried was dropped as already-upstream, so the diff here is now just the `append_many` change.

PR #168, opened 2026-08-17
Store a JSON message as itself, not with punctuation attached

### LINKED ISSUE #155: Streams that declare a content type cannot be replayed
When a stream declares that it holds JSON, the code adds the punctuation needed to send several events over the wire as one response, and then stores that punctuation as part of the event itself. The part of the system that sends responses knows to strip it again. The part that rebuilds tables does not, so it fails on the very first event.

This affects any stream with a declared content type, not just ones written in batches, and both storage backends behave the same way. The test that was supposed to cover this checks its result by sending it back out through the response path, which re-adds the punctuation — so it passes while the thing it describes is broken. Failing tests for both problems are in the repository.

joshbrooks:
When a stream declares that it holds JSON, its messages are joined into one array when read back. The code was adding the punctuation for that join at write time and storing it alongside the message, so what came out of storage was not valid on its own — anything reading a message directly, rebuilding tables most of all, failed on the very first one.

The punctuation now belongs to the response, which is the only place that needs it. Nothing changes on the wire: positions within a JSON stream are positions in the response body, so accounting for the separator that will follow each message leaves every position exactly where it was, and the conformance suite reports no change at all.

Closes #155. Detail in a comment.

### DIFF
--- a/src/django_rakaia/django_store.py
+++ b/src/django_rakaia/django_store.py
@@ -523,13 +523,14 @@ class DjangoStreamStore:
         if not is_json_content_type(stream.content_type):
             return [encode_payload(data, stream.content_type)]
 
-        processed = process_json_append(data, is_initial_create=is_initial_create)
-        if not processed:
+        # `process_json_append` validates and flattens, and now returns one
+        # payload per message rather than a framed blob — so its result is the
+        # split, and this no longer re-parses the body to redo the same work.
+        payloads = process_json_append(data, is_initial_create=is_initial_create)
+        if not payloads:
             # An empty array on create: a stream with no messages yet.
             return []
-        parsed = json.loads(data)
-        elements = parsed if isinstance(parsed, list) else [parsed]
-        return [(element, None) for element in elements]
+        return [(json.loads(payload), None) for payload in payloads]
 
     def _write(
         self,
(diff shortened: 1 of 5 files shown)


joshbrooks:
### The spec settles it

This was reported as a design divergence. It is a **protocol violation**. `docs/protocol.md` §7.1:

> For `application/json` streams, servers **MUST** preserve message boundaries.
> POST body `[{"event": "a"}, {"event": "b"}]` stores two messages: `{"event": "a"}`, `{"event": "b"}`

`process_json_append` returned the elements concatenated with trailing commas, and the in-memory store kept that blob as a single `StreamMessage.data`. Both halves of the MUST were broken.

### What changed

- `process_json_append(data) -> list[bytes]` — one unframed payload per message. It returned a framed blob.
- `format_json_response(payloads: Sequence[bytes]) -> bytes` — joins with `,` and wraps. It took a pre-concatenated blob and stripped a trailing comma.
- `StreamStore._append_to_stream` — appends one message per payload, returning the last (whose offset is the new head), which is the shape the durable store already documented.

### Offsets are unchanged, and that is the load-bearing claim

A JSON-mode offset is a position in the **response** body, where payloads are joined by commas. Advancing each message by `len(payload) + 1` therefore reproduces exactly the arithmetic the old `len(payload + b",")` produced. Every offset in every existing stream lands where it did.

**The conformance suite confirms it:** `276 passed, 56 failed, 6 skipped` — `0 NEW failure(s) · 0 newly passing · 56 expected (known gap)`. Byte-identical wire behaviour.

That `0 newly passing` is worth dwelling on: the conformance suite does not catch this violation either, because it asserts through the GET response, which round-trips the framing. Exactly the blind spot that let `server_store_contract.py`'s `test_a_json_array_append_flattens_into_separate_messages` pass while the property it names did not hold.

### A correction to the finding as originally reported

I said both stores shared the framing path, from grepping `process_json_append`. That was wrong. `DjangoStreamStore._payloads_for` called it **only for validation**, then re-parsed the body itself and returned one element per message — it was already correct. Its read side then had to *re-frame* the payloads (`b",".join(...) + b","`) purely to satisfy the shared formatter's expectations, only for the formatter to unframe them again. Both the re-parse and the re-frame are gone.

So the defect lived in the in-memory store alone. The durable store gets a simplification, not a fix.

### Tests

- Both #155 spikes pass, markers removed.
- `test_json_mode.py` rewritten where it pinned the old internal contract — the trailing comma and blob shape were asserted directly by five tests. Replaced with assertions on the list shape, plus a round-trip test that stored payloads are complete JSON values and framing is reversible.
- One of my own spike assertions was wrong: it expected `b'{"id": 1}'` where the store canonicalises to compact `b'{"id":1}'`. That compaction is pre-existing behaviour, so the test was corrected rather than the code.

### Verified

| | result |
|---|---|
| `pytest` (SQLite) | 1398 passed, 9 skipped, 5 xfailed |
| `pytest` (Postgres 16) | 1405 passed, 2 skipped, 5 xfailed |
| conformance | 0 new failures, 0 newly passing |
| `just demos` | all pass |

xfail count 7 → 5.

PR #169, opened 2026-08-17
Tell subscribers about a write only once it is permanent

### LINKED ISSUE #157: Subscribers are notified of writes that were then undone
Live updates are sent while the database change is still in progress, before it has been made permanent. If the surrounding work then fails and the change is undone, the notification has already gone out, and subscribers believe in an event that does not exist.

There is a second, related problem. Notification happens as a side effect of how a row got written, which meant one of the two write paths was invisible to it and had to be handled separately — that partition is currently held in place by a comment, and it has failed before. Making the store notify explicitly, once the change is permanent, fixes both. A failing test shows an undone write still reaching a subscriber.

joshbrooks:
Live updates were being sent while the database change was still in progress. If the surrounding work then failed and the change was undone, the notification had already gone out, and subscribers were left believing in an event that does not exist.

Sending now waits for the commit. The message is still built immediately, because building it reads rows that have to be read while the transaction is open, and outside a transaction the send still happens straight away — so an ordinary append is unchanged. Tests that assert what a subscriber receives are now asserting something that happens after commit, and have been adjusted to say so.

Closes #157. Detail, including why the second half of that issue is deliberately not done here, is in a comment.

### DIFF
--- a/src/django_rakaia/channels_signals.py
+++ b/src/django_rakaia/channels_signals.py
@@ -122,6 +122,14 @@ def broadcast_entries(stream_id: str, entries: Sequence[StreamEntry]) -> None:
     #
     # Outside a transaction Django runs the callback immediately, so an append in
     # autocommit still publishes synchronously.
+    #
+    # The alias is taken from the entries, the way `write_enveloped_event` takes
+    # it from the streams. An entry written to another database commits on *that*
+    # connection, so queuing against the default one would be queuing on a
+    # connection that is not in a transaction at all — Django then runs the
+    # callback immediately and publishes a write that has not happened yet, which
+    # is the very defect this defers to avoid (#157 on the path #159 opened up).
+    using = entries[0]._state.db
     frames = [_frame(stream_id, entry) for entry in entries]
 
     def _send() -> None:
@@ -129,7 +137,7 @@ def broadcast_entries(stream_id: str, entries: Sequence[StreamEntry]) -> None:
         for frame in frames:
             send(group_name, frame)
 
-    transaction.on_commit(_send)
+    transaction.on_commit(_send, using=using)
 
 
 @receiver(post_save, sender=StreamEntry)
--- a/src/django_rakaia/channels_signals.py
+++ b/src/django_rakaia/channels_signals.py
@@ -10,6 +10,7 @@ from typing import Any
 
 from asgiref.sync import async_to_sync
 from channels.layers import get_channel_layer
+from django.db import transaction
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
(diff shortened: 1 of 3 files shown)


joshbrooks:
### The change

`broadcast_entries` now frames eagerly and sends in a `transaction.on_commit` callback.

**Framing stays eager on purpose.** `_frame` reads `entry.event` and its related rows; deferring that would issue those queries after the transaction closed, against objects that may be stale. So the frames are built while the transaction is open and only the send is deferred — the callback closes over finished data.

Outside a transaction Django runs an `on_commit` callback immediately, so an append in autocommit publishes synchronously exactly as before.

One line in `channels_signals.py`; both publish paths get it, because both already funnel through `broadcast_entries`.

### Test re-timing

Predicted in the issue and it landed as expected: 12 tests failed on first run, all of them asserting a frame under a plain `django_db` marker whose transaction never commits.

- `test_architecture_spikes.py` — module marker moved to `django_db(transaction=True)`. These assert what a subscriber receives, which is now a post-commit fact, so running them transactionally is the honest semantics rather than a workaround.
- `test_decorators_consumer_contract.py::test_sse_broadcast_of_a_django_typed_payload_does_not_raise` — uses pytest-django's `django_capture_on_commit_callbacks(execute=True)`. It needs one frame, so running the pending callbacks explicitly is cheaper than paying for a truncating transactional run.
- `test_publish_on_append.py` and `test_channels.py` needed **no change** — both already used `transaction=True`.

Mutation check: replacing `transaction.on_commit(_send)` with a direct `_send()` fails `test_a_rolled_back_append_is_never_published`, so the deferral is what the test is actually pinning.

### The half I did not do, and why

The issue also asks for publication to become something the store does explicitly, with the `post_save` receiver removed — so the partition between the two triggers stops being held by a comment. I have left that alone, and think it is worth less than the card suggested:

1. **The partition is no longer only a comment.** Since #82 it is covered by `test_publish_on_append.py`, which drives the store and asserts what a subscriber receives. The failure mode that motivated the card is now tested.
2. **Removing the receiver has real consequences.** `test_raw_stream_entry_save_is_not_broadcast` pins that fixture rows (`raw=True` during `loaddata`) must not reach subscribers, and that the receiver's related-row dereferences would raise mid-`loaddata`. That protection lives in the receiver. Removing it means re-homing that rule *and* finding every ORM path that writes a `StreamEntry` outside the store — `@stream_model` goes through `create_stream_event`, not the store.
3. **Deferral fixes the defect on its own.** The reported bug is subscribers being told about undone writes. That is closed here.

Happy to do the unification as a follow-up if you disagree — but it is a larger change whose risk is concentrated in fixture loading, and it buys structure rather than correctness.

### Verified

| | result |
|---|---|
| `pytest` (SQLite) | 1397 passed, 9 skipped, 6 xfailed |

xfail count 7 → 6. Postgres and demos left to CI, which runs both.

PR #170, opened 2026-08-17
Keep a save and the event it emits on the same database

### LINKED ISSUE #159: Saving to a second database splits the write across two

joshbrooks:
A project can tell Django which database a save should go to. When it did, the row went where it was told and the event recorded alongside it went to the default one, because the instruction was discarded on the way through — one save, split across two databases.

Rather than pass the instruction hand to hand through every call, it now travels with the data: the event is written to whichever database its streams came from, and the offset high-water is read from whichever database the stream row came from. Nothing in the chain has to remember to forward anything, and the delete path is fixed by the same change as the save path.

Closes #159. Detail in a comment.

### DIFF
--- a/src/django_rakaia/decorators.py
+++ b/src/django_rakaia/decorators.py
@@ -29,9 +29,9 @@ DataclassTransformer = Callable[[models.Model], Any]
 DeleteEventType = Literal["delete", "update"] | None
 
 
-def _get_or_create_stream(stream_id: str) -> Stream:
-    """Get or create a Stream record."""
-    stream, _ = Stream.objects.get_or_create(stream_id=stream_id)
+def _get_or_create_stream(stream_id: str, using: str | None = None) -> Stream:
+    """Get or create a Stream record on `using` (the default database if None)."""
+    stream, _ = Stream.objects.using(using).get_or_create(stream_id=stream_id)
     return stream
 
 
@@ -40,6 +40,7 @@ def create_stream_event(
     to_dataclass: DataclassTransformer,
     instance: models.Model,
     action: str,
+    using: str | None = None,
 ) -> StreamEvent:
     """
     Create a stream event for the given model instance.
@@ -53,6 +54,10 @@ def create_stream_event(
         to_dataclass: Callable converting model instance to dataclass.
         instance: The model instance that changed.
         action: Event type ("create", "update", or "delete").
+        using: Database alias to write to. Defaults to the instance's own
+            (`instance._state.db`), so an event lands on the same database as
+            the save that produced it — the two are one write and must not be
+            split (#159). Pass explicitly to override.
 
     Returns:
         The created StreamEvent instance.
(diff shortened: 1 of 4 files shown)


joshbrooks:
### The alias follows the data

The issue proposed threading a `using` parameter down through the write path. Doing that made three of the four hops untestable-in-isolation, because most of them already know the answer. So the shape here is different, and simpler:

| Hop | How it gets the alias |
|---|---|
| `Stream.get_next_offset_block` | `self._state.db` — the stream row it locks |
| `write_enveloped_event` | `streams[0]._state.db` — the streams it writes into |
| `create_stream_event` | `using=` if given, else `instance._state.db` |
| `handle_post_save` / `handle_post_delete` | forwards Django's `kwargs["using"]` |

`self.entries` already followed the instance's alias, so the watermark not doing so was the inconsistency — allocation and tail-reporting are documented as agreeing, and on a non-default alias they did not.

### Mutation results, including the two that taught me something

| Mutation | Caught? |
|---|---|
| `write_enveloped_event` ignores the stream alias | **yes** |
| watermark ignores the instance alias | **yes** (after strengthening the test — see below) |
| receiver drops `kwargs["using"]` again | **no**, and correctly so |

**The receiver mutation is uncatchable, and that is the right answer.** `instance._state.db` is already `"overlay"` after `save(using="overlay")`, so the fallback covers it. The explicit forward is belt-and-braces: it uses the signal Django actually documents rather than inferring. I kept it, but no test can distinguish it, and pretending otherwise would be worse.

**The watermark mutation was not caught at first.** The test asserted only where the `StreamEvent` landed, so allocating offsets from the *default* high-water while writing the entry to `overlay` passed. That is precisely the leak the change exists to close — a rebuild against a scratch database reading its offsets from the live one — so the test now asserts the entry and the high-water land on `overlay`, and that no high-water for that path appears on `default`.

### Fallout worth naming

`create_stream_event` accepts anything with the attributes `to_dataclass` reads, and `test_envelope_writer.py` passes a plain stand-in with no ORM state. `instance._state.db` broke 8 of its tests. The lookup is now `getattr(state, "db", None)` — no ORM state means no routing, which is the default database.

### A defect in my own tooling, found while doing this

The script I have been using to strip `xfail` markers as each finding is fixed also stripped the decorator *below* it. Here it removed `@pytest.mark.django_db(databases=["default", "overlay"])`, which surfaced immediately as `DatabaseOperationForbidden`. Restored.

I audited the other open PRs for the same damage. **#169 is affected**: it lost `@pytest.mark.django_db(transaction=True)` from `test_a_rolled_back_append_is_never_published`. It is harmless there — that branch sets the same marker at module level, and the mutation check on that PR confirmed the test still fails when its fix is reverted — but it was accidental, and is recorded here rather than left unmentioned. #168 was unaffected.

### Verified

`pytest` 1397 passed / 9 skipped / 6 xfailed · `ruff`, `ruff format`, `pyright src/` clean. xfail count 7 → 6.

The larger half of the issue — a `DjangoStreamStore` constructible on another alias — is untouched and still open as **Worth exploring**. This closes the write-path half, which is what the failing test pinned.

Issue #172, opened 2026-08-19
Recording that a row was added, removed or moved, rather than working it out afterwards


joshbrooks:
Repeated collections inside a form are stored as a plain list, so the only record
of a structural change is the before and after picture. Reordering rows looks
identical to editing every row in place, which is a gap ADR 0001 documented and
left open. The proposal is three events — a row was added at this position, a row
was removed, a row was moved — each naming the row by a stable id rather than its
position.

The spike builds these and they behave, including minting the stable id the
source data does not carry. But the case they exist for happens eight times in
7,161 recorded collection changes: 97% are edits or additions at the end, which
the existing helpers already handle correctly. They also cannot help until
something upstream actually knows a row was moved — a whole-form save has to
guess, and guessing from two pictures is exactly the problem. Detail in the
comments.

joshbrooks:
## Spike

Branch `spike/delta-events`, same module and same payload shape as #171 — the two
proposals are one mechanism, so a patch event can mix a field edit and a row move
in one atomic event.

```python
from rakaia.deltas import AddPart, RemovePart, MovePart, PART_ID_KEY, parts_of

AddPart("repeaterBalance", 2, "b7f3", {"balance_cash": 400})  # key, index, id, row
RemovePart("repeaterBalance", "b7f3")
MovePart("repeaterBalance", "b7f3", 0)
```

The design decisions worth reviewing:

- **`AddPart` mints the identity.** ADR 0001 requires a stable child id and notes
  that when the source has none it must be *assigned at ingestion*. This is that
  assignment: the producer's `part_id` is stamped into the folded row under
  `_part`, so `reconcile_tree(id_fn=lambda r: r[PART_ID_KEY])` works directly on
  folded output — identity from the log, position from the fold, which is the
  split the ADR asks for. `parts_of(state, key)` returns `(part_id, position,
  row)` triples for exactly that.
- **`index` is a command parameter, never an identity** — a position in the
  collection at the moment of the edit. Remove and move name the row by id, so
  they stay correct when folded after a reorder they did not know about.
- **`MovePart.index` is the destination with the moved row taken out** — what a
  drag-and-drop UI reports, and the reading that makes moving a row to the place
  it already occupies a no-op rather than an off-by-one. Pinned by test.
- **`_part` is reserved**: `AddPart` rejects a value that already carries it, or
  two rows could claim one identity. A duplicate id, an unknown id, and an index
  past the end are all `DeltaConflictError`, never a silent skip.
[...]

Issue #171, opened 2026-08-19
An event that says what changed, instead of re-sending the whole form


joshbrooks:
Today every event carries a complete copy of its subject. When one field of a
form changes, the whole form is appended again — including any repeated child
collections, which in Partisipa's forms are most of the payload. The proposal is
a second kind of event that carries only the change: set this value, remove that
one.

A spike exists and works, but the measurements do not support the reason for
building it. Two thirds of saves in production are already byte-for-byte
identical and can be dropped with a feature rakaia already ships; on the
remaining third the smaller events would save about 8 MB across the entire
history, and 9% of the time they would be *larger* than the full copy. The
numbers and the six places this shape breaks existing behaviour are in the
comments.

joshbrooks:
## Spike

Branch `spike/delta-events` — `src/rakaia/deltas.py` plus 81 tests. Full gate
green (`ruff check`, `ruff format --check`, `pyright src/` 0 errors, `pytest`
1461 passed). `rakaia.deltas` is deliberately **absent from `__all__`**: the
public API promises nothing here.

Two operation types cover the partial update; the array ops are #172, and they
share one payload shape so the two proposals are one mechanism.

```python
from rakaia.deltas import SetField, ClearField, encode_patch, fold_snapshot

store.append(path, json.dumps(encode_patch([
    SetField(("fields", "total_amount"), 6000),
    ClearField(("fields", "note")),
])).encode(), AppendOptions(label="patch"))

state = fold_snapshot(store.read(path).messages)  # snapshots + patches -> state
```

- Envelope `label="patch"`, payload `{"ops": [...]}`. One save is one event
  however many things it touched, so atomicity holds and the audit log still
  reads one row per save.
- Paths are JSON Pointers (RFC 6901) on the wire, so the escaping rule for a
  segment containing `/` or `~` is a citation rather than a house invention.
- `ClearField(p)` is distinct from `SetField(p, None)` on purpose: "answered
  none" and "not present" are different facts.
- A set through a missing branch raises rather than autovivifying, so a patch
  written against one schema cannot silently invent a branch in another.
- `fold_snapshot` refuses rather than guesses: `NoBaseSnapshotError` when a patch
  has no preceding snapshot, `DeltaConflictError` (naming the offset) when an op
  does not fit the state. Both mean "the fold is not standing where the producer
  stood".

## Measurements — live Partisipa dataset

16,588 submissions with a `fields` blob; 70,910 history events over 16,791
[...]

PR #173, opened 2026-08-19
Spike: events that say what changed, instead of re-sending the whole form

joshbrooks:
Spikes #171 and #172. **Not for merge** — this is the measurement, not a proposal
to ship.

Adds a partial-update event and three array operations (add, remove, move a row),
with a pure fold that rebuilds current state from a snapshot plus the patches
after it. Rows get a stable id minted when they are added, which is what makes a
reorder recordable and closes a gap ADR 0001 documented and left open.

The design works and the gate is green, but the numbers it was built to justify do
not hold: two thirds of saves in production are already identical and need only a
feature rakaia ships today, the smaller events would save 8 MB across the whole
history, and the reordering case happens eight times in 7,161 collection changes.
The recommendation is to hold, and both issues carry the trigger that would change
the answer.

### DIFF
--- a/docs/research/delta-events.md
+++ b/docs/research/delta-events.md
@@ -1,0 +1,203 @@
+---
+icon: lucide/microscope
+---
+
+# Delta events: partial updates and array operations (spike)
+
+- **Date:** 2026-08-20
+- **Status:** research note — not a decision. Decisions go in `docs/adr/`.
+- **Spike branch:** `spike/delta-events` — `src/rakaia/deltas.py`,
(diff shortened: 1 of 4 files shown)


joshbrooks:
Closing with #171 and #172, shelved as `wontfix`.

The branch `spike/delta-events` is **not** deleted — it is the record of what the
shape would cost, and `docs/research/delta-events.md` on it carries the method
and the production numbers. Both issues name the triggers that would reopen the
question.

Issue #62, opened 2026-07-26
Deferred / minor future tasks (icebox)


joshbrooks:
Low-priority items intentionally not done now. Each lists its **un-defer trigger** — don't action an item unless its trigger fires. This is an icebox to keep the board readable, not a backlog commitment.

### `DjangoStreamStore` durable append query count
`Stream.get_next_offset` does `get_or_create` + a separate `select_for_update().get()` (a redundant SELECT), and runs a per-append `Max(offset)` aggregate over `entries` that is dead weight once the watermark is seeded. Roughly 2 → 4 queries per append.

- **Un-defer trigger:** the durable append path becomes hot (e.g. Partisipa bulk import).
- **Why safe to defer:** invisible at low write frequency.
- Possible shape: fold into a single locked upsert; gate the aggregate on `watermark.high == 0` (or seed via data migration and drop it). Touches the concurrency-critical allocation path, so keep it isolated from correctness changes.

### `StreamOffsetWatermark` unbounded growth
The watermark table keeps one permanent row per distinct stream path and is never reaped — intentionally, since it must survive `Stream` deletion to keep offsets globally monotonic across delete+recreate (#34 Defect #2).

- **Un-defer trigger:** stream paths become high-cardinality / ephemeral (e.g. `session:<uuid>:…`).
- **Why safe to defer:** harmless while paths are a bounded set (per-user / per-area / a few global feeds). Note the growth driver is path **cardinality**, not write rate — low write frequency does *not* bound it.
- Possible shape: retain-with-intent (document + leave), TTL/GC of watermarks with no live stream and an age threshold, or fold the high-water into a tombstone. Needs a retention-semantics decision before any code.

_Context: surfaced during review of #59 (follow-up to #34 / #35)._

joshbrooks:
**First item done** — the durable append query count, in #175. Its un-defer
trigger had fired: #11 landed, and the polyglot stress harness (#174) measured
every save at ten queries, four of them spent deciding which offset to use.

Allocation is now two queries. The `Max(offset)` aggregate over the entry table
runs once per stream path rather than on every append — a watermark still at zero
is the one case it is needed for, because that means either a new path or an
install upgraded across migration `0005`, which creates the table without
backfilling it. The redundant second read of the watermark row is gone too: one
locked `get_or_create` does what a plain one plus a `select_for_update().get()`
was doing. `Stream.current_offset` had the same scan on the read side and lost it
in the same change.

[...]

Issue #34, opened 2026-07-21
Store offsets should be globally monotonic + lexicographically sortable (protocol conformance; subsumes subscriber-cursor recreation gap)


joshbrooks:
Root-cause follow-up to the subscriber-cursor gap found while reviewing #33. The fix belongs in the **store/offset layer**, not the consumer.

## Two related defects

**1. The Django store emits non-lexicographic offsets — a protocol violation.**
`docs/protocol.md` §3/§5.2: offsets are opaque but **MUST** compare byte-wise lexicographically. `DjangoStreamStore` returns `str(entry.offset)` (`"1"`, `"2"`, … `"10"`), so `"10" < "2"` lexicographically — non-conformant. The in-memory store's zero-padded `{seq}_{bytes}` is fine. `rakaia.subscription` had to work around this with a numeric parse (`_offset_key`); a conformant store makes that unnecessary and lets cursor comparison be plain lexicographic per spec.

**2. Offsets are reused across delete+recreate, enabling silent data loss for consumers.**
On recreate, both stores reset offsets to the low end. A subscriber whose committed offset collides with the recreated stream's head reads as `caught_up` and silently skips the new content (the #33 adversarial finding; pinned by a strict-xfail there). Note the protocol already says *"if a stream is deleted a new stream SHOULD NOT be created at the same URL"* (§3) — so this only bites a protocol-violating caller — but the store can make it structurally impossible.

## Proposed fix — globally monotonic offsets

Adopt the EventStoreDB `$all` / Kinesis sequence-number model: an offset, once issued, is never reused for a given path.

- **Django store:** emit zero-padded, lexicographically-sortable offset strings (conformance), and never reset the high component on recreate — carry a per-path monotonic generation so a recreated stream issues offsets strictly greater than any prior. (`Stream` pk or a persisted counter is the natural generation source.)
- **In-memory store:** bump the offset's existing `read_seq` (high) component on recreate instead of resetting to `0`, so `{seq}_{bytes}` stays globally monotonic.

**Result:** a recreated stream's head always sorts *after* any prior cursor, so the consumer reads it as a normal `advanced` and the new content is delivered — no generation token needed in the consumer API, and the #33 strict-xfail flips to xpass.

## Alternative considered
A consumer-side **generation token** (Kafka topic-ID style) threaded through the cursor. Rejected as the primary fix: it widens the consumer API and only patches over a store that issues non-conformant, reusable offsets. The store fix is the root cause and subsumes it.

[...]

PR #175, opened 2026-08-20
Stop re-deriving an offset high mark the watermark already holds

joshbrooks:
Closes the first item in #62, whose un-defer trigger has fired: #11 landed, and
the polyglot stress harness in #174 measures every durable save at ten queries,
four of them spent working out which offset to use next.

Two of those four were avoidable. The high mark was recomputed from a scan of the
whole entry table on every append, even though the watermark row already holds
it and nothing can write an entry above it; and the watermark row itself was read
twice, once to make sure it existed and again to read it under a lock. Allocation
is now two queries. `Stream.current_offset` had the same scan on the read side and
is fixed alongside it, because fixing one half of this invariant and leaving the
other is exactly how #34/#59 went wrong before.

The aggregate could not simply be deleted: migration 0005 creates the watermark
table without backfilling it, so an upgraded install has entries beside a
watermark still at zero. It now runs once per stream, ever, and writes what it
finds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,29 @@ runnable demo for each.
 
 ## [Unreleased]
 
-Nothing yet.
+### Changed
+
+- **A durable append no longer scans the entry table to find its place.** Every
+  append recomputed the stream's high offset with an aggregate over all of its
+  entries, and read the offset high-water row twice — once to create it, once to
+  lock it. Allocation is now two queries: one locked read, one write. Reading the
+  stream head (`Stream.current_offset`, `DjangoStreamStore.current_offset`) drops
+  the same scan and is one query.
+
+  The high-water row is authoritative once it has been advanced, so the scan
+  survives for the single case that needs it: a watermark still at zero, which
+  means either a brand-new stream or an install upgraded across migration `0005`
+  (it creates the table without backfilling it). That runs once per stream and
+  seeds the row for every allocation after it. No migration is required and no
+  offset changes.
+
+  **Worth knowing if you write `StreamEntry` rows directly.** An entry inserted
+  without going through `Stream.get_next_offset_block` is now invisible to both
+  allocation and head-reporting once the watermark is above zero, where the old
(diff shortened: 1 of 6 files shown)


joshbrooks:
Reviewed with three rounds of `/review-loop`, a fresh sub-agent each time, each
round briefed to verify the previous round's fixes rather than trust them.

**Round 1** — one green mutation, which is the finding the loop exists to catch.
The read-side gate `if watermark_high:` has to be a truthiness test: no watermark
row reads as `None`, a row that exists but was never advanced reads as `0`.
Rewriting it as `is not None` returned a head of 0 beside entries running to 7
and left all 542 Django tests green, because the test for that case created no
row at all and so only ever exercised the `None` path. Plus a docstring still
naming the `max(entries, watermark)` expression this change deletes, and a
missing CHANGELOG entry. Fixed in `2e60a34`.

**Round 2** — two defects in the CHANGELOG prose. `DjangoStreamStore` has no
`current_offset` (it is `get_current_offset`, and two queries not one), and
"invisible" was the wrong word for a hand-written entry: the row is still stored
and still returned by a read, it just loses its place in the order. Fixed in
`e903f31`.

**Round 3** — clean. 8/8 mutations killed.

The claim I was least willing to take on trust was that
`select_for_update().get_or_create()` really carries `FOR UPDATE` into both the
`get` and the `IntegrityError` retry, because SQLite makes every assertion about
it pass vacuously. Two rounds attacked it independently on Postgres: four
concurrent first-writers allocating `[1, 3, 5, 7]`, then 8 connections × 10 block
allocations racing from a barrier — 80 blocks, zero duplicate starts, zero
overlaps — and the lock test goes red when the lock is removed. Round 3 also
confirmed the unique constraint is real (`unique_together = ["stream", "offset"]`,
`models.py:363`) and that a collision is what happens, not a silent overwrite.

Gates: SQLite 1397 passed, Postgres 1398 passed, ruff clean, pyright 0 errors.

PR #176, opened 2026-08-20
Pin the producer-fencing table where it is decided

joshbrooks:
Closes item 5 of #161, and item 1 alongside it.

`rakaia.producer` holds the whole producer-fencing rule, and both stores call it
so they cannot decide differently — but nothing tested it directly. The callers
do cover it, better than "untested" suggests: stubbing the stale-epoch branch
alone fails seven tests across five files. What they cover is the *outcomes*,
each reached through a live append. What they do not reach is the edges and the
payloads — measured, seven of nine mutations to the module leave the entire suite
outside this file green, and the test docstring lists which.

43 cases now cover every branch and state the boundaries rather than assuming
them. Two rows earned their own treatment: expiry is consulted before the epoch,
so fencing lasts only as long as the winner keeps its state warm; and the closed
outcome in the result union is never returned here because closure is decided
upstream. Also documents the `rakaia.replay` module/function shadowing (item 1),
which silently defeats a monkeypatch and cost a wrong measurement during #156,
and pins `PRODUCER_STATE_TTL_SECONDS`, which was asserted nowhere.

No behaviour change — tests and two docstrings.

### DIFF
--- a/src/rakaia/replay.py
+++ b/src/rakaia/replay.py
@@ -2,6 +2,24 @@
 Replay orchestrator: re-derive materialized state from a stream by running
 versioned handlers and an effect executor.
 
+.. warning::
+
+   **`rakaia.replay` the attribute is the function, not this module.** The
+   package root does ``from .replay import … replay``, which rebinds
+   ``rakaia.replay`` to the function — so ``import rakaia.replay as rp`` then
+   ``rp.build_pipeline`` raises ``AttributeError: 'function' object has no
+   attribute 'build_pipeline'``.
+
+   ``from rakaia.replay import X`` and ``sys.modules["rakaia.replay"]`` both work
+   normally; only attribute access through the package root is shadowed. The
+   trap is that a ``monkeypatch.setattr("rakaia.replay.<name>", …)`` lands on the
+   function object and silently patches nothing, so a test measuring call counts
+   reports zero and reads as a pass. That cost a wrong measurement once (#161).
+
+   Renaming either one would be a breaking change to `rakaia.__all__`, so this
+   is a documented sharp edge rather than a bug to fix. Import the names you
+   need directly.
+
 Per event the pipeline is:
     1. Load the raw event bytes from the stream
     2. Decode as JSON, upcast to current schema via UpcasterRegistry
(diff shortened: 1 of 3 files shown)


joshbrooks:
Reviewed with four rounds of `/review-loop`, a fresh sub-agent each round, each
briefed to verify the previous round's fixes rather than trust them.

The useful number: **three of the four rounds found a defect in the previous
round's fix**, and all of them were in one test.

`TestTheFunctionIsPure::test_no_outcome_mutates_the_state` was written four
times, and the first three each had a distinct way of not seeing a mutation it
was named for:

1. **v1** passed while the code mutated its input. The fixture's `last_updated`
   equalled the `now` being passed, so `state.last_updated = now` was a no-op the
   assertion could not detect.
2. **v2** claimed "every row of the table" and reached three branches — every
   case presented `epoch=2` against a state at epoch 2, so the stale-epoch and
   higher-epoch arms were never entered. Two live mutants against the full suite.
3. **v3** reached the epoch branches but always handed in a *fresh* state, so the
   unknown/expired arm was never entered and a mutation at the top of it survived
   all 1426 tests.
4. **v4** selects the arm explicitly. Round 4 enumerated all nine `return`
   statements in `producer.py` and mutated both mutable fields of the passed-in
   state at each — 18 mutations, 18 killed by that test alone, run in isolation
   so nothing is caught incidentally by a neighbour.

Each earlier fix was correct about the defect it was fixing and blind to the next
one, which is the argument for the loop rather than a single careful pass.

Two other things the rounds changed, both worth recording because they made the
file **smaller**:

- The nine-case closed-outcome parametrize and the `"monkeypatch" in docstring`
  assertion are both gone. They were substring checks on module source standing
  in for prose: satisfied by a passing mention in a comment, and unable to fail
  for the reason they claimed. The boundaries they described are now paragraphs
  where someone editing the code will read them, and the behaviour is asserted
  behaviourally elsewhere.
- The file's own justification was wrong twice. It first claimed two callers
  exercising one row each (actually seven tests across five files), then named
  three boundaries as unique to it (two were caught elsewhere). It now states a
  measured figure — seven of nine mutations survive outside this file — which was
  re-verified against the final file, not the one it was measured on.

Final: 45 cases, 26/26 mutants killed including `PRODUCER_STATE_TTL_SECONDS`
which was previously asserted nowhere in the repo. Suite 1427 passed, ruff clean,
pyright 0 errors.

PR #177, opened 2026-08-20
Put the value-equality rule where both sides can reach it

### LINKED ISSUE #160: Two parts of the system can disagree about whether a value changed

joshbrooks:
Closes #160.

Two paths ask the same question — does this stored value differ from the one the
log carries? The verify path asks it to check whether replaying reproduces the
rows already there; the write path asks it to decide whether a write is a no-op
it can skip. When they disagree the failure is quiet: a diff certifying a clean
rebuild while the executor rewrites every row over a coercion, or the reverse.

The rule lived in the verify module, so the write path imported it from there and
only the verify path could be handed a different normalizer set. It now lives in
its own module that both import and neither owns, and `DjangoExecutor` takes
`normalizers=` — so a consumer with a domain-specific rule can hand the same set
to both instead of having a diff that honours it and a skip path that does not.

Beyond the layering: reaching `canonical_value` no longer imports the projection
reader or the effect types, and the old import path still works — the exported
names are Tier 1, the module defining them is not.


joshbrooks:
Reviewed with four rounds of `/review-loop`, a fresh sub-agent each round, each
briefed to verify the previous round's fixes rather than trust them.

**The recurring defect in this PR was prose contradicting code — three times, in
three different files.** Worth recording because it is a property of *refactors*
specifically: moving code moves the sentences describing it, and a sentence that
was true where it came from is not automatically true where it lands.

- **Round 1** — I kept a dead `attname` fallback in `_resolve_field` and gave a
  reason that was wrong. I claimed it had no trigger on Django 6.0 but should stay
  because the floor is 4.2; the reviewer read 4.2's `options.py` and found
  `_forward_fields_map` indexes by `attname` there too, with a comment in Django's
  own source saying `get_field()` should fetch by attname. Verified against the
  4.2 wheel, then deleted the loop. Round 2 went further and probed MTI, GFK, M2M
  and O2O across 26 models — no field kind where `get_field(attname)` raises.
- **Round 1** — I narrowed a public surface while silencing a linter. `__all__`
  was added to `verification.py` to quiet three F401s, so it was written from the
  re-exports and silently dropped `VacuousVerification`, `GREEN`, `RED`,
  `VACUOUS`. Attribute access still worked, so nothing failed.
- **Round 2** — the three tests I then wrote to pin `__all__` could not see the
  six re-exports, which is the half the list exists for. Every way of *deriving*
  the list misses them: not defined here, and now mapped elsewhere by the package
  root. Dropping `canonical_value` left the whole suite green while genuinely
  breaking `from django_rakaia.verification import *`.
- **Round 3** — two sites in the new module still asserted that a custom
  `normalizers=` set "is not mirrored on the skip path". That is the exact
  limitation this PR removes. I had corrected the executor's copy when adding the
  parameter and carried these across verbatim. `canonical_value`'s docstring is
  what `docs/api-reference.md` renders, so the generated reference was describing
  the defect as current.
- **Round 4** — clean, after one last sentence in `verification.py` still saying
  the normalization "belongs here".

Two things the loop added that I would not have: the seventh-re-export guard (a
hand-written list tied back to reality from the other end, so adding a re-export
without a test fails), and the observation that its exclusion was name-based where
the criterion is structural — now classified on `__module__`, verified to catch a
new re-export and tolerate a new local constant.

Also caught: no CHANGELOG entry for a public parameter and a Tier 1 name, and a
subprocess test that resolved settings through the implicit path entry, so running
pytest from another directory failed as though the layering had regressed.

Final: 1418 passed, ruff clean, pyright 0 errors, `just api-reference-check` clean.

PR #197, opened 2026-08-20
Restore DjangoStreamReader, and document what it drops

joshbrooks:
Reverts #196, which should not have deleted this class.

The case for deleting rested on it having no production callers. That was true of
this repository and says nothing about consumers: a grep cannot see their code,
and the class was importable. Being absent from `__all__` makes it free to
*change* under the tier policy, not safe to assume unused — I treated one as the
other.

Restored as it was, plus what the deletion was actually about. This reader returns
the payload and nothing else, while the rows it reads carry a label, metadata, an
event timestamp and a position — and both readers satisfy the interface `replay()`
needs, which is why the loss is invisible at the call site. That is now written on
the class, with a pointer to the store for those tables and the two differences
that matter if you switch, and pinned by four tests instead of left as prose.

Also fixes a defect in #196's own tests that its review found: the case named for
the decorator path built its rows by raw ORM and never called
`create_stream_event`. It does now.

### DIFF
--- a/src/django_rakaia/streams.py
+++ b/src/django_rakaia/streams.py
@@ -1,28 +1,28 @@
 """
-Reading an existing Django table as a *virtual* stream.
+Adapters that expose Django data to rakaia's replay pipeline.
 
-`ModelStreamReader` lets `replay()` treat rows you already have as if they were
-events: it reads them in a defined order from a queryset and encodes each one
-with a `to_payload` callable you supply. Use it when a durable, ordered
-source-of-truth already exists and you do not want to mirror it into a stream.
+Two readers, picked according to where your event log lives:
 
-**It carries no envelope, and that is correct.** The rows it reads were never
-events, so there is no recorded label, actor or event time to carry — it
-synthesises a payload rather than reproducing one. Anything that reads the
-envelope (`envelope_actor`, `merge_replay(order_key=...)`,
-`history_effects(version_of=...)`) has nothing to read here by construction.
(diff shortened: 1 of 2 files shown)


joshbrooks:
Closing unmerged. The revert was over-cautious and I was wrong to make it.

Verified against the only consumer — Partisipa, `feat/durable-store` in the
`bigbang` worktree. `DjangoStreamReader` is not imported, called or referenced
anywhere: zero hits across `.py`/`.md`/`.toml`, zero for `django_rakaia.streams`,
no dynamic use (`RAKAIA_STORE` only ever selects `'durable'`, and every
`importlib` call is unrelated). They already import `DjangoStreamStore` —
the correct reader for those tables. `git log --all -S` turns up exactly one
historical hit and it is a docstring line, on a commit that is not an ancestor of
that branch. They also pin `rakaia-streams==0.2.*`, an exact-minor pin their own
comment calls deliberate, so a removal in 0.3 could not reach them silently.

My objection in this PR was that a grep cannot see a consumer's code. That is
sound about an unknown; it does not survive the unknown being resolved. With one
consumer, verified not to use it, behind an exact-minor pin, #196 was right.

Two things here are worth keeping regardless of the class existing, and go up
separately: the tests seeded through the real `create_stream_event` (#196's
version was named for that path and built its rows by raw ORM — its review caught
that), and the note that the store differs from the deleted reader in raising on
a missing stream and in touching the TTL clock, which matters now the docs point
people at it.

PR #200, opened 2026-08-20
One JSON shape for an event, and one timestamp for it

### LINKED ISSUE #185: The same event is turned into JSON in four places, and they have drifted
Four surfaces show an event to somebody — the live push, two read endpoints, and the admin. The small pieces they need were made single a while ago, but the assembly was not, so each surface still builds its own version and the versions no longer agree. Two of them cannot even call the same function, because one takes a database row and the others take a dictionary.

The admin calls neither and is wrong twice over, which is filed separately as a bug. One function for "an event, as JSON" would make the next surface correct by construction.

joshbrooks:
Closes #185.

Three surfaces turned an event into JSON and each assembled its own dict — the
live channel frame, the dashboard listing, and the per-stream API. #153 made the
small pieces they share single but left the assembly per-surface, so they drifted
again, this time on which timestamp they publish: the frame used the event's
`created_at`, both HTTP surfaces the entry's.

The entry's is right. `message_of` uses it for the message timestamp, so it is
what `read()` reports, and a message is per-stream — one fanned-out event has one
transport time per stream that received it, which a single event-level value
cannot express. So the frame was the outlier and unifying the assembly fixes it.
Both HTTP payloads come out byte-identical.

Tested at the surfaces rather than at the shared helper, because that is what
would have caught this: reverting the frame's timestamp left all 637 Django tests
green beforehand. There are now cases asserting what a frame, an API response and
the dashboard's template context each report, and that the three agree with
`read()`.

### DIFF
--- a/src/django_rakaia/channels_signals.py
+++ b/src/django_rakaia/channels_signals.py
@@ -14,7 +14,7 @@ from django.db import transaction
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
-from .event_message import event_label, payload_fields
+from .event_message import event_view_of_entry
 from .models import StreamEntry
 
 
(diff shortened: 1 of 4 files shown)


joshbrooks:
Reviewed with two rounds of `/review-loop`, fresh sub-agent each round.

**Round 1 verified the headline claim better than I would have.** It dumped
`response.content` on `main` and on this branch across five payload flavours —
ordinary JSON, `text/plain`, base64, labelled, and a multi-event stream — with
`StreamEvent.created_at` and `StreamEntry.created_at` frozen to *different*
constants so a swap would be visible, then diffed the bytes. Both HTTP surfaces
identical, key order included; the only difference in the dump is the intended one.

**The finding that mattered:** swapping `isoformat()` for `str()` changed the
published timestamp on all three surfaces and left the suite green. Every case I
had written either compared one surface against another — so a format change moves
both and cancels — or parsed with `datetime.fromisoformat`, which accepts a space
as readily as a `T`. Fixed, along with key order: byte-identity is this branch's
whole claim, a reordering of `event_view` had survived a mutation pass, and an
unpinned claim decays on the next edit.

Round 2 confirmed all four survivors now die in the test named for each, with no
incidental failures, re-verified byte-identity after the follow-up edits, and
checked the new annotations are genuinely enforced rather than silently `Any` —
`created_at="2020-01-01"` now raises `reportArgumentType`.

Two things worth recording for whoever reads this later:

- **The SSE surface shifted too**, not just the channel layer.
  `channels_views.py:69` routes SSE replay through `frame_event`, so a browser
  client on `stream_detail.html` sees the corrected timestamp. It is display-only
  there (`toLocaleString`), and nothing in `docs/`, `static/` or the conformance
  files reads the field.
- **Round 1 found a live defect outside this PR's scope.** `admin.py` has *three*
  badge methods rendering the label and #195 fixed one; `:165` and `:217` — the
  latter being the copy a consumer gets from `register_stream_event_admin` — still
  print the storage sentinel. Filed as #201 rather than mixed in here.

Gates: 1550 SQLite, 1563 Postgres, ruff, pyright 0 errors, `zensical build`,
`api-reference-check` no drift.

PR #203, opened 2026-08-20
Collapse a fanned-out Update into one statement

### LINKED ISSUE #199: DjangoExecutor: batch Update effects that share a model and defaults into one statement
`DjangoExecutor.apply` issues one `UPDATE` per `Update` effect. When a handler fans a single logical change out across many rows — one effect per row, all sharing a model and identical `defaults` — that becomes one statement per row where one would do. In our projection a status fold emits one `Update` per typed row (a parent plus each of its repeater rows), so saving a form with eight repeating rows runs nine `UPDATE`s that all set the same column to the same value.

Collapsing the fan-out in the handler is the obvious workaround, but it is the wrong place: emitting one effect per row is deliberate for us, because a verification pass replays the same effects and diffs them against live rows row by row. Batching them in the handler would hide that detail from the oracle. Batching inside the executor keeps the effects per-row and the statement count low, and looks semantically free — `Update` is already `filter(**lookup).update(**defaults)` with no signals and no `auto_now` advance, so N single-row updates sharing defaults are equivalent to one `filter(pk__in=[…]).update(...)`.

Would you take a patch for this? Details, and the two constraints I think it needs to respect, in the comment below.

joshbrooks:
Closes #199. Thanks for the write-up — the design and both constraints were right,
and this is essentially the sketch with one addition.

Consecutive `Update` effects sharing a model and identical `defaults`, each
matching one equality on the same field, become a single
`filter(field__in=[…]).update(…)`. Everything else applies per-effect, so a mixed
batch degrades rather than raising.

**The addition: grouping has to be contiguous, not gathered.** Your constraint 2
said adjacent-only was the conservative version. It turns out to be the *correct*
one, not merely conservative. Two members of one group genuinely cannot collide —
same field with different values matches disjoint rows, and the same value is the
same lookup, which `check_disjoint_defaults` already rejects. But a third effect
with different defaults can match rows that two same-key effects also match,
through a lookup the collision check cannot relate to either, since it keys on the
exact lookup. Gathering across the batch hoists the later same-key effect above
it and changes which write lands last. A gathering implementation passed every
other test in the file; there is now one three-effect interleave that
distinguishes them.

Also worth knowing: the run ends at an `Upsert` for a reason the Ref case hides.
Deferring updates to the end of the write pass leaves Refs resolvable — their
producers have all run — but breaks update-if-exists, because an update that
correctly matched nothing then matches the row the upsert just created.

Your other constraints hold as stated: refs resolved before grouping, unhashable
defaults fall back, multi-key and traversal lookups fall back. Measured on the
reported shape, nine effects now issue one statement.


joshbrooks:
Reviewed with five rounds of `/review-loop`, fresh sub-agent each round. **The
collapse is now opt-in — `DjangoExecutor(batch_updates=True)`, default off.**

The loop found a distinct class of hole in four successive versions of one guard,
which is the reason for the flag more than any single finding:

1. A `None` lookup value was stripped from the `__in`, so the null-matching effect
   wrote nothing. Silent, and `resolved_at IS NULL` is this codebase's own open
   predicate — the shape a reconcile actually produces.
2. "The lookup field must not appear in `defaults`" closed one of four routes into
   a row-overlap hazard. The others: a foreign key reachable as both `area` and
   `area_id`; a `Ref` and a literal resolving to one value, where
   `check_disjoint_defaults` runs pre-resolution and sees two lookups while the
   grouper sees one; values distinct in Python but equal in the database.
3. "Nothing `Combinable`" missed `Q` — non-idempotent, hashable, accepted by
   `.update()`, not `Combinable`. And separately, grouping `defaults` by Python
   equality wrote `1` where `1.0` was meant, with no row overlap at all, so none
   of the idempotence reasoning applied.
4. The resulting allowlist still admitted `float`, `Decimal`, and aware
   `datetime`/`time`, whose `__eq__` deliberately ignores the representation the
   database stores: `-0.0 == 0.0` storing as `-0.0` and `0.0`;
   `Decimal("1.0") == Decimal("1.00")`; one instant in two time zones.

Round 5 found no fifth hole in the value logic and I believe that part is now
right — five types, each with a one-line proof of the property the grouping key
actually needs. But it did find two more instances of "`field__in=[v]` is not
`field=v`" on the least-examined side: an unbounded `IN` list raising past
SQLite's 32766 bind cap, exactly at the fan-out size the collapse exists for; and
an out-of-range integer that Django compiles away under `exact` but hands to the
driver under `IN`, aborting its siblings. Both fixed.

So the flag. The benefit here is small, known, and yours to ask for; the cost of
one more mistake was every consumer's writes changing silently. If it earns its
keep in production, turning it on by default later is a much easier argument to
make with evidence than with another review.

**For your side:** `DjangoExecutor(batch_updates=True)` and your nine statements
become one. Two things worth knowing. Only literal `defaults` collapse — `str`,
`bytes`, `int`, `bool`, `None` — so a status FK id qualifies and an `F()`
expression deliberately does not. And if your projection models are
history-tracked, collapsing N updates into one changes the number of history rows
written; nothing in this repo triggers per-row, but pghistory does, so check that
before enabling it on a tracked model.

joshbrooks:
Docs simplified, and the exclusions tested rather than argued.

**Why the docs were long.** They justified the value allowlist type by type, each with a Python counterexample — `Decimal("1.0") == Decimal("1.00")`, `-0.0 == 0.0`, one instant in two time zones. Those are true statements about Python. None of them is a statement about what a column stores, and that gap is what the wordiness was hiding.

**So I measured them.** `tests/test_django_rakaia/test_update_batching_equivalence.py` runs the same effects down both paths — unbatched on `default`, batched on `overlay`, identical starting rows — and compares `repr` of every column (not `==`, which would agree with itself on exactly the pairs under test). Then it *forces* the collapse past the allowlist and reports whether anything changes. Both backends:

| exclusion | forcing the collapse changes stored data? |
|---|---|
| expression (`F()`) over **overlapping** rows | **yes** |
| expression, disjoint rows | no |
| literal over overlapping rows | no — the safety argument, now measured |
| `float` `-0.0` vs `0.0` | no — round-trips as `0.0` |
| `Decimal("1.0")` vs `("1.00")` | no — the column's scale erases it |
| aware datetime, two zones | no — Django normalises to UTC |
| `1` vs `True` (with `type(v)` dropped from the key) | no — coerced to `1` on the way in |

So of six exclusions described as demonstrated, one is. The allowlist is still the right shape — conservative, and being wrong costs a missed collapse — but the docstring should not have claimed the others were dangerous, and now doesn't.

**Two findings.** The expression hazard needs *both* halves: non-idempotency is harmless without row-set overlap, and my first version of this harness reported everything clean precisely because it never staged the overlap — the lookup field has to be one the `defaults` write. Worth stating because a differential harness only finds what its seeds let happen, which is the same failure mode as the guard it is testing.

And: every `models.TextChoices` member is a `str` subclass, so `type(v) in _COLLAPSIBLE_DEFAULTS` is `False` and Django's own idiom for the commonest kind of status fan-out does not collapse. Safe to admit (`__eq__`/`__hash__` come from `str`, and Django stores the value) but out of scope here — pinned by a test so widening the allowlist is a deliberate change rather than a side effect.

**Anti-vacuity.** The harness asserts whether a collapse actually *happened*, not only that the rows agree. A case that declined and a case that collapsed both leave the columns identical, so without that assertion every test in the file would pass against an executor that ignored the flag.

Also confirmed not to be a hazard: `_update` does not wire `skip_unchanged`, and `ApplyReport` does not count `Update` effects — so the flag is unobservable through the return value as well as through the rows.

Gate green both legs: `pytest` 1594 passed, `just test-pg` 1607 passed, `ruff`, `pyright src/` 0 errors, `zensical build`.

joshbrooks:
`TextChoices` and `IntegerChoices` now collapse.

They were excluded by accident rather than by decision: `type(v)` for a `TextChoices` member is the enum class, not `str`, so the exact-type allowlist never matched one — and a status change fanned across many rows is the commonest shape of the case this feature exists for.

`_is_collapsible_value` admits an `Enum` member whose **value** type is on the allowlist and which **compares the way that type does**. Two conditions, each ruling out something different, and both mutation-covered:

| mutation | caught by |
|---|---|
| no enum widening at all | `test_a_text_choices_default_collapses`, `test_an_integer_choices_default_collapses` |
| drop the value-type test | `test_an_enum_over_a_str_subclass` |
| drop the equality test | `test_an_enum_that_overrides_equality`, `test_a_plain_enum_with_no_mixin` |
| admit any `Enum` | all three of the above |

The value-type test is what keeps out an enum over a `str` *subclass* — `v.value` is an instance of the subclass, not of `str`, so something free to define equality however it likes cannot get in — and one over a `float`. The equality test keeps out an enum that overrides equality on itself: two members with plain `str` values `'y'` and `'Y'` pass the value-type test, and a case-insensitive `__eq__`/`__hash__` makes them group, one member's value reaching the other's rows. **That one is demonstrated rather than argued** — forcing it changes what the column holds, unlike the representation cases, because these really are two different strings. Same discipline as the `F()` case.

A plain `Enum` with no mixin is declined even though identity equality is *stricter* and would therefore be safe: it stores `str(Bare.YES)` — `'Bane.YES'`-style, not `'y'` — so there is nothing to gain.

One SQLite-hid-it note, the same shape as the earlier `integer_field_range` one. That last fact is why a case moved from `History.marker` to `SukuProjection.status`: `'Bare.YES'` is eight characters into a `max_length=1` column, which Postgres rejects and SQLite quietly accepts. The first version of the test passed on the default leg for the wrong reason.

24 tests in the equivalence file. Both legs green: `pytest` 1601, `just test-pg` 1614, `pyright src/` 0 errors, `zensical build`.

Still unmeasured: whether the statement saving is a *time* saving. That needs a benchmark against a realistic save, and it now has a fair chance of being worth something, which it did not before this change.

joshbrooks:
Real-data cases pushed (`c76587a`). Four tests and a fixture pair copied off the production schema — a main row and its repeater rows, keyed by UUID with a denormalised status column. Every case in the file until now was invented to make a rule fail; none was the shape the consumer's database actually holds, and the shape turns out to set the ceiling.

**It corrects the size of the win, in both directions.** Measured over ~16.7k submissions: the worst real fan-out is **33 statements, not the 9 #199 reported**, and it collapses to **2, not 1** — the main row is in a different table from its repeaters. So the bound is the number of *tables*, not the number of rows: a form with three repeater tables costs 4 statements however many rows each holds. SF 1.3 pays that on nearly every save, its median fan-out being 8.

Still statement counts rather than wall-clock, but far better grounded than "nine becomes one".

**It also covers a hazard the equivalence file missed.** The real lookup column is a `UUIDField` receiving *both* a `UUID` and a `str` for the same key — the event carries `str(pk)`, the reader returns `UUID` objects. The equivalence tests only mixed `str`/`int` against a `CharField`. This sits squarely in the "`IN` and `exact` are not the same query" family that produced both of the round-5 defects, and Postgres has a native `uuid` type where SQLite stores char32, so I checked that leg specifically. Both fine: different rows under different types share one `IN` list and both get written.

And a two-sided finding worth keeping: `check_disjoint_defaults` canonicalises a lookup through `json.dumps(default=str)`, so its notion of row identity is *coarser* than the grouping key's — the same row arriving under both types collides there before it can ever reach the grouper. Coarser is the safe direction, and now something asserts it from both ends.

Verified before pushing (fast-forward, no rebase): `pytest` 1605 passed, Postgres 1618 passed, `ruff` + `ruff format --check` clean, `pyright src/` 0 errors. The new migration is in the test app's own numbering (`0009_repeaterprojection_submissionprojection`), so it does not collide with `django_rakaia`'s `0009_alter_stream_last_seq`.

joshbrooks:
Benchmarked. Fewer statements is less time, and the shape of the answer is more useful than the headline.

**Each statement not issued is worth about 0.4 ms**, near-constant from 2 statements to 501 — so any fan-out's saving is arithmetic rather than something to re-measure. Postgres 16, local container, one `apply()`, median of 150–300 runs:

| what is being saved | statements | before | after | |
|---|---|---|---|---|
| typical form, real SF 1.3 shape (3 repeater tables) | 8 → 4 | 5.9 ms | 4.4 ms | 1.3× |
| worst single form in their data (SF 2.3) | 33 → 2 | 17.2 ms | 3.7 ms | 4.7× |
| reconcile over 500 rows | 501 → 2 | 217 ms | 10.6 ms | 20.5× |

**It is not a network effect**, which I did not expect. An empty round trip to the same database is 0.1 ms against the 0.4 ms per statement, and re-running the whole benchmark on in-memory SQLite — round trips at 0.006 ms — still saves 0.13 ms per statement. Most of the cost is Django building the query and the server parsing and planning it. So the win survives a local socket, and grows with latency: at 1 ms RTT each statement costs roughly 1.3 ms, turning the worst form into about 40 ms saved rather than 13. (That last figure is arithmetic from the measured split, not a measurement.)

**So the recommendation is specific, and roughly the inverse of how #199 framed it.** This is a replay / reconcile / backfill option that also helps interactive saves a little — not a form-save optimisation. A typical save gains 1.3× because the bound is the number of tables rather than the number of rows; the reconcile gains 20× because there the bound does not bite.

Recorded in `docs/dry-run-and-executors.md` ("Is it worth turning on?") and the changelog, with the caveat that these are one machine's numbers and the per-statement figure is the part worth carrying.

The benchmark itself was a scratch file and is not committed. Worth adding as a re-runnable harness under `scripts/` if you want it — I left it out rather than put a timing-sensitive test near CI.

PR #204, opened 2026-08-20
One read of the stream row per append, not two

### LINKED ISSUE #202: DjangoStreamStore.append: 13 queries per call, with the stream row and watermark each read twice
A single `DjangoStreamStore.append` to an existing stream costs 13 queries, and three of them look like duplicates of work already done in the same call: the stream row is selected twice, the offset watermark is selected twice, and a `MAX(offset)` scan asks for a number the watermark row exists to hold. `append_many` amortises all of this and is flat at 13 queries for 1, 4, 9 or 20 events — but a single append pays the full amount, and plenty of callers legitimately append one event at a time.

We have already moved our batchable path to `append_many` (a nine-event save went from ~95 queries to ~20), so this is not blocking us. It is a report that the per-call floor looks reducible by roughly half, which would benefit every single-event append without any caller change.

joshbrooks:
Every append to the durable store looked up the same stream twice — once before the write began, and once when the write actually started. The first lookup was only there so that a stream which had aged out still got cleaned up, because the cleanup was otherwise undone when the append gave up and reported the stream as missing. One lookup now does both jobs, and an append that finds an aged-out stream still cleans it up.

Closes #202, which reported thirteen database queries per append and named three that looked like duplicates. Measured against `main` today it is eight, not thirteen: two of the three went in #175. This removes the third, leaving seven, and every one of them now does something. There is a new test that writes down what each kind of append costs, statement by statement, because none of this changes anything a caller can see — the cost can drift back up without any other test noticing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,32 @@ runnable demo for each.
 
 ### Changed
 
+- **A single append reads the stream row once, not twice.** Every
+  `DjangoStreamStore.append` did two `SELECT`s of the same row: an unlocked one
+  before the transaction opened, purely so that an expired stream would still be
+  reaped when the write went on to report `StreamNotFound` — the reap is a
+  delete, and the rollback that reports the 404 would otherwise undo it — and
+  then the locked one the write actually decides against. The expiry now leaves
+  the transaction as a signal and is reaped on the far side of the rollback, so
+  the locked read does both jobs.
+
+  A steady-state append is **7 statements, down from 8** — `BEGIN`, the locked
+  stream read, the event `INSERT`, the locked high-water read, its `UPDATE`, the
+  entry `INSERT`, `COMMIT` — and each one of them now does something. The
+  reporter of #202 measured 13 against 0.2.0 and named three statements that
+  looked redundant; the other two, a high-water read repeated either side of a
+  get-or-create and a `MAX(offset)` scan beside the watermark it duplicates, both
+  went in #175 and are already absent from a steady-state append. The
+  set of statements is now pinned per append shape by
+  `tests/test_django_rakaia/test_append_query_cost.py`, because none of this
+  changes an observable outcome and so nothing else in the suite notices when a
+  statement comes back.
+
+  The first append to a path still costs four more: the high-water row has to be
+  created, and seeded from a `MAX(offset)` scan, because `high = 0` means either
+  "new path" or "install upgraded across migration 0005" and only the entry table
+  can tell those apart. That is once per path, ever.
+
 - **The value-equality rule moved to `django_rakaia.canonicalisation`.**
   `canonical_value`, `Normalizer`, `DEFAULT_NORMALIZERS` and the three
   `normalize_*` functions were defined in `verification.py`, which meant the write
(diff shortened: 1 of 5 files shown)


joshbrooks:
## Measurements

`CaptureQueriesContext` around one append, Postgres 16, steady state (watermark row present). Same counts on the SQLite leg — both backends report `BEGIN`/`COMMIT` through the cursor.

| | `main` | this branch |
|---|---|---|
| steady-state `append` | 8 | **7** |
| first `append` to a path | 12 | **11** |
| `append_many(1 / 4 / 20)` | 8 | **7** |

The seven: `BEGIN`, `SELECT … rakaia_stream … FOR UPDATE`, `INSERT rakaia_streamevent`, `SELECT … rakaia_streamoffsetwatermark … FOR UPDATE`, `UPDATE rakaia_streamoffsetwatermark`, `INSERT rakaia_streamentry`, `COMMIT`. That is #202's "plausible floor of ~6" plus transaction control.

Against the issue's three candidates, as of `main`:

1. **stream row read twice** — real, and what this PR removes. `_reap_if_expired` before `atomic()` was an unlocked read whose only purpose was making the expiry delete outlive the rollback that reports `StreamNotFound`. `_locked_write` now takes one locked read and carries the expiry out of the transaction as an internal signal, reaping past the rollback.
2. **watermark read twice** — already gone (#175 folded the plain `get_or_create` and the `select_for_update().get()` into one locked `get_or_create`).
3. **`MAX(offset)` beside the watermark** — already gone in the steady state (#175). It still runs on the *first* append to a path, with the `get_or_create` insert and its savepoint: `high == 0` means either "new path" or "install upgraded across migration 0005", and only the entry table separates those. Once per path, ever.

Not attempted: collapsing the watermark `SELECT`+`UPDATE` into one `UPDATE … RETURNING`, or the first-append `get_or_create` into an upsert. Both need backend-specific raw SQL, and the `select_for_update()` they would replace is one of the three lock sites `CLAUDE.md` says are covered deliberately. Worth one query per append, but not inside this change.

## Review

Four `/review-loop` rounds, each fixing what the previous one found and re-verified by the next. Round 1: a cost test that compared the 50th append to the 1st, and so was blind to the per-append constant its docstring claimed to guard — the `MAX(offset)`-always mutation left it green while five siblings went red. Round 2: the module docstring gave a false reason for filtering transaction control. Round 3: `create`'s pre-transaction reap could be deleted with the whole suite green, so the justification written into its docstring was prose only; it now has a test (an expired JSON stream recreated with a non-JSON body). Round 4: one clause of that test's docstring overstated what had not yet been written. Every test in the PR was shown to fail with its mechanism broken, and the stream-row lock is still covered by `TestStreamRowLock` on the Postgres leg.

## Adjacent, pre-existing, not fixed here

The expiry reap is `self.delete(path)`, an unconditional delete-by-path with no lock held. A competitor that recreated the stream in the window between the lock being released and the delete landing has its live stream deleted. Confirmed present on `main` — `_get_if_not_expired` has the same shape, and it is on every read path — so not a regression, but this PR does put the delete immediately after a `FOR UPDATE` is released, which is exactly when a blocked competitor is let in. Happy to file it separately.

Issue #190, opened 2026-08-20
The fast verification path makes you pass the same list twice


joshbrooks:
There is a faster way to check a rebuild that avoids one database query per row. Using it means building a helper with your list of expected writes, then passing the same list again to the comparison, with an unwritten rule that both have to be the same list. Nothing enforces it, and getting it wrong produces a wrong answer rather than an error.

Since the comparison already receives that list, it could build the helper itself and the fast path would cost one option instead of one extra object plus two rules to remember.

PR #205, opened 2026-08-21
One call to prove a rebuild is trustworthy

### LINKED ISSUE #184: Proving a rebuild is trustworthy takes six calls and some discipline
To get an answer you can rely on, a caller has to copy the log to memory, arm two different guards, build a collector, run the replay, build a snapshot reader, and then run the comparison — and separately remember that a pass means nothing unless they also deliberately break one of the guards to check it was armed.

That is six interfaces for one question, and the discipline part is not enforced anywhere. The first consumer to use this left one of the guards unwired for months. It should be one call that arms its own guards, checks that it checked something, and hands back the answer.

joshbrooks:
Closes #184.

Checking that a rebuild is trustworthy used to take six separate calls in the right order, plus one rule that was written down but not enforced: the answer means nothing unless the safety guards were actually switched on. Nobody remembers to check that, and the first team to use this left one of the two guards switched off for months without noticing.

There is now a single call that does the whole thing and checks its own work. It deliberately breaks one guard to confirm it is live, refuses to run on top of a previous run's leftovers, and reports whether anything was actually compared rather than just whether anything disagreed. The individual pieces still work on their own for anyone asking a different question.


joshbrooks:
Technical detail.

**Interface.** `rebuild_and_verify(stream_path, *, into, live_models, source=None, live_using="default", registry=None, upcaster_registry=None, normalizers=None, event_match=None, on_drift="warn") -> DiffReport`. Subsumes #190 (the caller no longer builds the `PreloadedProjectionReader`).

**Three things it does that the six-call assembly did not.**

1. `_prove_read_guard_armed` issues a deliberate `SELECT` on the guarded alias inside the deny region and raises `GuardNotArmed` if it does not fail. This is the check `hermeticity.py` documented and delegated to the caller. Note it deliberately does *not* also call `armed_deny_aliases()`: the probe strictly subsumes a wrapper-presence check (no wrapper → the query succeeds → it raises anyway), and a mutation pass showed the presence check made the probe's own `raise` unreachable — two checks where the weaker hides the stronger is one check plus dead code.
2. `ScratchAliasNotEmpty` when `into` already holds rows for any of `live_models`. A leftover row is read by stage > 0 handlers as if the log produced it, and the diff is against *live*, so nothing downstream notices. It refuses rather than truncating: deleting the caller's database on the strength of a keyword argument is the worse of the two behaviours.
3. `live_models` is required, no default. `assert_no_live_writes()` with no models is a documented no-op, so a default would disarm half the gate silently.

**Why a recording tee rather than `CollectingExecutor`.** Both halves are needed at once. Applying is what lets a stage > 0 handler read what stage 0 wrote — the fact a collector cannot supply, and the reason a from-scratch rebuild of cross-form links is not verifiable with one. And `ReplayResult` exposes `effects_applied` as a *count*, so the effects the diff needs have to be captured on the way past. `_RecordingExecutor.apply` materialises the batch before forwarding; `Executor.apply` takes `Iterable`, so a generator would otherwise reach the inner executor exhausted — recorded as verified, applied nowhere.

**Which alias the diff reads.** `PreloadedProjectionReader(..., using=live_using)`, not `into`. Pointing it at `into` is a tautology that certifies everything; it is mutation-covered by `test_a_drifted_row_is_reported_not_raised`.

**The drain, and a docs correction.** `_Drained` is a one-method `ReadableStore` holding the messages verbatim — no re-append, so offsets, `label`, `metadata` and `event_ts` survive as the durable log recorded them. Both `hermeticity.py` and `docs/dry-run-and-executors.md` claimed `DjangoStreamStore(using=)` removed the need for the drain. It removed the *untested* part. The docs' example puts the log on the scratch alias, which is only true for a restored dump; when the log is on the alias being guarded — the ordinary case — the messages must leave it first, and no keyword argument can make reading the guarded database acceptable, because that is the property under test. Both are corrected here.

**Testing.** 15 tests, 10/10 mutations caught, each by the test written for it:

| mutation | caught by |
|---|---|
| drop `_require_empty_scratch` | `test_a_stale_scratch_alias_is_refused_rather_than_rebuilt_onto` |
| drop `_prove_read_guard_armed` | `test_a_disarmed_read_guard_is_refused_not_certified` |
| probe stops objecting (drop the `raise`) | same |
| `deny_database_access()` with no alias | 11 tests |
| drop `*live_models` from the write guard | `test_the_write_guard_covers_more_than_the_replay` |
| diff reader → `using=into` | `test_a_drifted_row_is_reported_not_raised`, `test_a_missing_live_row_is_a_difference` |
| handler reader → default alias | `test_a_later_stage_sees_what_the_earlier_stage_wrote` |
| executor → default alias | 8 tests |
| replay the live store, not the drain | 9 tests |
| tee does not materialise the batch | `test_it_applies_a_generator_it_has_also_recorded` |

**One test failure worth recording**, since it is the CLAUDE.md gotcha biting from the other side. `test_a_write_the_read_guard_cannot_see_is_still_caught` demonstrates the leak only the write guard catches — a write on another thread, invisible to `deny_database_access` because Django keeps connections in a thread-local (which is why `armed_deny_aliases` exists). It is skipped on SQLite, where a second connection blocks on the test transaction. On Postgres it ran, and its write committed *outside* pytest-django's rollback — polluting 8 later count assertions across `test_rebuild_isolation.py` and `test_using_seam.py`. Fixed with a per-test `django_db(transaction=True, ...)`. The lesson is the inverse of the usual one: a test skipped on the default leg gets no scrutiny for the discipline the other leg requires.

Full gate green: `ruff check`, `ruff format --check`, `pyright src/` (0 errors), `pytest` (1572 passed), `just test-pg` (1586 passed), `zensical build`, `just api-reference-check`.

joshbrooks:
Review loop: 2 rounds.

**Round 1** — seven green mutations, all the same shape: an argument that stops being forwarded to `replay`, or an alias hardcoded back to `"default"`. `live_using` was the serious one — four use sites (drain, write guard, read guard, diff reader), every test leaving it at its default, so all four could be replaced with a literal and the suite stayed green. In a gate whose job is to say whether a verdict is evidence, that is the worst available failure mode: the call still returns a `DiffReport` and it still looks like proof. Fixed with `TestWhatItForwardsToReplay` (3) and `TestProductionNeedNotBeTheDefaultAlias` (2, running the whole gate with the aliases swapped, one with production disagreeing so the *diff* is pinned to the parameter too). Also fixed a comment claiming the write guard wraps the drain — it cannot, since the drain is the read that moves the log off the guarded alias.

**Round 2** — nothing blocking. Re-ran all seven of round 1's mutations independently and confirmed each turns its named test red; probed the five new tests for defects of their own (state leaking from the two that rebuild into `default`, vacuous assertions, the `event.pop` in the upcaster lambda) and found none; found two green mutations and correctly judged both non-defects (`_Drained`'s discarded `up_to_date` flag, and `_require_empty_scratch` ordering, which is order-indifferent because `into != live_using`). One accurate nit: a comment listed a renamed stream path among the vacuity risks, when it actually raises `StreamNotFound` from the drain. Fixed.

Zero of round 2's findings were defects in round 1's fixes.

Issue #138, opened 2026-08-15
Our stream positions have a size limit we chose by hand


joshbrooks:
Every position we hand out is a zero-padded number, and the padding width was
picked by hand — sixteen digits in one backend, twenty in the other. The code
says so, and warns that a backend which ever exceeded the limit would have to
widen the field and the starting value together, in step. Both backends also
have to coordinate to hand out the next position, which costs a lock and a
couple of database round trips on every write.

The protocol suggests an alternative that removes both problems, and names it
specifically. It would also make positions unique without anyone counting. But
it moves the ordering guarantee from a database lock onto the clock, which is a
real trade and not obviously the right one for us.

joshbrooks:
**What the spec says.** Upstream `PROTOCOL.md` §1192: offsets **MUST** be lexicographically greater than all previously assigned; implementations **MUST NOT** use schemes such as raw UTC timestamps that can duplicate or go backwards; and *"Time-based identifiers like ULIDs, which combine timestamps with random components to guarantee uniqueness and monotonicity, are acceptable."* Format guidance nearby: opaque, case-sensitive, no `,` `&` `=` `?` `/`, URL-safe, under 256 characters.

A ULID is 128 bits as 26 Crockford-base32 characters — a 48-bit millisecond timestamp big-endian first, then 80 bits of randomness. Because the timestamp leads and the alphabet is in ASCII order, byte-wise comparison equals time order. Fixed length, no width decision, URL-safe, and the alphabet omits `I`/`L`/`O`/`U`.

**What we do now.** In-memory: `f"{read_seq:016d}_{new_byte_offset:016d}"` (`src/rakaia/store.py:652`), with a comment at `:645-651` stating the ordering guarantee "holds only while both fields stay < 10**16" and that a durable backend exceeding it "must widen the fields (and INITIAL_OFFSET) in lockstep". Durable: 20-digit zero-pad (`src/django_rakaia/offsets.py:21-29`), allocated through `StreamOffsetWatermark` with `select_for_update()` (`src/django_rakaia/models.py:112-153`).

**What ULIDs would buy:**
- the hand-chosen width ceiling and its lockstep-widening caveat both disappear
- no coordination to mint one — no read of the current maximum, which is the redundant `SELECT` plus per-append `Max(offset)` aggregate already logged as the first item in #62
- monotonicity across delete-and-recreate for free, which is currently the entire reason `StreamOffsetWatermark` exists and survives stream deletion (#34 defect 2)

[...]

PR #206, opened 2026-08-21
One home for each of a stream position's five rules

### LINKED ISSUE #178: Comparing two stream positions guesses when it cannot know
A consumer that remembers where it got to in a stream can be told it has fallen behind when it is actually ahead, and the reverse. It happens when the position was saved by one storage backend and is then compared against the other — which is what moving from the in-memory store to the database-backed one involves, and that move is a documented upgrade path.

The two backends write positions in different shapes. The comparison tries to read both shapes, cannot line them up, and quietly falls back to comparing them as text, where the answer is wrong rather than merely uncertain. Nothing raises, so the consumer acts on a confident wrong answer.

### LINKED ISSUE #182: The rules for a stream position live in one place for one store and are scattered for the other
A stream position has about five rules: what the first one is, how wide it is, how to make the next one, how to tell a valid one, and how to compare two. For the database-backed store all five live in one small file. For the in-memory store they are spread across two files with several stated inline, and the rule for comparing two positions has no home at all — which is why a comparison ended up guessing, and getting it wrong.

Both stores are real, working implementations of the same idea, so there is a genuine shared shape here that nobody has written down. Doing so would also give the hand-picked width its one home.

joshbrooks:
Closes #182. Closes #178. Addresses #138.

A stream position has about five rules: what the first one is, how wide it is, how to make the next one, how to tell a valid one, and how to compare two. One store keeps four of its five in a single file. The other has them scattered across two files with the size written out three times, under a comment saying a change has to update all three together — which is a rule telling you it has no home.

The fifth rule, comparing two positions, had no home in either. So the one place that needed it made its own guess, and the guess was wrong: a reader four events into a stream could be told it was past the end of a stream of forty-two. It now refuses to answer rather than guessing, because there is no correct answer to give — the two stores count different things.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,10 +22,17 @@ runnable demo for each.
 
   `OffsetFormat` now owns all five per store, as `COMPOUND` (the in-memory
   `{seq}_{byte}`) and `PLAIN` (the durable padded integer). The hand-picked
-  widths are written once each, and the pattern that validates a token is derived
-  from the widths that render it, so a format cannot be checked against a shape it
-  no longer produces. `ForeignOffset` is exported for consumers that catch it; it
-  subclasses `InvalidOffset`, which both stores already raised.
+  widths are written once each, and the shape that recognises a token is derived
+  from the field count that renders it, so a format cannot be checked against a
+  shape it no longer produces. `ForeignOffset` is exported for consumers that
+  catch it; it subclasses `InvalidOffset`, which both stores already raised.
+
+  The refusal is narrow on purpose: it fires only when both tokens are
+  recognisably from *different* rakaia stores, the one pair with no answer. A
+  format this library does not recognise — a third-party `CursorStore` issuing
+  ULIDs or timestamps, which `protocols` documents as a supported backend — is
+  still compared byte-wise, which is the ordering rule the protocol states for
+  any opaque offset.
 
 - **`DjangoExecutor(batch_updates=True)` collapses a fanned-out `Update` into one
   statement.** A handler that fans one change across many rows emits one `Update`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,22 @@ runnable demo for each.
 
 ### Added
 
+- **`rakaia.offsets` — a stream position's rules, one home each.** An offset has
+  about five rules: what the first one is, how wide it is, how to make the next
+  one, how to tell a valid one, and how to compare two. The durable store kept
+  four of its five in one file. The in-memory store had them spread across two
+  files with the field width written out three times — and a comment saying a
+  change must update them "in lockstep", which is a rule announcing that it has
+  no home. **The fifth, comparison, had no home in either**, so the one place
+  that needed it carried its own guess.
+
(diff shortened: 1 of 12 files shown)


joshbrooks:
Technical detail.

**The defect.** `subscription._after` parsed both tokens into int tuples and, when it could not line them up (`len(ka) != len(kb)`), fell back to `a > b`. The two formats are padded to different widths, so `'0000000000000000_0000000000000008'` sorts *above* `'00000000000000000042'` — `'0' < '3'` settles it before any digit that means anything. `poll` read that as "the cursor is beyond the head" and returned `rewound`, which instructs a consumer to discard its derived state and re-read from the start. The state converges, so nothing is corrupted; the reason is false and the re-read is for nothing. The reverse direction already raised `InvalidOffset`.

**Why refusing, not fixing.** `protocols.ReadableStore` already tells consumers to compare offsets from *the same store*: an in-memory `{seq}_{byte}` counts events and bytes, a durable offset is an entry id. There is no cross-store ordering to compute, so the only options were refuse or guess.

**Shape.** `OffsetFormat` owns all five rules per store — `COMPOUND` (in-memory, widths `(16, 16)`) and `PLAIN` (durable, widths `(20,)`). The validating pattern is *derived* from the widths that render, so a format cannot be checked against a shape it no longer produces. Each store's existing "not an offset I issued" guard now delegates rather than restating it; `ForeignOffset` subclasses `InvalidOffset`, so every existing `except InvalidOffset` is unaffected.

**Three mutations survived a first pass, and each named a real gap:**

| mutation | why it survived | now caught by |
|---|---|---|
| drop the `is None` half of `after`'s guard | still refuses when *one* token is unrecognised, via `None is not PLAIN` — only the **two**-unknown case falls through, to `None.key(...)` | `test_two_unrecognised_tokens_are_refused_not_crashed_on` |
| `require()` never raises | nothing tested a format against the other store's token directly | `test_a_format_rejects_the_other_store_s_token`, `test_the_store_rejects_a_foreign_offset_on_read` |
| `key()` skips `require()` | — | three existing `django_rakaia` tests, in files I was not running |

The second is worth a note: with the format check disabled the **full suite hangs** rather than failing. A foreign offset sorts below every token the in-memory store issues, so a long-poll waits on a position that can never arrive. That is how the gap was found. The third reintroduces the hazard the old `parse_offset` docstring warned about — `int()` reads `_` as a digit separator, so `int('0000000000000000_0000000000000008')` is `8`, and a resume read would silently return the wrong window.

**#138** (the size limit chosen by hand) is addressed rather than closed by fiat: both widths now have exactly one home, with the reasoning for each next to it. Closing it is your call.

**Scope note.** `docs/django-integration.md` presents the in-memory store as the default with a documented "Migration path (in-memory → durable)", which is what made #178 look like a live production bug. It isn't — the in-memory store is for tests, demos and the conformance suite. Correcting that, and flipping the `RAKAIA_STORE` default to `"durable"`, is a breaking change and is going in a separate PR rather than riding along with a refactor.

Gate green both legs: `pytest` 1661 passed, `just test-pg` 1675 passed, `ruff`, `pyright src/` 0 errors, `zensical build`, `just api-reference-check`.

PR #208, opened 2026-08-21
Two more admin screens showed a label no reader reports

### LINKED ISSUE #201: Two more places in the admin show a label no reader reports
The event list in the admin was fixed to stop showing the internal placeholder we
use for an event that carries no label. Two other admin screens do the same thing
and were missed: the entries list, and the one built by the helper that lets an
app register its own event model. Both still print the placeholder, so a plain
append shows as APPEND there while the audit trail and the fixed screen both
report no label.

Same one-line cause in all three, which is why fixing one did not fix the others.
Found while reviewing an unrelated change; details in a comment.

joshbrooks:
The event list in the admin was fixed to stop showing the internal placeholder we use for an event that carried no label, but two other screens were missed: the entries list, and the one an app gets when it registers its own event model. Both still printed the placeholder, so a plain append showed as APPEND there. The registered-model screen also previewed a binary body as unlabelled gibberish and emitted markup that came out as visible tags.

All three screens now render the label and the payload through one shared piece of code instead of each keeping its own copy, so the next screen cannot disagree by forgetting. Each screen has its own test for both cases, and each fix was reverted on its own to check the test catches it.

Closes #201.

### DIFF
--- a/src/django_rakaia/admin.py
+++ b/src/django_rakaia/admin.py
@@ -10,7 +10,7 @@ from typing import Any
 
 from django.contrib import admin
 from django.utils.html import format_html, format_html_join
-from django.utils.safestring import mark_safe
+from django.utils.safestring import SafeString, mark_safe
 
 from django_rakaia.event_message import decode_payload, event_label
 from django_rakaia.models import Stream, StreamEntry, StreamEvent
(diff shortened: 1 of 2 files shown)


joshbrooks:
Technical detail.

Three badge copies at `admin.py:72` (fixed in #195), `:165` (`StreamEntryAdmin`) and `:217` (inside `register_stream_event_admin`). Now two module-level helpers, `_event_badge(event_type)` and `_event_data_preview(data, payload_encoding)`, called by all three; `_BADGE_COLORS` replaces the three duplicated colour dicts. The badge goes through `event_label`, which inverts the `"append"` sentinel, and renders an em-dash for the empty result so an absent label is not a blank cell. The preview goes through `decode_payload`, and describes a non-UTF-8 body as `<N bytes, base64>` rather than printing the stored base64.

The factory copy of `data_preview` also did `.replace("\n", "<br>").replace(" ", "&nbsp;")` on a plain `str`, which Django escapes — so those screens showed literal `<br>` text. It also truncated at its own 100/97 constants instead of `_truncate`.

Tests are in `TestEveryAdminScreenAgrees` in `tests/test_django_rakaia/test_admin.py`, one per screen, asserting rendered output. The registered-model screen is exercised through `django_admin.site._registry[AppStreamEvent]` — the instance a consumer would actually be served, since `tests/test_django_rakaia/models.py` registers it at import. A last test asserts all three render the same append identically.

Mutation results, each fix reverted alone:

- entries-list badge back to `obj.event.event_type.upper()` — 4 red, including `test_the_entries_list_shows_no_label_for_a_labelless_append`
- factory badge back to `obj.event_type.upper()` — 3 red, including `test_a_registered_event_model_shows_no_label_for_a_labelless_append`
- factory preview back to the stale copy — 4 red, including `test_a_registered_event_model_does_not_preview_a_body_as_base64` and `..._does_not_emit_raw_markup`

Gate: `ruff check`, `ruff format --check`, `pyright src/` (0 errors), `pytest` (1656 passed, 16 skipped, 1 xfailed), `zensical build` all green. No public API change, so `docs/api-reference.md` is unchanged.

joshbrooks:
Review loop: two rounds, both reporting nothing blocking.

Round 1 found a green mutation — `test_a_registered_event_model_previews_a_text_body_as_its_text` asserted containment, and `json.dumps("hello, world")` contains `hello, world`, so a screen ignoring `payload_encoding` left it green. The same weakness was in the test #195 wrote for the event list. Both now assert the exact preview (`8517c3b`), and mutating `_event_data_preview` to ignore the encoding turns both red.

Round 2 confirmed that fix and found one more green mutation: dropping `_truncate` from the decoded-text branch left all 22 admin tests green, because every encoding test used a short body. Only the JSON branch was pinned. `666ce31` adds the missing case; removing the call now reddens the test named for it.

Round 1 also noted that `list_filter` and `StreamEvent.__str__` still render the raw column. Out of scope for #201 — filed as #210.

PR #209, opened 2026-08-21
Asking for the faster rebuild check no longer means passing the same list twice

### LINKED ISSUE #190: The fast verification path makes you pass the same list twice

joshbrooks:
Checking a rebuild without one database query per row used to mean building a helper with your list of expected writes and then handing the same list to the comparison, with nothing checking that the two matched. Getting it wrong gave you an answer instead of an error. The comparison now takes a single `preload=True` option and builds the helper itself from the list it was given, so there is no second list to keep in step.

The helper is still available on its own for anyone who needs it, but handing one back to the comparison must now cover the writes being checked, and asking for both a helper and the option is refused rather than one quietly winning. The option is documented as being for read-only checking: it reads everything once up front, so it is the wrong tool while a replay is still writing rows.

Closes #190.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,24 @@ runnable demo for each.
   The individual pieces stay exported. This is the shortest route to a
   trustworthy answer, not the only permitted one.
 
+- **`diff_effects_against_rows(effects, preload=True)` — the bulk verification
+  path as one option.** Avoiding one query per row used to mean building a
+  `PreloadedProjectionReader` with your effects and then passing *the same*
+  effects to the diff, with nothing enforcing the "same". Get it wrong and you
+  got a report rather than an error: one resting partly on a snapshot of a
+  different batch, or — handing a generator to both — on nothing at all. The flag
+  builds the reader from the list the diff is already given, so there is no second
+  list to keep in step, and `using=` names the alias that reader reads from.
+
+  The bulk fetch is a point-in-time snapshot, so the flag is for read-only
+  verification and stays off during live staged replay, where the rows change as
+  the replay writes them. `PreloadedProjectionReader` remains exported for a
+  caller that needs it standalone, but handing one back to the diff now has to
+  cover that call's effects — `PreloadMismatch`, a new exported error, instead of
+  a half-snapshot answer — and `reader=` together with `preload=`/`using=` is a
+  `TypeError` rather than one of them quietly winning. `rebuild_and_verify()`
+  keeps its single bulk read and no longer composes the reader itself.
+
 - **`DjangoExecutor(normalizers=...)`, and `Normalizer` as an exported name.**
   Deciding whether a stored value differs from the one the log carries is one
   question, and two paths asked it: `diff_effects_against_rows` to check that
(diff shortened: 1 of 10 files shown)


joshbrooks:
Technical detail.

**Shape.** `diff_effects_against_rows(effects, *, reader=None, preload=False, using=None, ...)`. `preload=True` materialises `effects` once and builds `PreloadedProjectionReader(effects, using=using)` from that same list; `using=` names the alias for whichever reader the function builds (previously only reachable by constructing a reader by hand, which is what forced the two-list pattern on `rebuild_and_verify`). `reader=` with `preload=`/`using=` raises `TypeError` — the combination has a defined answer rather than a precedence.

**The residual mismatch is now loud.** `PreloadedProjectionReader` records the keys its bulk fetch requested (`_preloaded`, deliberately not `_cache`, so a memoised live fallback does not count — it is a reading taken at another moment). When a `PreloadedProjectionReader` is passed as `reader=`, the diff checks every preloadable lookup it is about to make is covered, and raises the new `PreloadMismatch` otherwise. Empty and relation-spanning lookups are excluded: they are documented as always-live, so their absence says nothing about which batch built the reader.

**Live staged replay.** Nothing can detect it from inside the diff, so it is handled by the flag being opt-in and off by default, with the constraint stated on the parameter, in the reader docstring, and in `docs/projection-cookbook.md`.

**Tests** (`tests/test_django_rakaia/test_preloaded_reader.py::TestPreloadOption`, plus one in `test_rebuild.py`): fast path agrees with the slow path on the same input including a RED verdict; query count 1 vs 10 on a 10-effect sweep (`CaptureQueriesContext`, the idiom in `test_append_query_cost.py`); a one-shot generator is consumed once (the old pattern gave the diff an exhausted iterator and compared nothing); both `TypeError` combinations; a reader built from another batch, and one laundered by a live `get()` first, both raise; a covering batch is still accepted; `using="overlay"` routes both readers (`django_db(transaction=True, databases=[...])` per CLAUDE.md); and `rebuild_and_verify` still reads the live rows in exactly one non-count query.

**Mutation check.** Each mechanism was broken in turn and the test named for it went red: the `TypeError` guard, `preload` building the preloaded reader, the `list(effects)` materialisation, the coverage check, `_preloaded` vs `_cache`, `using=` plumbing, and `rebuild.py`'s `preload=True`. No green mutations.

Gate: `ruff check`, `ruff format --check`, `pyright src/` (0 errors), `pytest` (1658 passed), `zensical build`, `just api-reference-check` all pass. New exported name `PreloadMismatch` — `__init__._EXPORTS`, `scripts/gen_api_reference.py` GROUPS, the pinned public-API set, and the regenerated `docs/api-reference.md` are all updated.

PR #211, opened 2026-08-21
Decide a batch write with one rule, not one per store

### LINKED ISSUE #181: The two stores disagree about what a batch write is allowed to do
Deciding whether a single write is allowed is one shared rule that both storage backends call, which is what stops them drifting. Deciding whether a *batch* of writes is allowed is not shared: each backend grew its own version, and the two are genuinely different. One carries the full rule; the other checks less and skips the part that fences out a stale writer.

So a caller has to know which backend it is talking to in order to predict what a batch will do. This has already caused one defect. The shared rule needs to cover a batch, not just a single write.

joshbrooks:
Deciding whether a single write is allowed has been one shared rule since #167, which is what stops the two storage backends drifting. Deciding whether a *batch* is allowed was still written twice, and the two versions really did disagree — in both directions. One of them refused a batch over a sequence number nothing had actually written; the other, when the stream was already closed, told a writer "closed" where the single-write path would have told it "that was your own write, it landed". A caller had to know which backend it was talking to in order to predict what a batch would do.

There is now one rule for a batch too, sitting next to the one for a single write: what makes a batch admissible, and how the picture of the stream moves along from one item to the next. Both backends call it. #167 had already done the per-item half of this; what was left was the batch level, which is what this finishes. The four new tests run against both backends, and each was failing on exactly one of them before the change.

Closes #181.

### DIFF
--- a/src/django_rakaia/django_store.py
+++ b/src/django_rakaia/django_store.py
@@ -43,7 +43,11 @@ from typing import Any
 from asgiref.sync import sync_to_async
 from django.db import transaction
 
-from rakaia.append_decision import AppendVerdict, StreamFacts, decide_append
+from rakaia.append_decision import (
+    StreamFacts,
+    decide_append,
+    decide_append_batch,
+)
 from rakaia.context import merge_provenance
 from rakaia.json_mode import (
     format_json_response,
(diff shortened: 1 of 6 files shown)


joshbrooks:
Technical detail.

**The two pre-flights genuinely disagreed, in both directions.** Pinned first as
characterising tests in `tests/server_store_contract.py`, which both backends
subclass; three of them were red before the change, each on one backend only:

- `test_append_many_does_not_advance_seq_for_a_refused_item` — **in-memory red.**
  Its hand-rolled scan advanced `last_seq` for every item carrying a `seq`,
  including one the fence would refuse. A batch of `[stale-epoch seq="9", plain
  seq="3"]` raised `SequenceConflict: 3 <= 9` against a sequence nothing wrote.
  Django accepted the second item, correctly.
- `test_append_many_to_a_closed_stream_recognises_the_closing_producer` —
  **Django red.** `if stream.closed: return [...]` short-circuited before
  `decide_append`, dropping the `ProducerDuplicate` that
  `append`/`append_with_producer` give for a re-send of the closing tuple.
- `test_append_many_items_after_a_close_are_admitted_like_a_loop` — **Django
  red.** Same cause one level in: `items[cut:]` got a bare `stream_closed=True`
  with no `producer_result`.

Also added `test_append_many_fences_the_second_item_against_the_first`, which was
already green on both — the in-batch fencing #167 introduced, now pinned in the
shared suite rather than only in `test_django_store.py`.

**Not a semantics decision.** In each case the right answer was already written
down: both `append_many` docstrings promise "semantically identical to calling
`append` once per item", and a loop of `append` settles all three unambiguously —
`last_seq` is only assigned after a write lands, and a post-close item observes a
real `closed_by`. So this stayed a refactor plus two bug fixes rather than a
choice.

**Where the rule lives.** `decide_append_batch` in `src/rakaia/append_decision.py`,
returning a `BatchVerdict` of one `AppendVerdict` per input item plus the
`last_seq` the batch leaves behind and the options object that closed it. It is
pure and copies `producer_states`, so a store can abandon a batch on a conflict
without having had its own view advanced under it. The two stores consume it
differently by design: Django uses the verdicts directly (one transaction, one
offset block); the in-memory store calls it for the all-or-nothing pre-flight and
then *is* a loop of `append`, discarding the verdicts because `append` re-derives
each one against really-advanced state. `store.py` no longer imports
`ContentTypeMismatch` or `SequenceConflict` at all — ruff removed them, which is
a decent signal the duplication is gone.

**Unit cover.** `TestBatchAdmissibility` in `tests/test_rakaia/test_append_decision.py`,
11 cases in the same table style as the existing file — no store, no database.

**Postgres leg.** `just test-pg`: 1681 passed, 2 skipped, 1 xfailed. No lock test
added, deliberately: `_locked_write` is untouched, no `select_for_update()` site
changed, and `CLAUDE.md` is explicit that a lock test that cannot be shown to
fail with the lock removed is worse than none.

**Mutations run, one at a time; all six killed.**

| mutation | killed by |
|---|---|
| advance `last_seq` before the `if not verdict.write` guard | both backends contract test + unit |
| drop the `closed_by` update on an in-batch close | Django contract + unit |
| drop the in-batch producer-state advance | Django contract + unit |
| restore Djangos `if stream.closed: return [...]` | Django contract test |
| skip the shared call in the in-memory pre-flight | in-memory `validates_content_type_per_item` |
| `states = producer_states` instead of `dict(...)` | the purity unit test |

Two are worth flagging: mutations 2 and 3 leave the **in-memory** contract test
green. That is not a coverage hole, it is the consequence of that store being a
literal loop of `append` — it re-reads committed state per item, so it cannot
observe an in-batch advancement bug. The unit test and the Django contract test
kill both. Worth knowing before trusting an in-memory-only green on this path.

Gate: `ruff check` / `ruff format --check` (src tests examples, matching CI),
`pyright src/` 0 errors, `pytest` 1667 passed, `just test-pg` 1681 passed,
`zensical build` clean, `api-reference-check` clean (nothing here is in either
`__all__`), `conformance/expected-failures.txt` unmoved.

joshbrooks:
Review loop: four rounds, fresh reviewer each, converged.

Rounds 2 and 3 each found a **green mutation** — a fact the in-memory store hands
the shared rule that could be blanked with the whole suite green, and where the
resulting behaviour was a real divergence between the two backends. Both were the
same shape, and both were found by *deleting a whole fact* rather than by
mutating logic inside it:

- **`closed`** (`store.py:370`). Setting it to `False` left 1694 green. A batch
  carrying a `Stream-Seq` or content-type conflict against an already-closed
  stream then raised where the durable store returns a refusal per item — a 400
  on one backend and a clean "closed" on the other. The existing closed-stream
  batch test uses a *clean* batch, which such a store answers correctly. Pinned
  by `test_a_closed_stream_refuses_a_conflicting_batch_rather_than_raising`
  (`3c6fdf7`).
- **`last_seq`** (`store.py:373`). Setting it to `None` left 1696 green, and left
  a **written prefix** behind a refusal on the in-memory store where the durable
  store rolls back. Needs an admissible item in front of one that conflicts with
  the *pre-batch* value — the earlier sibling starts from a fresh stream, so its
  conflict is in-batch and a store that never reads the stream still catches it.
  Pinned by `test_a_batch_conflicting_with_the_stream_writes_nothing`
  (`c131d4b`).

Round 3 also flagged that deleting the in-memory pre-flight *entirely* left only
one test red — the content-type one — so the `Stream-Seq` all-or-nothing path was
covered only by a single-item batch, which cannot tell a refusal from a refusal
that happened to leave nothing behind. Closed by
`test_a_seq_conflict_late_in_a_batch_writes_nothing` (`3f785f8`). Deleting that
pre-flight now takes four tests down; deleting the durable store's takes fifteen.

Verified across rounds and worth recording:
- Committing the last fencing outcome per producer is equivalent to committing
  each accepted item in turn — `_commit_producer` is a full-overwrite
  `update_or_create`, checked against an epoch bump mid-batch, accepted-then-stale,
  and two interleaved producers.
- `append_many` ≡ a loop of `append` on both stores, by differential probe over
  several hundred random batches (two independent reviewers, ~300-400 each), on
  per-item outcome, final stream shape, and the persisted `StreamProducer` rows.
  The only difference is the deliberate one: on a conflict `append_many` writes
  nothing where a loop leaves a prefix.
- `closed_by` at `store.py:371` is inert at that call site — a closed stream never
  reaches a raise, and the in-memory store discards the verdicts. Left in place
  deliberately: the store's job is to gather the facts, and pruning the fact set
  to what today's rule happens to read is the drift this module exists to stop.
- No `select_for_update()` site changed and nothing moved across the
  `_locked_write` boundary, so no test was added to `test_concurrent_appends.py`
  or `test_locking.py`. Two reviewers agreed independently.

Round 4 found two divergences on a **different axis**, both pre-existing and
outside this diff: a batch containing an invalid JSON body is refused with a
written prefix on one backend and silently accepted on the other, and a
top-level JSON list flattens on one and not the other. Filed as #214 rather than
fixed here — the durable store's non-flatten is a written-down decision and the
in-memory behaviour is inherited rather than chosen, so picking one changes what
an existing caller gets.

Rebased onto `main` after #206 landed; the conflict was `store.py`'s import block
and both contested names turned out to be unused.

PR #212, opened 2026-08-21
One thing that knows whether a rule's code has changed

### LINKED ISSUE #187: Noticing that a rule's code has changed is spread across seven places
We warn when the code behind a stored rule has changed since it was registered, because a replay would then produce something different from what was recorded. Understanding that warning means reading three separate files, and three nearly identical checks that differ only by a word.

Worse, the shared entry point takes two separate options that have to agree with each other. Pass one and forget the other and you silently lose either the warn-once behaviour or the caching — and the correct pairing is only assembled in one place deep inside the machinery. One object handed the job would fix both halves.

joshbrooks:
We warn when the code behind a stored rule has changed since it was registered, because a replay would then produce something different from what was recorded. That warning was answered by three near-identical checks — one each for handlers, reducers and upcasters — differing only in a word, and the reducer one had no test at all: deleting it outright left all 1648 tests green. Worse, the shared entry point took two options that had to agree, so passing one and forgetting the other silently lost either the warn-once behaviour or the caching.

Now one object owns the question: whether this rule's code has drifted, whether it has already said so, and what it has hashed. There is one check, reached three ways, and one option to pass instead of two that can disagree — the mis-pairing is no longer expressible. Normalising a single event on read can be handed the same object, which is the path that previously had no way to ask.

Closes #187.


joshbrooks:
Technical detail.

**Shape.** New `rakaia/drift.py` holds `OnDriftPolicy`, `HandlerDriftError` (moved next to the rule that raises it, re-exported from `rakaia.registry` and `rakaia.replay` so every existing import path still works) and `DriftLedger`. The ledger owns four things that were previously spread across `_ReplayCtx`, two module functions and a closure: the policy, the `(kind, name, stored_hash)` reported-set, the per-callable hash memo, and the emitted `warnings` / `drifted` lists. `check(kind=, name=, stored_hash=, fn=)` is the only entry point; `kind` is a word in the message, not a branch.

**Sites collapsed.** `_check_handler_drift`, `_check_reducer_drift`, `_record_drift`, `live_source_hash` and the `_upcaster_drift` closure in `build_pipeline` are all gone (about 70 lines), replaced by three `ctx.drift.check(...)` calls and one `drift.check(...)` in `UpcasterRegistry.apply_chain`. `apply_chain` / `upcast_to_current` lose `drift_callback` and `hasher` for a single `drift: DriftLedger | None`. `ReplayResult.warnings` and `.drift_detected` are now properties reading the ledger, so the result and the log cannot disagree; `ReplayResult` gains a `drift` field. `upcast()` gains `drift=` (still silent by default — a one-off read has nowhere to report to, and changing that would be a semantics decision, not a move).

**The uncovered path.** Before this change, deleting the reducer drift check entirely left the whole suite green — `grep -ri drift tests/ | grep reducer` was empty. Handler and upcaster drift were covered; the third variant was not. Same shape as #193 and #195.

**Mutations.** Each mechanism broken one at a time, whole suite run per mutant, all caught: handler check removed (10 failed), reducer check removed (3), upcaster check removed (18), warn-once memory removed (9), hash memo removed (2), dedup keyed by name instead of by registration (2), `on_drift="raise"` ignored (7), ledger not handed to the upcaster chain (7). No green mutants.

**Tests.** `tests/test_rakaia/test_drift_ledger.py`. The duplication is pinned by parametrising over the three kinds and reaching each through `replay()` — not through the shared helper, since a test of the helper cannot see a call site that forgot to use it — asserting the warning string, the reported names and the raise behaviour by exact equality. The mis-pairing is pinned two ways: the signature carries exactly one keyword-only drift option (a second one added later fails and asks why), and a ten-event replay through a drifted upcaster yields exactly one warning *and* exactly one hashing of its source, so both halves are observable from the one default.

Gate: `ruff check`, `ruff format --check` (src/ tests/ examples/, CI's scope), `pyright src/` 0 errors, 1666 passed / 16 skipped / 1 xfailed, `zensical build` clean, `just api-reference` regenerated and committed.

PR #213, opened 2026-08-21
Work out a read's answer, then send it

### LINKED ISSUE #158: Response header rules can only be checked by making a real HTTP request
The code that answers a read request is about 190 lines with eight different exit points, and it sends its response directly rather than working out an answer and handing it back. That means no rule about what headers a response carries can be checked without standing up a server and making a request, and the setup for doing that is currently repeated across several test files.

This one is a shape problem rather than a defect: unlike the rest of this epic there is no failing test, because the duplicated pieces have not actually drifted apart. It is included because the same file already does this well elsewhere — one part was pulled out into a plain function that works out an answer, and it is the only part with clear, direct tests.

### LINKED ISSUE #188: What a read puts in its response headers can only be checked by making a real request
Reading from a stream decides half a dozen things — whether you are at the end, whether you are up to date, whether the stream is closed, what to cache against — and every one of them is decided in the middle of a very long function with eight different exit points. So the only way to check any of them is to make a full HTTP request and inspect the bytes.

The same six things are then worked out a second time, independently, for the live-push form of the same read. Two answers to one question. The write side of the server already has a separate module that decides and returns rather than deciding and sending; the read side never got one.

joshbrooks:
Reading from a stream decides about half a dozen things — whether you are at the end, whether you are up to date, whether the stream is closed, what to cache against — and every one of them was decided part-way through a very long function with eight different exit points that sent its response rather than working out an answer. So the only way to check any of it was to make a full HTTP request and inspect the bytes, and the setup for doing that was repeated across three test files. The live push then worked all six out a second time, independently: two answers to one question.

There is now one place that answers them, sitting next to the equivalent module the write side already had, and both forms of read go through it. Nothing a client sees changes. What changes is that the rules can be asked a question — 78 new tests do exactly that, and along the way it turned out nothing in the suite ever checked that a conditional read actually comes back "not modified", so that is checked now too.

Closes #188. Closes #158.


joshbrooks:
**Shape.** `src/rakaia/read_decision.py`, mirroring `append_decision.py` deliberately closely: `ReadFacts` (the `StreamFacts` equivalent — the fields a response is built from, no store attached), `ReadRequest`, `ReadVerdict` (`AppendVerdict`), and `decide_*` entry points per exit. `handler.py` keeps only the I/O: `_facts()` narrows a store stream, `_send_verdict()` sends. `_handle_read` drops from ~190 lines to ~144 with one send per exit; `handler.py` 1295 → 1237 lines.

Public names (all internal — `rakaia.__all__` unchanged, `just api-reference-check` clean): `closed_at_tail`, `stream_ended`, `read_etag`, `read_param_error`, `decide_head`, `decide_catch_up`, `decide_wait_ended`, `decide_read`, `sse_control_fields`. No clock and no randomness inside — response cursors are generated by the caller and handed in, so every verdict is a pure function of its arguments.

**Did the two reads disagree?** Yes, on one rule, latently. `_handle_read` required `up_to_date` before reporting closure (`closed and at_tail and up_to_date`); `_build_sse_control` did not (`closed and at_tail`). Not reachable today — `StreamStore.read` and `DjangoStreamStore.read` both hardcode `up_to_date=True` — so no defect is claimed. Unified on the plain reads rule via `closed_at_tail`. That forced a related change: `_closed_control` was passing `up_to_date=False` on the grounds that the cursor arguments were never read, which under the shared rule would have emitted a cursor instead of the terminal `streamClosed`. It now passes `True`, which is what the terminal frame means, and the bytes are unchanged.

`stream_ended` is deliberately *not* the same rule. Ending the live push is a liveness question: a closed stream at its tail can never produce another message, so folding it into `closed_at_tail` would let a store reporting `up_to_date=False` loop forever. Both are named and tested side by side.

**Confirmed non-findings.** #158 suspected the ETag rule had drifted between GET and HEAD. It had not, and `test_head_and_a_full_get_at_the_tail_agree` now pins that. The `a and b or c and d` long-poll predicate parses as intended and is unchanged.

**Preserved quirks, now tested rather than latent.** A `304` carries only its ETag and no CORS headers. The immediate closed-at-tail `204` carries no `Stream-Cursor` where the two post-wait exits do. Both are asserted with exact-equality assertions so they cannot silently change.

**Tests.** `tests/test_rakaia/test_read_decision.py`, 78 cases, 0.17s, no HTTP. The `httpx.ASGITransport` incantation moves to `tests/asgi_client.py` — 5 duplicated setups across three files collapse to one helper. No coverage deleted.

**Mutation: 27 mutants, 27 caught**, gated on pytest exit code *and* a `N failed` summary line so a collection error is never mistaken for a caught mutant. 10 of the 27 delete a decision outright rather than flipping a condition, which is what found the two below.

**Two green mutants found and fixed** — both instances of the self-covering-common-path trap from #193:

1. *The handler never re-fetches the stream.* `_handle_read` fetches, waits, then fetches again before deciding. Replacing the second fetch with the first left all 1648 tests green, because `StreamStore.get` hands back the live object — every core test covered itself. `DjangoStreamStore.get` returns a detached row, so the fix is a durable-store test (`test_a_close_landing_during_a_long_poll_is_reported` in `test_protocol_server.py`): long-poll parked at the tail, close lands during the wait, response must carry `Stream-Closed` and a `:c` ETag. The in-memory twin is kept and its docstring says outright that it passes either way.
2. *No end-to-end cover for the conditional read.* Nothing in the suite ever sent `If-None-Match` — the only mention was a CORS preflight asserting the header is *allowed*. Deleting the 304 branch was green. Two HTTP tests added.

**Conformance baseline unchanged:** 276 passed / 56 failed / 6 skipped — 0 new failures, 0 newly passing, 56 expected.

PR #219, opened 2026-08-21
Cover the read wiring, not just the read rules

joshbrooks:
The change that moved every read decision into one place landed with direct tests for each rule. Reviewing it afterwards showed the rules were well covered but the code that *uses* them was not: the cross-origin header could be dropped from every kind of read except one, a long poll could stop sending the value that lets a CDN collapse requests, the cache key could forget which stream and which position it belongs to, and a live connection could send its final message twice — each of those with every test still passing.

All of it is tests. No behaviour changes. The one substantive addition is the case where a stream is closed without any accompanying data while a client is parked waiting: that has to reach the client, and it sits one line away from the case that was already covered.

Follows #213.

### DIFF
--- a/tests/test_rakaia/test_handler.py
+++ b/tests/test_rakaia/test_handler.py
@@ -411,6 +411,68 @@ class TestConformanceGaps:
         assert again.content == b""
         assert again.headers["etag"] == etag
 
+    async def test_a_read_carries_the_cors_header(self, client: httpx.AsyncClient):
+        """Every read response is cross-origin readable, not just the preflight.
+
+        The preflight above and `HEAD` were the only cover; the four read exits
+        that build their own header set could each drop CORS unnoticed, which a
+        browser would see and no test would.
+        """
+        await client.put("/foo", headers={"content-type": "text/plain"}, content=b"one")
+        for url in ("/foo", "/foo?offset=now"):
+            response = await client.get(url)
+            assert response.headers["access-control-allow-origin"] == "*", url
+
+    async def test_two_streams_at_the_same_offset_cache_separately(
+        self, client: httpx.AsyncClient
+    ):
+        """The ETag names the stream, so one stream's tag cannot match another's.
+
+        Both streams here sit at the same offset with the same content, so a tag
+        that forgot the path would match across them and serve a 304 for a body
+        the client has never seen.
+        """
+        for path in ("/a", "/b"):
+            await client.put(
+                path, headers={"content-type": "text/plain"}, content=b"one"
+            )
+
+        etag_a = (await client.get("/a")).headers["etag"]
+        crossed = await client.get("/b", headers={"if-none-match": etag_a})
+        assert crossed.status_code == 200
+        assert crossed.headers["etag"] != etag_a
+
+    async def test_the_etag_names_the_offset_the_client_asked_for(self):
+        """`?offset=now` caches under `now`, not under the tail it resolved to.
+
+        `now` means "wherever the stream is when you get this", so two clients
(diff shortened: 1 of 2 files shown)


joshbrooks:
Detail. All 13 gaps were found as **green mutations** — the mechanism broken on purpose, the whole suite still passing. The harness gates on pytest's exit code *and* a `N failed` summary line, so a mutation that fails to parse (a collection error, which never runs the test) is never mistaken for a caught one.

Round 2 of review (six findings, all wiring):

1. **CORS dropped from three of the four read exits.** `{**(cors or {})} → {}` in `decide_read` / `decide_catch_up` / `decide_wait_ended`, and dropping `cors=cors` at three call sites: green every time. Only `decide_head`'s merge was covered. Browser-visible. Fixed by `test_a_read_carries_the_cors_header` plus assertions on the two long-poll 204s.
2. **The wait's own closed signal.** `closed = True → False` on the `stream_closed` arm of the merged long-poll exit: green. Reachable — a close-only `POST` (empty body + `Stream-Closed`) while a client is parked. Fixed by `test_a_close_with_no_data_ends_a_parked_long_poll`. This is the direct sibling of the arm covered in #213's last round, one line away.
3. **The long-poll `Stream-Cursor`.** Two independent mutations green: `live=live → None` and dropping the whole `response_cursor=` expression. The rule was covered; the wiring was not.
4. **`offset` as sent, not as resolved.** `offset=offset → effective_offset`: green, despite `ReadRequest`'s docstring saying normalising `now` here would change what clients cache against. Reachable via `?offset=now&live=long-poll` with an append landing during the wait.
5. **The ETag's path component.** `path=path → "/wrong"`: green. With a constant path an `If-None-Match` from `/a` returns 304 on `/b`.
6. Four minors: the live push's stop at a closed tail (a missing stop sends a duplicate terminal frame, which a data→control pairing check cannot see — the frame sequence is now asserted as the exact list); the pre-wait 204's deliberately absent cursor at its *call site*; `offset_count` reaching the duplicate-parameter check; and the SSE `upToDate` field.

Round 3 found one more: the order of the first two parameter refusals (`?offset=&offset=x`) was unpinned while the two adjacent orderings were.

Round 3 then reported nothing blocking. It independently confirmed all 12 round-2 mutations caught by the test named for each, ran the seven concurrency-dependent tests 25x with no flakes, and checked that the two new ETag tests fail for the reason their names claim.

**Accepted as unreachable, not fixed** (recorded so a later review does not re-raise them): three green equivalence-class mutations — the no-messages `response_offset` fallback, passing `latest` as `facts` at the `decide_read` call site, and `effective_offset or stream.current_offset` at the long-poll wait. The first two need an append to land inside `_handle_read` with `messages` empty; the third is dead because long-poll without an offset is already a 400. Also `up_to_date=False` anywhere, since both stores hard-code `True`. Two further greens (`store.touch` on the catch-up path, and the `offset == "now"` SSE resolution) are byte-identical to `f27629b` and so pre-existing.

**Also observed, pre-existing and not addressed here:** `docs/protocol.md:687` requires the already-closed-at-connect SSE control event to carry both `streamClosed: true` and `upToDate: true`, but `sse_control_fields` makes the two mutually exclusive — as the code it replaced did. Worth its own issue rather than a silent change.

Gate: `pytest` 1841 passed / 16 skipped / 1 xfailed · `ruff check` + `ruff format --check` over `src/ tests/ examples/` · `pyright src/` 0 errors · `zensical build` clean · `just api-reference-check` clean (no public names added) · conformance baseline unchanged (276 passed / 56 failed / 6 skipped, 0 new failures, 0 newly passing).

PR #220, opened 2026-08-22
The event filter and page header stop showing an internal label

### LINKED ISSUE #210: The admin filter and page title still show the placeholder for a labelless append
The badges in the admin now say "no label" for an event that carried none, but two smaller places on the same screens still print the internal placeholder: the filter list down the side of the event and entry lists, and the title an event gets on its own page. So a reader can see an em-dash in the table and the word "append" in the filter beside it, for the same event.

Found while fixing the badges. Neither is a wrong answer to a question anyone asks deliberately, which is why it was left out of that fix, but it is the same disagreement in the same screen.

joshbrooks:
Two small places on the stream admin screens still printed the internal placeholder for an event that carried no label, so the same event read as a dash in the table and as "append" in the filter next to it. The filter beside the badges and the header above the change form now both say what every other reader of that event says.

The filter still puts the stored value in the address bar and matches on it — only the text of the link changed — so an existing bookmark selects exactly the rows it used to.

Closes #210.

### DIFF
--- a/src/django_rakaia/admin.py
+++ b/src/django_rakaia/admin.py
@@ -12,7 +12,11 @@ from django.contrib import admin
 from django.utils.html import format_html, format_html_join
 from django.utils.safestring import SafeString, mark_safe
 
-from django_rakaia.event_message import decode_payload, event_label
+from django_rakaia.event_message import (
+    decode_payload,
+    event_label,
+    event_label_display,
+)
 from django_rakaia.models import Stream, StreamEntry, StreamEvent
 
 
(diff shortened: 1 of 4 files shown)


joshbrooks:
Shape: `EventTypeFieldListFilter(admin.AllValuesFieldListFilter)` in `src/django_rakaia/admin.py`, overriding only `choices()`. It relabels by position — Django yields "All", then one choice per non-null value in `lookup_choices` order, then the empty-value choice — so the query string, the `selected` flag and the ` (N)` facet suffix are all passed through as Django built them. The three `list_filter` sites use the `("event_type", EventTypeFieldListFilter)` tuple form, so the field path is unchanged. The placeholder itself is `event_label_display` / `NO_LABEL_DISPLAY` in `event_message.py`, next to the `event_label` rule it inverts, and the badge now goes through it too.

The first attempt rebuilt the filter as a `SimpleListFilter`. Review found two defects that shape had to reimplement and got wrong: `?event_type=a&event_type=b` selected only `b` (Django ORs repeated values; `SimpleListFilter` keeps `value[-1]`), and the entries sidebar built its choices with a `DISTINCT` over the fan-out `StreamEntry` table instead of reading `StreamEvent`. `TestRelabellingTheFilterChangedNothingElseAboutIt` pins all three properties and each of those tests goes red against the earlier shape.

Mutations, all killed, gated on pytest exit code: deleting the `choices()` override (6 tests red), reverting each of the three `list_filter` entries independently (3 / 2 / 1), reverting `StreamEvent.__str__` to the raw column (2), blanking `event_label_display` (10), dropping the facet-count suffix (1), and an off-by-one that relabels the "All" choice (6). The one new test no mutation kills is the querystring-equality control, which is a companion assertion rather than the cover.

Known and left alone: a stored `event_type` of `""` or of literally `"—"` displays the same as the sentinel, so two stored types would share a sidebar label (their query strings still differ). Neither value is reachable through the store, and the badges from #208 already behave this way.

PR #221, opened 2026-08-22
Name the thing the two replays differ about

### LINKED ISSUE #189: The replay tests reach inside replay in order to work
Some helpers were pulled out of replay so that the single-stream and merged-stream versions could share their middle. The tests now import those helpers directly, which means the test suite is checking things through a side door rather than through replay itself — usually a sign the pieces are cut in the wrong place.

It matches where the awkwardness actually is: one version keeps a second hand-written loop because reading events lazily matters for how far it gets before failing, and the other always reads everything up front and does not say so. Naming the thing they differ about would let both share one loop.

joshbrooks:
Replay had two versions of its middle. The single-stream one kept a second
hand-written loop because it reads events one at a time, and the merged one
always read everything before starting and never said so. Since neither
behaviour had a name, the tests could only get at them by importing the helpers
directly — checking the machinery through a side door instead of through replay.

The difference now has a name and one loop serves both: a replay reads ahead
only as far as it has to. Nothing a caller does behaves differently; what
changed is that the rule is written down — a plain replay applies the events
before a bad one, while a staged replay or a merge applies nothing, because both
have to see everything first. The tests go through the two entry points, each
shared behaviour run against both.

Closes #189.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,28 @@ runnable demo for each.
 
 ### Changed
 
+- **A replay now says how far it gets before failing.** `replay` and
+  `merge_replay` share one staging pipeline, but `replay` kept a *second*
+  hand-written loop for its single-stage path, and the reason was real: it
+  decodes one event at a time, so a malformed event at offset N fails with the
+  first N already applied. `merge_replay` always read everything up front and
+  never said so. Nothing tested the difference, so the behaviour the second loop
+  existed to protect was uncovered.
+
+  That difference now has a name — `EventSource`, the events a replay will
+  dispatch, deliberately an `Iterable` rather than a `Sequence` — and one loop
+  serves both: `run_passes` consumes its source as it arrives when one pass will
+  do, and materialises it when it needs more than one. **No caller sees anything
+  different.** What changed is that the rule is written down: a single unstaged
+  pass applies the events before a malformed one; a staged replay or a merge
+  applies nothing, because more than one pass has to see every event and a merge
+  cannot know the order until every stream is read.
+
+  The pipeline tests used to import `build_pipeline` and `run_passes` and drive
+  them on a fabricated list of events. They now go through `replay` and
+  `merge_replay`, each shared behaviour run against both, since there is no
+  longer anything behind them worth reaching past (#189).
+
 - **A single append reads the stream row once, not twice.** Every
   `DjangoStreamStore.append` did two `SELECT`s of the same row: an unlocked one
   before the transaction opened, purely so that an expired stream would still be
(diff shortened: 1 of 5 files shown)


joshbrooks:
Technical detail.

**What was named.** `rakaia.replay.EventSource = Iterable[PipelineEvent]` — every event a replay will dispatch, in dispatch order. Deliberately `Iterable`, not `Sequence`: how much of the range is decoded before the first handler runs is observable, and it is the one thing the two entry points differ about. `run_passes` now decides from `is_staged(ctx)` alone — `source = list(events) if staged else events` — so a single pass consumes the source as it arrives and more than one materialises it first. `replay()`'s second loop is gone; it yields a generator. `merge_replay()` hands over a list, and the comment there now says that is forced rather than chosen (the sort has already read every stream).

`events_processed` moved from `len(events)` to a per-pass count, so it still counts each event once however many stages ran, and works for a source with no length.

**Defect or merely unstated?** Merely unstated. Both eager cases are eager because they have to be: a second pass needs the whole range, and a merge cannot know the order until every stream is read. Nothing observable changed — the fix is that the rule is now written in the module docstring, both entry-point docstrings, `docs/staged-replay.md` and `docs/multi-stream-merge.md`.

**Internal imports.** `from rakaia.replay import build_pipeline, run_passes` is gone from `tests/test_rakaia/test_replay_pipeline.py`. The two remaining references to the name `build_pipeline` in the suite (`test_public_api.py`, `test_architecture_spikes.py`) are about the function-shadows-module sharp edge, not the pipeline, and are untouched. Nothing moved between modules, so that trap is unaffected.

**Self-covering pair.** Every shared behaviour is parametrised over an `Entry` describing one entry point and its seeding, and each mutation was run against the `replay` half and the `merge_replay` half *separately*. All seven shared-pipeline mutations went red on both halves independently: deleting the reader-required call, blanking the reader-required rule, collapsing the stage passes to one, running every pass at stage `None`, deleting the `events_processed` count, deleting the per-stage reducer run, and dispatching only the first event of each pass. Plus, whole-file: making the source always lazy (red), always materialised (red), and deleting each of `replay()`'s two range bounds (red against `test_replay.py`). No green mutations.

**The gap this found.** With `source = list(events)` unconditionally — i.e. the laziness `replay()`'s second loop existed to protect, removed — the entire suite minus the new file passes: 1863 passed, 16 skipped, 1 xfailed. The behaviour that justified the duplicated loop had no test anywhere. `TestHowFarAReplayGetsBeforeFailing` is now the only thing holding it, and it pins all three shapes (single pass applies what came before; staged applies nothing; merge applies nothing even unstaged).

Gate: `just lint`, `just fmt-check`, `just typecheck` (0 errors), `uv run pytest` (1890 passed, 16 skipped, 1 xfailed), `uv run zensical build`, `just api-reference-check` — all green. No exported name changed, so `docs/api-reference.md` is unmoved.

PR #222, opened 2026-08-22
Decide a batch's bodies before writing any of them

### LINKED ISSUE #214: A batch write still disagrees about what a payload is
The rule for whether a *batch* is allowed is now shared between the two storage backends (#181). What each backend does with the batch's *contents* is not, and the two answer differently in two ways.

Give the same batch to both: a valid body followed by a body that is not valid JSON, on a stream declared as JSON. One backend refuses it and keeps the first body, so a caller who has been told the batch failed still finds half of it written. The other accepts both, storing something that is not JSON in a JSON stream. Separately, a body that is a JSON list becomes several entries on one backend and a single entry on the other. Neither of these is new, and neither is what #181 was about, but they are the same problem one layer in: a caller has to know which backend it is talking to.

joshbrooks:
A batch of appends checked its options before writing anything but never its bodies, so a valid message followed by one that was not JSON went two different ways: the in-memory store refused the batch and left the first message written anyway, and the durable store did not refuse at all and stored a body that was not JSON in a stream declared as JSON. Body validity is now decided beside the shared batch rule, per item, before anything is written, and both stores are made to hand their bodies over.

The other half was whether a batch item holding a JSON list becomes one message or several, which the two stores also answered differently. That is now decided in favour of several, matching what a single append already does on both: the durable store splits a body through the same helper its own `append` uses. An item can write more than one row, and still returns exactly one result, carrying the last message it became. The tests that recorded the divergence without judging it are one assertion again. Closes #214.

### DIFF
--- a/src/django_rakaia/django_store.py
+++ b/src/django_rakaia/django_store.py
@@ -882,8 +882,10 @@ class DjangoStreamStore:
         The rest of :meth:`append`'s semantics hold per item too, because the
         same ``decide_append`` rule decides every item: a closed stream refuses
         every item with ``stream_closed=True``; per-item ``content_type`` and
-        ``seq`` are validated (the whole batch, before any row is written — a
-        conflict raises and writes nothing); **producer fencing applies per
+        ``seq`` are validated, and so is every admitted item's *body* against
+        the stream's content type (the whole batch, before any row is written —
+        a refusal raises and writes nothing, so a JSON-mode stream can no longer
+        take a body that is not JSON, #214); **producer fencing applies per
         item**, with an item's outcome returned in its own slot rather than
         aborting its siblings; and an item with ``close=True`` closes the
         stream, refusing the items after it, exactly as a loop of ``append``
@@ -921,6 +923,7 @@ class DjangoStreamStore:
             batch = decide_append_batch(
                 self._facts(stream),
                 [options for _data, options in items],
+                payloads=[data for data, _options in items],
                 producer_states={
                     pid: self._producer_state(stream, pid) for pid in producer_ids
                 },
--- a/src/django_rakaia/django_store.py
+++ b/src/django_rakaia/django_store.py
@@ -877,7 +877,11 @@ class DjangoStreamStore:
         that returns ``[]`` without touching the database (so it never raises
         for a missing stream). A non-empty batch raises ``StreamNotFound`` if
         the stream does not exist, like :meth:`append`, and returns one
-        ``AppendResult`` per item in input order.
+        ``AppendResult`` per item in input order — one per *item*, not per row:
+        in JSON mode an item whose body is a top-level array flattens into one
+        row per element, exactly as :meth:`append` flattens it, so a batch of N
+        items can write more than N rows and that item's result carries the last
+        of them (#214).
 
         The rest of :meth:`append`'s semantics hold per item too, because the
         same ``decide_append`` rule decides every item: a closed stream refuses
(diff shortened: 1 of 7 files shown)


joshbrooks:
## The flattening half: recommendation

**Make the durable `append_many` flatten, matching `append` on both backends.** Not "move the rule to HTTP", and not "make the in-memory store stop".

**1. Does conformance pin store-level flattening? No — and this is the decisive fact.** The suite is HTTP-only (`conformance/conformance.test.mjs` does nothing but `runConformanceTests({ baseUrl })`) and runs only against the in-memory store (`conformance/run.sh:29` launches `uv run uvicorn rakaia:app`, and `src/rakaia/__init__.py:146` is `app = create_app()`). More importantly, **`append_many` has no HTTP surface at all** — `src/rakaia/handler.py:1104-1108` is the only body-writing call site and it uses `append`/`append_with_producer`; `grep -rn append_many src/` finds no handler. So the conformance suite cannot see this divergence in either direction, and the decision is free of it. Baseline confirmed unmoved on this branch: 276 passed, 56 failed, 6 skipped, 0 new failures (the 56 are the `Fork - *` family in `expected-failures.txt`, #61).

The suite's flatten tests are POST-body tests despite their titles. `should store arrays as single messages` calls `stream.append(JSON.stringify([{id:1},{id:2},{id:3}]))` and expects `[[{id:1},{id:2},{id:3}]]` back — because the JS client wraps every JSON append in one extra array level (`@durable-streams/client/dist/index.js:2712-2716`, `encodedBody = `[${bodyStr}]``). The wire body is `[[...]]` and the expectation proves the server flattened exactly one level of it.

The only store-level pin is `tests/server_store_contract.py::test_a_json_array_append_flattens_into_separate_messages`, which uses `append` (singular) and holds on **both** backends. Nothing pins `append_many` — `tests/store_contract.py::test_append_many_matches_append_loop` looks like it would, but its batch is three plain objects.

**2. POST body or store append?** Every normative sentence in §7.1 is phrased about HTTP. `docs/protocol.md:789`: *"When a **POST request body** contains a JSON array, servers MUST flatten exactly one level… This enables clients to batch multiple messages **in a single HTTP request**."* Line 791 labels its examples *"(direct POST to server)"*; line 802's empty-array rule is *"MUST reject POST requests…"*. Strictly, neither store is obliged. But the repo has already chosen where the rule lives, and ADR 0002 is explicit: `docs/adr/0002-framework-vs-protocol-server-boundary.md:136-138` puts *"JSON mode"* inside *"a **shared conformance suite** every store must pass"*, and line 102 names `json_mode` as one of the four tier-independent shared-vocabulary modules. So **candidate (a), moving flattening to the HTTP layer, contradicts a decided ADR** and would delete a store-contract test that both backends currently pass.

That leaves the #211 identity: `append_many` ≡ a loop of `append`. `append` flattens on both backends. Candidate (c) — stopping the in-memory flatten — would break that identity *within a single backend*, and would need `append_many` to stop being a loop of `append`. The durable store is the one out of step.

**3. What breaks for an event-sourcing caller whose payload is a list?** Nothing that works today. No caller appends a list (`grep -rn 'json.dumps(\[' src/ examples/` is empty), and **no example declares `application/json` at all** — every one is `store.create(path)` with no content type, where neither backend flattens and `encode_payload` stores the parsed value (`src/django_rakaia/event_message.py:63-67`, `src/django_rakaia/models.py:314-317`). And a list payload is already broken downstream regardless of this decision: `src/rakaia/replay.py:618` `_decode_event` is declared `-> dict` and unchecked, so the list reaches `upcast_to_current` and dies at `src/rakaia/registry.py:956` `event.get("schema_version", 1)` with `AttributeError: 'list' object has no attribute 'get'`. `src/rakaia/json_mode.py:62` is the only `isinstance(..., list)` in either package. So the case the durable store's comment protects cannot currently be exercised, and callers who genuinely need a list as one event already have the right tool: a stream with no declared content type, which is what all of them already use.

**What making the durable store flatten would break.** `append_many` returns one `AppendResult` per input item, so a batch of N items could store more than N messages, with `AppendResult.message` being the last message of a flattened item — which is exactly what the in-memory store does today, so it is a parity gain, not a new wart. Any durable caller that appends array payloads to a stream it declared `application/json` would see its stored shape change; none exists in this repo. `tests/store_contract.py::test_append_many_matches_append_loop` should gain an array item once the decision lands.

## Payload validity: what changed

`check_payload` in `src/rakaia/append_decision.py` answers "can this stream's content type hold this body", and `decide_append_batch` calls it **inside** the per-item loop, after that item's admission verdict and only when the item is to be written. The placement is the load-bearing part: a pass over the payloads before the scan would tell a producer to fix a body whose write can never land, and one after it would be too late to be all-or-nothing. `payloads` is a required keyword rather than an optional one so a store cannot get a verdict without having its bodies checked.

The in-memory `append_many` is **still a plain loop of `append`** — the pre-flight discards its result, exactly as it already discarded the verdicts, and the loop re-parses each body on its way to storing it. No structural change there.

Nine mutations run, each killed by the test named for it: deleting the `check_payload` call in the batch loop; blanking `check_payload`'s body; deleting its JSON-mode guard; moving the call above the `verdict.write` test (killed by `test_a_closed_stream_answers_closed_rather_than_raising_on_a_bad_body`, `test_a_bad_body_after_a_close_in_the_batch_is_not_parsed_either`, `test_a_body_the_fence_refuses_is_not_read`); blanking each store's `payloads` argument separately; dropping `strict=` from the items/payloads zip; and flipping each backend's `append_many_flattens_json_arrays` characterising flag. Postgres leg: 1917 passed. Conformance baseline unmoved.

joshbrooks:
Correction to claim (a) in the comment above: the conformance client's extra array wrap for a plain `stream.append()` comes from the batching path, not `#appendDirect`. Batching is on by default (`@durable-streams/client/dist/index.js:2489`, `this.#batchingEnabled = opts.batching !== false`), so the call routes through `#sendBatch` at 2782-2786, which for a one-item batch produces the same `[bodyStr]` wrap that 2712-2716 does. The conclusion stands — the suite's "should store arrays as single messages" cases are one-level-flatten tests on a POST body — only the line reference was the wrong branch.

Issue #207, opened 2026-08-21
replay calls the executor once per event, so apply() never sees a batch it can collapse


joshbrooks:
`replay` dispatches one event at a time and calls the executor once per event, so `DjangoExecutor.apply` almost always receives a single effect. Two consequences follow. The batching added in #203 can never engage on this path — a run of one is never collapsed — and because `apply` opens its own `transaction.atomic()`, a submission with nine events pays nine savepoint/release pairs, which on our measurements is 18 of the 135 statements a form save issues.

Applying a whole stage's effects in one call would fix both, but it changes when a handler's writes become visible, so it needs care rather than a straight move. Numbers, the exact call pattern, and the constraint I think it turns on are in the comment below.

joshbrooks:
## The call pattern

Instrumented `DjangoExecutor.apply` during one live form save, rakaia main (`f27629b`), Postgres:

| form | events | `apply()` calls | effects per call |
|---|---|---|---|
| SF 2.3 | 9 | **9** | `[1,1,1,1,1,1,1,1,1]` |
| CFM 12/FF 12 | 4 | **4** | `[1,1,1,1]` |
| TF 6.1.1 | 3 | 3 | `[5,1,1]` |

So on the two multi-row forms every call carries exactly one effect.

## What it costs

One SF 2.3 save is 135 statements. Attributing `SAVEPOINT` by stack:

```
  9x  effect_executor.py:204 apply      <- one atomic per apply() call
  2x  django_store.py:343 _locked_write
  1x  (import_service)
  1x  (import_service)
```

9 `SAVEPOINT` + 9 `RELEASE` = **18 of 135 statements (13%)**, purely from `apply` being entered nine times instead of once.

And #203 cannot help here. `_flush_run` collapses a run of `Update`s, but a run only ever has one member, so the `len(run) == 1` branch always wins. We do see #203 working where we call `apply` ourselves with a full fan-out (a status change across nine rows went from 9 statements to 2) — it is specifically the replay-driven path that cannot benefit.

## Why not just batch per stage

Because it changes when a handler's writes become visible, and that is load-bearing. Constraints as I understand them:

1. **Batch within a stage, never across.** A stage-1 handler is called `fn(event, reader)` and reads what earlier stages committed. Deferring stage-0 effects past the end of stage 0 would break that. Batching the whole of one stage's per-event effects, flushed before the next stage begins, preserves it.

[...]

PR #223, opened 2026-08-22
A renamed effect field no longer loses references in silence

joshbrooks:
When one change to the database needs the identifier of a row another change is
about to create, it leaves a placeholder that gets filled in just before the
write. Finding those placeholders worked by trying three field names in turn and
asking each change whether it had one. Anything that did not answer to a name was
skipped without a word, so renaming a field — or adding a new kind of change
carrying a field the list had never heard of — left the placeholders unfilled and
wrote the placeholder object itself into the database.

Each kind of change now says which of its own fields can hold a placeholder, in a
form the type checker reads, so both cases fail the build instead of producing
wrong data. Two of the fields had never been covered by a test at all, which is
exactly why those two were the silent ones; they are covered now.

Closes item 2 of #161.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,24 @@ runnable demo for each.
 
 ## [Unreleased]
 
+### Fixed
+
+- **A renamed effect field no longer loses references in silence.** Substituting
+  a symbolic reference (`Ref`) used to work by trying three field names in turn —
+  `lookup`, `defaults`, `patch` — and asking each effect whether it had one. An
+  effect that did not answer to a name was skipped without a word, so renaming a
+  field, or adding a fifth kind of effect carrying a mapping the list had never
+  heard of, left references in it unresolved and wrote the placeholder object
+  itself to the database.
+
+  Each effect now names its own reference-bearing fields in code the type checker
+  reads, and the four kinds are matched exhaustively, so both cases are a
+  `just typecheck` failure instead of wrong data. Verified by mutation: renaming
+  `Retire.patch` produced **no** error inside the resolver before and produces two
+  now, and adding a fifth variant to `Effect` fails the exhaustiveness check.
+  `Retire.patch` and a `Delete`'s flat `Exclude` had no test coverage at all —
+  which is why those two renames were the silent ones — and now do. (#161 item 2)
+
 ### Added
 
 - **`DriftLedger` — one object that knows whether a rule's code has changed.**
--- a/src/rakaia/effects.py
+++ b/src/rakaia/effects.py
@@ -31,7 +31,7 @@ from collections.abc import Callable, Iterable
 from dataclasses import dataclass
 from dataclasses import field as dc_field
 from dataclasses import replace as dc_replace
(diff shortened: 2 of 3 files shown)


joshbrooks:
## What changed

`RefResolver.resolve_effect` walked `_MAPPING_FIELDS = ("lookup", "defaults", "patch")`
with `hasattr`/`getattr` and rebuilt via `dc_replace(effect, **changed)`, with the
signature `(effect: Any) -> Any`. It now dispatches on the four variants of the
`Effect` union, reads each field directly, and rebuilds with explicit keyword
arguments, generic in the effect type (`_EffectT`) so `_flush_updates` still gets an
`Update` back from an `Update`.

`typing.assert_never` is 3.11+ and this package is `requires-python = ">=3.10"` with
zero runtime dependencies, so the exhaustiveness check is written out as a local
`_assert_never(value: NoReturn)`. pyright treats it identically.

Ref-bearing fields per variant, which is now stated in code rather than guessed:

| variant | fields |
|---|---|
| `Upsert` | `lookup`, `defaults` |
| `Update` | `lookup`, `defaults` |
| `Delete` | `lookup`, `spare` (`Exclude.lookup` / `SpareKeys.keys[*]`) |
| `Retire` | `lookup`, `patch`, `spare` (`SpareKeys.keys[*]`) |

## Mutations, with a control

Both applied and watched fail. The first has a control, because "it errors now" is
only interesting against "it did not error before".

| mutation | before (`origin/main`) | after |
|---|---|---|
| rename `Retire.patch` → `soft_delete` | **0 errors in `effects.py`** | 2 errors, at the resolver itself |
| add a fifth variant to `Effect` | not applicable | 1 error at `_assert_never` |

The control is the point: on `main`, `pyright src/rakaia/effects.py` after that rename
reports `0 errors, 0 warnings`. Other modules (`executors.py`, `projections.py`,
`effect_executor.py`) did error, because they touch `.patch` directly — but the
resolver, the one place whose job is to know that field exists, said nothing. That is
the silent failure the issue describes, reproduced.

## Coverage gap this closes

There was no test placing a `Ref` in `Retire.patch` — the existing retire case
(`test_refs.py:124-134`) puts it in `spare` and leaves `patch` a literal — and none for
a `Delete`'s flat `Exclude`. Those are precisely the two renames that were silent;
`lookup` and `defaults` were covered and would have broken something. Six cases added
in `TestEveryRefBearingFieldIsReached`, one per (variant, ref-bearing field), plus the
identity fast path re-checked for all four variants since each now takes its own branch.

## Gate

`ruff check` / `ruff format --check` / `pyright src/` (0 errors) / `pytest` **1941 passed,
16 skipped, 1 xfailed** / `api-reference-check` clean — `__all__` is unchanged, the fix
is internal.

## Landing order

Please land this **before** #224 (#207). Both touch `src/rakaia/effects.py` — this one
in `RefResolver`, that one refactoring `check_disjoint_defaults` — and both add a
`CHANGELOG.md` `[Unreleased]` entry. Rebased on `f57953d`.

PR #224, opened 2026-08-22
A replay hands the executor a whole pass of changes at once

### LINKED ISSUE #207: replay calls the executor once per event, so apply() never sees a batch it can collapse

joshbrooks:
A replay used to apply one event's worth of changes at a time. That meant the
work we did earlier to collapse a fanned-out change into a single statement could
never do anything on the path that produces most of the changes, because there was
only ever one change to collapse. It also meant a form save with nine events
opened nine separate database transactions: eighteen of the hundred and
thirty-five statements that save issued were just the opening and closing of them.
On the same shape it is now one transaction and one statement.

Widening a batch is only safe if nothing downstream can tell, so the batch is
broken the moment holding one more change back would alter the outcome — a write
arriving behind a delete, a second write to the same column of the same row, a
repeated identifier, or a soft-delete that asked for notifications. Passes whose
handlers can read the database are left alone entirely and still apply event by
event, because a change held back would be invisible to them.

Closes #207.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,23 +9,6 @@ runnable demo for each.
 
 ## [Unreleased]
 
-### Fixed
-
-- **A renamed effect field no longer loses references in silence.** Substituting
-  a symbolic reference (`Ref`) used to work by trying three field names in turn —
-  `lookup`, `defaults`, `patch` — and asking each effect whether it had one. An
-  effect that did not answer to a name was skipped without a word, so renaming a
-  field, or adding a fifth kind of effect carrying a mapping the list had never
-  heard of, left references in it unresolved and wrote the placeholder object
-  itself to the database.
-
-  Each effect now names its own reference-bearing fields in code the type checker
-  reads, and the four kinds are matched exhaustively, so both cases are a
-  `just typecheck` failure instead of wrong data. Verified by mutation: renaming
-  `Retire.patch` produced **no** error inside the resolver before and produces two
-  now, and adding a fifth variant to `Effect` fails the exhaustiveness check.
-  `Retire.patch` and a `Delete`'s flat `Exclude` had no test coverage at all —
-  which is why those two renames were the silent ones — and now do. (#161 item 2)
 
 ### Added
 
@@ -158,6 +141,23 @@ runnable demo for each.
 
 ### Changed
 
+- **A replay now hands the executor a whole stage's changes at once.** It used to
+  apply one event's worth at a time, so the batching added for a fanned-out
+  change could never do anything on the path that produces most of the changes —
+  a run of one is never collapsed. On the reported form save, nine events became
+  nine separate database transactions: eighteen of the hundred and thirty-five
+  statements the save issued were the opening and closing of those transactions.
+  Measured on the same shape, it is now one transaction and one statement.
+
+  Widening a batch is only safe if nothing can tell, so the batch is broken the
+  moment holding one more change back would alter the outcome: a write arriving
+  behind a delete, a second write to the same column of the same row, a repeated
+  correlation id, or a soft-delete that asked for notifications. Passes whose
+  handlers are handed a reader are left alone entirely — they can see the
+  database, so they keep applying event by event. One deliberate widening: a
+  reference to a row created by an earlier event in the same pass now resolves,
+  where it used to fail. (#207)
+
 - **A replay now says how far it gets before failing.** `replay` and
   `merge_replay` share one staging pipeline, but `replay` kept a *second*
   hand-written loop for its single-stage path, and the reason was real: it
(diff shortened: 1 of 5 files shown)


joshbrooks:
## Measured, on the shape the issue reported

Nine events, each fanning one identical `Update` at its own row, replayed through
`DjangoExecutor`. `tests/test_django_rakaia/test_replay_batching_cost.py`.

| | before | after |
|---|---|---|
| `SAVEPOINT` / `RELEASE` pairs | 9 | **1** |
| `UPDATE` statements (`batch_updates=True`) | 9 | **1** (with an `IN` clause) |

Both verified against the control — with `ctx.buffer` forced to `None` the same two
assertions report `9 == 1`. So these are measurements, not tautologies.

## The design is the flush rule

Batching at all is easy; the work is in not changing behaviour. Four things force an
early flush, each mapped to an existing mechanism so each is testable:

1. **Collision.** `check_disjoint_defaults` runs per `apply()` and raises
   `EffectCollisionError` with nothing applied. Scoped to one event, "event 2
   supersedes event 1 on the same column" is ordinary; widened to a stage it raises.
   This is contract (`tests/executor_contract.py:428-441`), and it is **the hazard
   neither the issue nor its comment lists** — the one most likely to break a naive
   per-stage batch.
2. **Write-order rank.** A batch is applied writes → deletes → retires, so a later
   event's `Upsert` would hop above an earlier event's `Delete`.
3. **Repeated `produces=`.** `DuplicateProducesError` in one batch, normal across two.
4. **A `Retire` carrying a `Transition`.** Keeps `ReplayResult.external` in
   handler-emission order.

`check_disjoint_defaults` was refactored into a reusable `_WrittenFields` the buffer
consults, rather than a second copy of the collision rule in the orchestrator — the
concern #152 exists for. Its raising path keeps the same message and the same effect
indices.

**Only reader-free passes batch.** Stage 0 handlers are `fn(event)` with no reader, so
constraint 2 from the issue (a handler reading a sibling's write mid-stage) is
unreachable there. Stage > 0 handlers and reducers do get a reader, and both readers
go straight to storage — `DjangoProjectionReader` to the ORM, `InMemoryProjections` to
its own tables — so a buffered write would be invisible to them *silently*. Those
passes keep applying per event. No `flush()` was added to the `Executor` protocol: it
is `runtime_checkable` and `tests/executor_contract.py:126` asserts `isinstance`, so a
new required method would break every out-of-tree executor and four in-tree doubles.

## Mutations

Nine, each applied and watched fail. Named in the docstring of the test that catches it.

| # | mutation | test that went red |
|---|---|---|
| 1 | buffering off | 4 batching tests + both cost tests |
| 2 | batch every stage | `test_stage_one_still_applies_per_event` |
| 3 | drop collision arm | `test_a_second_write_to_the_same_column…` + row-equivalence |
| 4 | drop rank arm | `test_a_write_after_a_delete_starts_a_new_batch` |
| 5 | drop `produces` arm | `test_a_repeated_produces_id_starts_a_new_batch` |
| 6 | drop `_self_collides` | `test_an_events_own_colliding_effects_still_raise_on_their_own` |
| 7 | drop transition flush | `test_a_transitions_notification_still_precedes_a_later_events_own` |
| 8 | `_drain` out of `finally` | `test_a_single_pass_replay_applies_the_events_before_a_malformed_one` |
| 9 | `_drain` after reducers | `test_a_reducer_sees_the_stages_writes` |

**Mutation 7 is worth reading.** My first version of that test passed green with the
transition flush removed — the rank rule already covered the case I had written, so the
flush was dead code. The case it is actually for is a later event emitting *only* an
external effect: no database effect, so nothing else forces a flush, and the retire's
notification gets filed behind the later event's. The test was rewritten to that, and
then it goes red. A guard I had not watched fail was not yet a guard.

## Behaviour changes, deliberate

- **A `Ref` may now bind to a `produces=` row from an earlier event in the same pass.**
  It used to raise, because refs do not cross an `apply()` call. A widening, not a
  break, and pinned by `TestRefsWidenRatherThanBreak`. The executor-level contract that
  refs do not resolve across two *explicit* `apply()` calls is untouched.
- The executor sees fewer, larger batches. `effects_applied` is counted at buffer time
  so it reads the same either way; the flat list of effects is unchanged in content and
  order, which `TestBatchingChangesNoRow` checks against a one-event-at-a-time control
  across five shapes.

## Gate

`ruff` / `pyright src/` (0 errors) / `pytest` **1953 passed, 16 skipped, 1 xfailed** /
`just test-pg` **1967 passed, 2 skipped, 1 xfailed** / `api-reference-check` clean.
Rebased on `f57953d`.

## Landing order

Please land #223 (#161 item 2) **first** — both touch `src/rakaia/effects.py` and both
add a `CHANGELOG.md` `[Unreleased]` entry.

PR #225, opened 2026-08-22
Record why stream positions stay a counted number

joshbrooks:
The protocol allows a time-based identifier instead of the counted, padded number
we hand out, which would remove both the size limit we picked by hand and the
brief lock every write takes to get the next position. This is a decision record
that works the trade through and declines it for now: the lock is what makes a
position record the order writes actually arrived in, and a clock cannot give that
back once it is gone.

Two of the three original reasons to switch had already been dealt with by other
work, which is most of why the answer comes out this way, and the record says so
rather than leaving the old case standing. It also names three costs the original
write-up missed — including that our own server would currently reject a
time-based position as malformed — and lists what would make us look again.

Left as **Proposed** rather than Accepted: the call is yours, and the record
should not spend it. Docs only, no code changes. Addresses #138.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -265,6 +265,22 @@ runnable demo for each.
   `Retire.patch` and a `Delete`'s flat `Exclude` had no test coverage at all —
   which is why those two renames were the silent ones — and now do. (#161 item 2)
 
+### Documentation
+
+- **Recorded why stream positions stay a counted number.** The protocol allows a
+  time-based identifier instead, which would remove both the size limit we chose
+  by hand and the brief lock every write takes to hand out the next position. The
+  decision record works through that trade and declines it for now: the lock is
+  what makes a position record the order writes actually arrived in, and a clock
+  cannot give that back once it is gone.
+
+  Two of the three original arguments for switching had already been answered by
+  other work — the size limit now lives in one place and is a one-line change, and
+  the repeated table scan on every write is gone — which is most of why the answer
+  came out this way. The record also names three costs the original write-up
+  missed, including that our own server would currently reject a time-based
+  position as malformed. (ADR 0005, #138)
+
 ## [0.2.0] - 2026-08-15
 
 The first release that describes the library. `0.1.0` was the initial
--- a/docs/adr/0005-stream-positions-stay-a-counted-offset.md
+++ b/docs/adr/0005-stream-positions-stay-a-counted-offset.md
@@ -1,0 +1,196 @@
+# ADR 0005 — Stream positions stay a counted offset; ULIDs are declined for now
+
+- **Status:** Proposed
+- **Date:** 2026-08-22
+- **Deciders:** rakaia maintainers
+- **Related:** [ADR 0002](./0002-framework-vs-protocol-server-boundary.md) (framework
+  vs protocol-server boundary); `docs/protocol.md` § *Strictly Increasing*;
+  `src/rakaia/offsets.py`, `src/django_rakaia/offsets.py`,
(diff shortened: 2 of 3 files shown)


joshbrooks:
## Two of #138's three stated benefits have already been banked

This is the main reason the ADR lands where it does, and it is a correction to the
issue rather than a disagreement with it.

- **The hand-picked width already has one home.** #206 put both formats' five rules in
  `OffsetFormat`. Widening a field is now a one-line edit, not three that must agree.
- **The per-append `Max(offset)` aggregate is already gone.** #138's comment cites
  #62's first item as a ULID benefit. Allocation is now a single locked `get_or_create`
  with the aggregate gated on `watermark.high == 0` — reached once per stream path ever
  (`models.py:181-214`). Pinned by
  `test_a_steady_state_allocation_does_not_aggregate_over_entries` and
  `test_a_steady_state_allocation_costs_two_queries`
  (`tests/test_django_rakaia/test_offset_allocation.py:47`, `:90`).

Remaining cost: **two queries and one row lock per append**, plus the watermark table.
Not a growing table scan.

## Three costs the original write-up missed

1. **Our own server would reject a ULID.** `VALID_OFFSET_PATTERN` (`types.py:66`) is
   `^(-1|now|\d+(_\d+)?)$` — digits only — and `read_decision.py:183` guards every
   client-supplied offset with it. It is also the one position rule #206 did *not* move
   into `rakaia.offsets`. The ADR flags fixing that as worth doing on its own merits,
   independent of this decision, since it would currently reject a conforming
   third-party offset.
2. **Block allocation is a genuine density dependence.** `django_store.py:975-978`
   reserves one contiguous integer block and assigns `start + i`. Nothing needs offsets
   to be *gapless* — reads are a `>` filter or a scan — but this needs N consecutive
   positions from one lock, and `get_next_offset_block(count)` disappears with it. This
   is the direct answer to the issue's "does any consumer rely on offsets being dense
   or arithmetically comparable": yes, one, and it is in allocation, not in reads.
3. **`channels_views.py:44`** uses `parse_offset(...) if last_event_id else 0` as a
   "before everything" sentinel, which has no ULID equivalent.

## Blast radius, counted on this tree

- **46** padded-string literals across **6** test files.
- **17** test functions in `test_offset_allocation.py` asserting integer
  block-allocation semantics that ULIDs would delete rather than rewrite.
- Two columns change: `StreamEntry.offset` (`BigIntegerField`, under a
  `unique_together` and two indexes) and `StreamOffsetWatermark`.
  `ConsumerCursor.offset` is **already** `CharField(max_length=64)`.
- Safety net if we ever do switch: `tests/store_contract.py:171-196` asserts only that
  offsets strictly increase and resume correctly, and `test_offset_format.py`'s
  `TestAThirdPartyStoreIsStillOrdered` already exercises non-digit offsets.

Queries are named inline in the ADR so re-taking these is a paste, not a
reconstruction — they are dated observations, not live state.

## Already ULID-tolerant, just not ULID-minting

#137 answered this same question for `Stream-Seq` the other way:
`migrations/0009_alter_stream_last_seq.py` widened an int column to a `CharField` so a
conforming ULID could be stored, with
`test_append_decision.py::test_seq_is_compared_lexicographically` pinning that nothing
may reject one. `offsets.after` already falls back to byte order for an unrecognised
format. Consuming a foreign ULID costs nothing and works today; *minting* one is what
trades away definite append order, and those are different commitments.

## Status

Left as **Proposed**. The decision is yours to make — flip it to Accepted, or push back
on the recommendation and I will rewrite it the other way. `zensical build` passes with
the nav entry; docs only, no code.

PR #227, opened 2026-08-23
Prepare 0.3.0

joshbrooks:
Marks everything on main since 0.2.0 as a release: the version, the changelog
heading and the upgrade notes. This is a new minor rather than a patch because
several of the changes are breaking — constructing a replay result the old way is
refused, the dashboard views now turn away a position they did not issue, and
timestamps compare equal to the column they came from. Worth knowing before it
ships: partisipa pins the minor exactly, so it will not pick this up until that
pin is moved and validated.

It also adds the upgrade note that should have gone in with the replay batching
change, along with a test for the one part of it a consumer could be relying on.
Batching a whole pass into one transaction means an earlier event's write no
longer survives a later event's failure. That was measured both ways rather than
assumed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to Rakaia are documented here. The format is based on
 New here? The [guided tour](docs/whats-new.md) walks these capabilities with a
 runnable demo for each.
 
-## [Unreleased]
+## [0.3.0] - 2026-08-23
 
 
 ### Added
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -15,7 +15,45 @@ ones you are crossing.
 
 ---
 
-# Unreleased
+# 0.3.0
+
+## A replay applies a whole pass of effects at once, not one event at a time
+
+`replay()` used to call `executor.apply()` once per event. It now buffers a pass's
+effects and hands them over in as few batches as it can, breaking the batch
+whenever holding one more back would change the outcome — a write arriving behind
+a delete, a second write to the same column of the same row, a repeated
+`produces=` id, or a retire that asked for notifications. Passes whose handlers
+receive a reader are not batched at all, because a buffered write would be
+invisible to them.
+
+**What changes for you.** For most consumers, only the transaction and statement
+count: on a nine-event save this went from nine `transaction.atomic()` blocks to
+one, and with `DjangoExecutor(batch_updates=True)` from nine `UPDATE`s to one. The
+rows written, and the order they are written in, are unchanged — there is a test
+that replays the same effects both ways and compares.
+
+Three cases are worth checking against your own code:
+
+- **A custom executor sees fewer, larger batches.** If yours does per-call work —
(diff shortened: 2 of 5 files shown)


joshbrooks:
## Why 0.3.0 and not 0.2.1

`UPGRADING.md` already carried breaking changes for this unreleased window, and the
release adds none of its own. The ones that force a minor:

- `ReplayResult(warnings=…, drift_detected=…)` is no longer accepted; pass
  `drift=DriftLedger(...)`.
- The dashboard views answer `400` for an offset in the other store's format, where
  they used to resolve it to an unrelated position and answer `200`.
- `DEFAULT_NORMALIZERS` gained `normalize_temporal`, so `skip_unchanged=True` stops
  re-`UPDATE`ing rows it used to rewrite — `auto_now`, `post_save` and replication go
  quiet for those rows.
- `poll` raises `ForeignOffset` where it used to return `Poll(status="rewound")`.

13 headline changelog entries in the window.

## The consumer cannot see this yet, by design

`partisipa-import` pins `rakaia-streams==0.2.*` — an exact-minor pin its own comment
calls deliberate and says not to relax to `>=`, because rakaia is pre-1.0. So
publishing 0.3.0 does **not** reach it. Testing it there is a second, coordinated
step: move the pin to `==0.3.*`, work `UPGRADING.md` § 0.3.0 top to bottom, and
validate. That repo's comment already requires validation on a rakaia bump.

Flagging it here so "cut a release for partisipa to test" is not mistaken for one
task. It is two, in two repos.

## The gap this closes

#224 shipped without an `UPGRADING.md` entry. That was my miss. The batching change is
consumer-visible in three ways, now written down:

1. A custom executor sees fewer, larger batches (the flat effect sequence is identical).
2. A `Ref` to an earlier event's `produces=` row now resolves where it used to raise —
   a widening; the executor-level rule that refs do not cross two explicit `apply()`
   calls is unchanged and still pinned.
3. **A failure part-way through a pass discards the whole batch.** This is the one a
   consumer could genuinely be leaning on, and it had no test.

Measured both ways rather than asserted, on the same scenario — two events, the second
writing a column that does not exist:

| | rows surviving |
|---|---|
| per-event apply (pre-#224) | `['good']` |
| batched pass (now) | `[]` |

`test_a_failure_part_way_through_a_pass_discards_the_whole_batch` pins the second row,
and the control for the first is in its docstring. Per-event commit granularity within
a pass is no longer on offer; that is now stated in `UPGRADING.md` rather than
discovered.

## Gate

`ruff check` / `ruff format --check` / `pyright src/` 0 errors / `pytest` **1960 passed,
16 skipped, 1 xfailed** / `zensical build` no issues / `api-reference-check` clean /
`test_version.py` passes against the bumped version (it is single-sourced from
`pyproject.toml`; nothing in `src/` hard-codes it).

## After merge

`just release` runs `check` + `build`, and `.github/workflows/publish.yml` handles
PyPI. Tag `v0.3.0` off the merge commit. I have not tagged or published anything.

PR #237, opened 2026-08-26
Give two shared rules one home each

### LINKED ISSUE #161: Smaller items from the same review

joshbrooks:
Closes the last two items of the smaller-findings list from the architecture review, so #161 and the epic above it can close.

The payload a resolved-item notification carries was invented in the replay orchestrator, with its state written as a bare literal, while the request for one is defined elsewhere. It now has a single definition beside that request. The ASGI entry point resolved its store from a setting internally, so the only way to drive it was to reconfigure a process-wide cache — which is why it had no tests at all. It now accepts a store, as the replay entry point already did.

Closes #161.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -23,7 +23,7 @@ page is the contract.
 |---|---|---|---|
 | `app` | `rakaia` | `(scope: 'Scope', receive: 'Receive', send: 'Send') -> 'None'` | — |
 | `create_app` | `rakaia` | `(store: 'StreamServerStore \| None' = None, options: 'ServerOptions \| None' = None) -> 'Any'` | Create a plain ASGI application implementing the Durable Streams protocol. |
-| `get_asgi_app` | `django_rakaia` | `(options: rakaia.handler.ServerOptions \| None = None, store: typing.Any \| None = None) -> object` | Get the Rakaia ASGI application configured with the store `RAKAIA_STORE` names. |
+| `get_asgi_app` | `django_rakaia` | `(options: rakaia.handler.ServerOptions \| None = None, store: rakaia.protocols.StreamServerStore \| None = None) -> object` | Get the Rakaia ASGI application configured with the store `RAKAIA_STORE` names. |
 | `ServerOptions` | `rakaia` | `(long_poll_timeout: 'float' = 3.0, cursor_options: 'CursorOptions' = <factory>, enable_fault_injection: 'bool' = <factory>) -> None` | Configuration for the ASGI handler. |
 | `get_store` | `django_rakaia` | `() -> Any` | Get the configured stream store. |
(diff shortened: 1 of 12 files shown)


joshbrooks:
Technical detail, and the controls for both items.

**Item 4 — the notification payload.** `TRANSITION_RESOLVED` and `transition_payload()` now sit beside `Transition` in `effects.py`; `_synth_transitions` calls it. Neither is exported — the item was about one definition, not new public surface.

The obvious test does not work here, and that is worth recording. Asserting the payload *value* against `transition_payload(...)` passes whether or not the orchestrator calls it, because an inline `{**patch, "key": …, "state": "resolved"}` builds an equal dict — I wrote that test first and the control confirmed it: re-inlining the dict left all four `TestSynthTransitions` cases green. The pin that works substitutes the shared definition and asserts the orchestrator followed it. Control: re-inlining fails that test, and only that test.

That substitution goes through `sys.modules["rakaia.replay"]`, not `monkeypatch.setattr("rakaia.replay.transition_payload", …)` — the dotted form resolves `rakaia.replay` to the *function*, which is item 1 of this same issue, and would silently patch nothing.

**Item 3 — the store global.** `get_asgi_app` takes `store=`, matching `django_rakaia.replay`, and `reset_store_cache()` replaces the private-dict poke in `test_store_selection.py` (the one place the suite reached past an interface by design). `test_a_refused_backend_is_not_cached` is now behavioural — calling twice must raise twice — rather than inspecting `_stores`.

New file `test_integration_entry_point.py` covers the entry point that had none: an append lands in the supplied store, two apps with two stores do not share, an explicit store bypasses the setting entirely, and `options` are threaded through. Control: reverting `get_asgi_app` to resolve internally turns three of the five red. `reset_store_cache` is tested from both sides — that memoisation is real, and that the reset drops it — since a no-op reset would pass a test that only checked the second.

`reset_store_cache` is the one added export, so `test_public_api.py`, `GROUPS` in `scripts/gen_api_reference.py` and the generated reference all move with it.

Gates: `ruff check`, `ruff format --check`, `pyright src/` (0 errors), `pytest` (1996 passed, 16 skipped, 1 xfailed), `zensical build`, `just api-reference-check`. SQLite leg only — nothing here touches locking.

Issue #229, opened 2026-08-23
Keep streams in plain text files on disk


joshbrooks:
Rakaia can currently keep a stream in one of two places: in the memory of the running process, or in a Django database. The first is only good for tests and demos, because everything vanishes when the process stops and nothing else can read it. The second needs a whole database. There is nothing in between, and there is no third example proving that someone else could plug their own storage in — which is a promise the library makes.

This adds a third option that writes each stream to ordinary text files in a folder, one line per event, rolling over to a new file every so many events. It needs no database and no extra software, the files can be read with any text editor, and pointing it at a memory-backed folder makes it as fast as the in-memory option while still being readable from another process. If that works out, the in-memory store stops being the only way to run without a database, and we can look at retiring it.

joshbrooks:
Implementation sketch. Full scoping in `docs/research/alternative-backends.md` (uncommitted as of writing); this is the JSONL half of it.

**Layout**

```
data/<stream>/
  meta.json              head, closed, closed_by, content_type, ttl, expires_at,
                         last_seq, producer states
  000000000000.jsonl     entry ids 1..9999
  000000010000.jsonl     entry ids 10000..19999
```

One JSON object per line: entry id, offset token, label, metadata, `event_ts`, payload. A segment seals at N entries; sealed segments are immutable, the filename is the index, and retention becomes "delete a file".

**Surface.** `rakaia.protocols` has two levels. Level 1 is `WritableStore` + `CursorStore` (`read`, `has`, `create(path)`, `append`, `append_many`, `get_current_offset`) — what `replay()` and the registry need. Level 2 is `StreamServerStore`, adding `run_sync`, the widened `create(...)`, `get`, `touch`, `delete`, `format_response`, `append_with_producer`, `close_stream`, `close_stream_with_producer`, `wait_for_messages`. Definition of done is `tests/store_contract.py` and `tests/server_store_contract.py` both green against the new store — a ~15-line subclass inherits ~1,300 lines of behaviour.

Reused rather than rewritten: `rakaia.append_decision` (admission ordering and the all-or-nothing batch pre-flight), `rakaia.producer`, `rakaia.json_mode`, `rakaia.offsets`, `rakaia.context`.

**The four real costs**

1. *Serialisation.* No `select_for_update` equivalent. An exclusive `fcntl.flock` on a lockfile in the stream directory, held across the whole check-then-write including the `meta.json` update. POSIX-only unless we add a Windows path.
[...]

Issue #232, opened 2026-08-23
A saved reading position should say which store it came from


joshbrooks:
Every consumer of a stream remembers where it got to, as a short position marker. Two of the three storage backends now produce markers that look exactly alike, because they both count events the same way. That is what makes moving a log between them work smoothly, but it also means a position saved while reading from one backend is silently accepted by the other, and points at whatever event happens to sit at that number there.

The fix is to record which backend a position came from, next to the position itself, and refuse to use it against a different one. That turns a wrong answer nobody notices into an error that names the problem. It does not need the position markers themselves to change, which would rewrite the meaning of every one already stored.

joshbrooks:
Technical detail. Split out of #229; see that issue's parity comment for how this arose.

**The gap.** `rakaia.offsets.FORMATS` distinguishes `COMPOUND` (in-memory, `{read_seq}_{byte_offset}`) from `PLAIN` (a zero-padded entry id) by field count, and `after()` raises `ForeignOffset` when two recognised tokens come from different formats. `DjangoStreamStore` and `JsonlStreamStore` both issue `PLAIN`, so that guard cannot separate them. A cursor saved against one and used against the other resolves to whatever entry sits at that id, and `poll()` returns `advanced`/`caught_up` — a confident wrong answer, which is the exact outcome `rakaia/offsets.py` was written to refuse for the other pair.

**Where the guard belongs: the cursor, not the token.** `django_rakaia.subscription.poll_consumer(store, consumer_id, stream_path)` already has the store *and* the cursor row in one place. Proposal:

- add a `backend` column to `ConsumerCursor` (a short identifier the store reports — `"durable"`, `"jsonl"`, `"memory"`);
- `commit_cursor` stamps it; `load_cursor` refuses a mismatch with an error naming both backends and pointing at `migrate_stream`;
- an unstamped row (from before the column existed) is accepted once and stamped, so existing deployments do not break.

The store needs to report its own identity for this. Either a small `backend_id` attribute on the store classes, or a mapping in `django_rakaia.store` alongside `BACKENDS` — the mapping avoids putting a Django-facing concept on a framework-tier class.

**Explicitly not proposed:** a discriminator inside the offset token (`jsonl:000…`). It rewrites the meaning of every offset the durable store has already handed out, forces a data migration, reopens ADR 0005's format table, and catches nothing that the cursor stamp does not catch at the only point where it can do harm.

[...]

PR #238, opened 2026-08-26
Keep streams in plain text files on disk

joshbrooks:
A third place to keep a stream: ordinary text files in a folder, one line per event, rolling over to a new file every so many events. No database, no extra software, and the log can be read with any text editor. Pointing it at a memory-backed folder makes it about as fast as the in-memory option while still being readable from another process.

The point is not only the backend. Rakaia promises that someone can plug their own storage in behind the same server, and until now there was no third example proving it. This one is held to the two existing contract suites and to the database-backed store's own test files, so the promise is now demonstrated rather than asserted. It also comes with a way to copy a log from one backend to another that reports whether saved reading positions survived the move.

Rebased from `spike/jsonl-store` onto current main; that branch is left untouched. Detail, including what is deliberately not covered, is in a comment.

Part of #229.


joshbrooks:
Rebased from `spike/jsonl-store` (8 commits, forked at #235) onto main, which had moved two commits since — 0.3.1 (#236) and #237. `spike/jsonl-store` is untouched; this is a new ref.

**What the rebase collided with, since it is worth knowing what was reconciled rather than dropped.** Both conflicts were in files #237 had just changed:

- `django_rakaia/store.py` — the singleton comment. #237 rewrote it to scope the "never rebuilt or replaced" invariant to the serving path and record what `reset_store_cache` costs a non-test caller; this branch rewrote the same lines to mention the JSONL store is stateless. Merged both, and extended the reset note to say the other two backends survive a reset because their state is not in the object.
- `test_store_selection.py` — pure adjacency. #237 added `TestTheCacheAndItsReset`, this branch added `TestTheJsonlStoreDoesNotBroadcast`. Both kept. Checked specifically that the rebase did not reintroduce the private `_stores` poke #237 removed: it did not.

**Two gate failures the branch had before the rebase**, both mechanical and now fixed in the last commit:

- `ruff format --check` failed on `docs/store-streams-in-files.md` — ruff formats Python blocks inside Markdown, and that file had never been through it.
- `just api-reference-check` was one name behind, since `reset_store_cache` landed in #237 after this branch generated its reference. This is exactly the count-line collision CLAUDE.md warns two branches will both rewrite.

**Gates on the rebased tree:** `ruff check` clean, `ruff format --check` clean (471 files), `pyright src/` 0 errors (only the pre-existing `reportUnsupportedDunderAll` warning), `zensical build` clean, `just api-reference-check` clean, full suite **2190 passed, 16 skipped, 1 xfailed** — up from 1996 on main, and the #161 item-1 strict xfail still xfails. Postgres leg not run locally; CI will.

**Draft on purpose.** This is 3,704 lines including a file-locking concurrency store, and I have verified that it rebases and passes, not that it is correct. The issue thread records six defects the parity pass found *in the spike* — a lock that held nothing, two writers racing for one scratch file, a reader rolling the head backwards — which is the signature of code that rewards adversarial review. Recommend a review loop before this comes out of draft.

**Declared gaps, carried over from the issue thread, not re-litigated here:** POSIX-only (`fcntl`, optional import so `import rakaia` still works on Windows); unsafe over NFS; no channels broadcast, reported as `rakaia.W002` at `manage.py check`; and the offset-format question — this store issues `PLAIN` tokens like the durable one, so a cursor cannot be told apart between the two. That last one is #232, and it is a live silent wrong answer the moment this lands.

PR #239, opened 2026-08-26
Write down that changing a backend is a copy

### LINKED ISSUE #233: Write down the rule for moving a log between backends
There is now a way to copy a log from one storage backend to another, and it reports whether reading positions survived the move. What is not written down anywhere is the rule that copying is the *only* supported way to change backends — pointing the application at a different one and restarting is not, because it starts reading an empty log while every consumer still holds a position that looks valid.

The decision record covering how stream positions work also predates the third backend, so its table of position formats is one row short and does not mention that two of the three now look identical. Both are small pieces of writing, and both are the kind of thing that is much cheaper to record now than to reconstruct later.

joshbrooks:
There is a supported way to move a log between backends and an obvious way, and they are not the same one. Changing the setting and restarting moves nothing: the application comes up against a different, empty log while every saved reading position still looks valid. That rule now has a decision record, rather than living only in a page you read once you have already chosen the file store.

The record covering how positions work needed amending too, and the amendment is smaller and stranger than adding a backend usually is: no new kind of position was invented, because the new backend reuses the one the database-backed store already hands out. So the table gains a second store against an existing entry rather than a new entry, and what changes is what the table *means* — a position can no longer be told apart by its shape, which was never a guarantee anyone designed. It held by accident while the two kinds happened to differ.

Closes #233.

### DIFF
--- a/docs/adr/0005-stream-positions-stay-a-counted-offset.md
+++ b/docs/adr/0005-stream-positions-stay-a-counted-offset.md
@@ -7,6 +7,8 @@
   vs protocol-server boundary); `docs/protocol.md` § *Strictly Increasing*;
   `src/rakaia/offsets.py`, `src/django_rakaia/offsets.py`,
   `src/django_rakaia/models.py`, `src/rakaia/types.py`.
+  [ADR 0006](./0006-changing-backends-is-a-copy.md) (changing a backend is a copy —
+  amends the format table below).
   Issues: [#138](https://github.com/joshbrooks/rakaia/issues/138) (this decision),
   #206 / #182 / #178 (one home for the five position rules — landed),
   #34 (global monotonicity — closed, and the reason the watermark exists),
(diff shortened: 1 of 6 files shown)


joshbrooks:
Technical detail, and the one thing writing it down caught.

**ADR 0006** states the rule and, more usefully, the three-row table it follows from — which pairs of stores refuse each other's positions and which pair does not:

| Old → new | Resumed consumer gets |
|---|---|
| in-memory → either other | `ForeignOffset`. Loud, immediate. |
| `DjangoStreamStore` ↔ `JsonlStreamStore` | **Accepted.** Both issue `PLAIN`. |

The accepted row splits in two, and only one half is safe. A cursor *ahead* of the new store's head is reported as `rewound` and re-reads — over-cautious but loud. A cursor *at or below* it is accepted silently, and every event before it is skipped permanently. Measured, not reasoned: `rewound=False`, no error, events 0-2 of a 10-event log never delivered.

That difference is the reason this is a decision record rather than a note. An operator cannot derive it from behaviour, and reaching the bad half needs only a backend switch — a once-per-deployment action done from memory.

**ADR 0005's format table** gains the third store and a dated amendment saying what the two-row version could not: two backends issuing one format means `offsets.after` can no longer tell a foreign cursor from a local one. Nothing in 0005's decision changes; the widths, the lock and the ULID answer are untouched.

**The table is pinned by tests, not left as prose** — the #176 precedent for the producer-fencing table. `test_cross_backend_cursors.py` covers the two loud rows and *both* directions of the accepting one, including an explicit assertion of the silent skip, so leaving that defect is a recorded decision rather than an oversight and #232 landing announces itself here. `test_django_rakaia/test_offset_format_pin.py` carries the fact that makes the pair real: `DjangoStreamStore` issues the same format, asserted against a real durable store rather than a stand-in.

**Writing the alternatives down caught one of them being wrong**, which is the part worth keeping. I had written "give `JsonlStreamStore` its own offset format" as a real option, declined on trade-offs. It is not an option as stated: `OffsetFormat.owns` matches on **field count, not width**, deliberately, so a client can send an unpadded `?after=42`. Minting 18-digit offsets instead of 20 left `format_of` still reporting `durable` and every cross-store cursor still accepted — my first mutation control passed for exactly this reason, which is how I found it. Separating the two would need a second numeric field or a non-numeric shape. The ADR says that now, and cites the test that already pins it.

Control for the pin: giving the file store a genuinely two-field format turns 3 of the 8 new assertions red and errors 3 more. Gates: `ruff check`, `ruff format --check` (474 files), `pyright src/` 0 errors, `pytest` 2205 passed / 16 skipped / 1 xfailed, `zensical build`, `just api-reference-check` — all clean.

Rebased onto `main` after #238 merged mid-work; `--onto` rather than a plain rebase, since the squash means git cannot see those 8 commits as landed.

PR #240, opened 2026-08-26
Importing the framework no longer starts a server, and the split question is answered

### LINKED ISSUE #191: Revisit whether the two halves of rakaia should be separate packages
An earlier decision recorded that the project is really two things sharing a package — the event-sourcing framework people build on, and the streaming protocol server — and deliberately did not split them, on the grounds that it was too early. It named the condition for revisiting: when the database-backed store grows into a full protocol server implementation.

That has since happened, and the decision record itself notes it. This is not a proposal to split, only a note that the trigger the earlier decision set has fired, so the question is now open rather than settled. Several smaller items sharpen the types the two halves share, which is the evidence needed to answer it either way.

joshbrooks:
Reaching for the event-sourcing half of this library used to drag the whole protocol server in with it, and build a stream store in every process that imported it, whether or not anything ever served a request. That was about eighty milliseconds and one unwanted store, for people who wanted neither. Names now resolve when something asks for them, so a consumer loads the half it uses and nothing else.

The other half is a decision. A record from July said the two halves ought to eventually become separate packages, and set a condition for revisiting that; the condition fired a while ago and the question has since been measured twice with the same answer. It is closed rather than deferred a third time — they stay together, and the reasons for splitting have been answered without splitting. New conditions for reopening are written down in place of the spent one.

Closes #191.

### DIFF
--- a/docs/adr/0002-framework-vs-protocol-server-boundary.md
+++ b/docs/adr/0002-framework-vs-protocol-server-boundary.md
@@ -1,6 +1,7 @@
 # ADR 0002 — Name the framework ↔ protocol-server boundary; formalize the framework seams
 
-- **Status:** Accepted (items 2 and 5 landed; boundary enforced 2026-08-21 — see the update notes below)
+- **Status:** Accepted (items 2 and 5 landed; boundary enforced 2026-08-21; the
+  package split declined and #191 closed 2026-08-26 — see the update notes below)
 - **Date:** 2026-07-21
 - **Deciders:** rakaia maintainers
 - **Related:** [`django-integration.md`](../django-integration.md),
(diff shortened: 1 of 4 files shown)


joshbrooks:
Technical detail, and the one thing that nearly shipped as a regression.

**Measurements, all on this tree.**

| | before | after |
|---|---|---|
| `import rakaia` | 80 ms, 10 server modules, 1 `StreamStore` | 2.4 ms, 0, 0 |
| `from rakaia import replay, Upsert, register_handler` | 80 ms, 10 server modules | 3.3 ms, **0** |
| `from rakaia import create_app, StreamStore` | 80 ms, 10 | 25 ms, 7 |

The store mattered as much as the milliseconds: `app = create_app()` ran at module scope, so every process that imported the package allocated an in-memory store it never used.

**`replay` cannot be lazy, and finding out why is the useful part.** It is the only export whose name is also a submodule name. Left lazy, `import rakaia.replay` binds the *module* onto the package, and once that attribute exists `__getattr__` is never consulted — so `from rakaia import replay` returned a **function or a module depending on what else the process had imported first**. Measured across four import orders: three gave the function, one gave the module. That is strictly worse than the deterministic shadowing of #161 item 1, which it would otherwise have fixed. It is bound eagerly now, and all four orders give the function, matching the previous behaviour exactly.

**The `TYPE_CHECKING` block is a second copy of the export list**, and drift in it is silent in the worst way: a missing name still resolves at runtime, but pyright types it `Any`, so annotations checked against it quietly stop being checked. Ruff cannot see the block as used either, now that `__all__` is computed. So both directions are asserted — every `_EXPORTS` key appears in the block, and the block imports nothing unexported — and the `F401` exemption in `pyproject.toml` cites those tests rather than claiming a guarantee that was not there.

Worth flagging: I first wrote that exemption asserting a test enforced the agreement, before the test existed. Writing the justification is what exposed it.

Controls (each mutation reverted, file-changed asserted): dropping one name from the block, adding an unexported one, and listing `replay` in it each turn exactly one of the three new tests red.

**On the split (#191).** Two new measurements decide it. Deleting the *entire* protocol-server tier leaves **11 of 12** framework modules importing cleanly — the twelfth is `seed`, the one entry already in `ALLOWED_CROSSINGS`. And `django_rakaia`, the largest consumer, imports **16 framework / 3 protocol-server / 14 shared**: it is mostly a framework consumer, so a split would leave it depending on all three resulting packages and isolate nothing. The boundary also survived #229 — 1,405 lines of new protocol-server code — with the crossing list unchanged, which the tier map forces rather than leaves to luck.

Gates: `ruff check`, `ruff format --check` (474 files), `pyright src/` 0 errors, `pytest` 2208 passed / 16 skipped / 1 xfailed, `zensical build`, `just api-reference-check` — all clean. The public surface is byte-identical: 103 names, none added, none missing, all resolvable.

PR #241, opened 2026-08-26
Stop deciding what another store's positions look like

### LINKED ISSUE #226: Our own server would turn away a position it did not issue
A position in a stream is meant to be opaque: the store that issued it is the only
thing that can say what it means, and the check in front of our server is only
supposed to catch junk arriving in a web address. It does more than that — it
insists a position looks like one of our own two formats. Rakaia is built to let
someone plug in their own storage behind the same server, and if that storage
numbers positions any other way, our server rejects the positions it just handed
out, before the storage is ever asked.

Nothing is broken today, because both of our own stores number positions the way
the check expects. It is worth knowing that this has already happened once, though:
the check used to reject the format our database-backed store uses, which was
harmless right up until that store could sit behind the server, at which point
every resume failed on a position the server had itself issued.

Deferred on purpose — see the comment for the un-defer trigger. Comes out of ADR
0005 (#138), which names it as worth fixing on its own merits and independent of
the decision that record makes.

joshbrooks:
The check in front of our server was described as a guard against junk in a web address, and was really a list of our own two position formats. Rakaia invites people to plug their own storage in behind the same server — and now ships three examples of doing so — but any storage that numbers positions differently would have had its positions turned away before it was ever asked, on the grounds that they do not look like ours.

This has already happened in-house, which is the reason to fix the rule rather than the case. The list once held only the in-memory format; that was harmless right up to the moment the database-backed store could sit behind the server, at which point every resume failed on a position the server had itself just issued. It was fixed by adding a format to the list.

The rule now says what the specification says, and nothing about shape. Widening it is safe because a store still refuses a position it did not issue, with the same status — what changes is which layer decides, and the store is the layer the design already names as the authority.

Closes #226.

### DIFF
--- a/docs/adr/0005-stream-positions-stay-a-counted-offset.md
+++ b/docs/adr/0005-stream-positions-stay-a-counted-offset.md
@@ -98,11 +98,15 @@ backwards would violate the protocol's own MUST.
 
 Three further costs are concrete, and two of them were not in the issue:
 
-1. **Our own server would reject a ULID.** `VALID_OFFSET_PATTERN`
-   (`src/rakaia/types.py:66`) is `^(-1|now|\d+(_\d+)?)$` — digits only — and
-   `read_decision.py:183` guards every client-supplied offset with it. It is also
-   the one position rule #206 did not move into `rakaia.offsets`, so it is the
-   first thing any ULID work has to relocate and widen.
+1. **Our own server would reject a ULID.** `VALID_OFFSET_PATTERN` (then in
+   `rakaia.types`) was `^(-1|now|\d+(_\d+)?)$` — digits only — and guarded every
+   client-supplied offset. It was also the one position rule #206 did not move
+   into `rakaia.offsets`, so it was named here as the first thing any ULID work
+   would have to relocate and widen. **Done, 2026-08-26 (#226):** it is
+   `offsets.is_syntactically_valid`, and it now implements §6's actual rule — no
+   forbidden characters, no whitespace, under 256 — rather than enumerating
+   rakaia's own formats. This cost is retired, and it was worth paying whether or
+   not we ever mint a ULID, which is decision 3 below.
 2. **Block allocation is a genuine density dependence.** The bulk-append path
    reserves one contiguous integer block and assigns `start + i`
    (`src/django_rakaia/django_store.py:975-978`). Nothing depends on offsets being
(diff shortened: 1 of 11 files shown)


joshbrooks:
Technical detail.

**The rule, verbatim from `docs/protocol.md` §6.** `offsets.is_syntactically_valid` refuses: empty, anything containing `,` `&` `=` `?` `/`, anything with whitespace, and anything at or over 256 characters. That is all. `MAX_OFFSET_LENGTH` is exported so the ceiling has one home rather than being a literal in a regex.

Whitespace is the one addition to what §6 names. The spec asks for URL-safe characters without listing them; a token with a space in it is a mangled request rather than a store's considered choice of format, and it is what `test_read_decision`'s existing case uses.

**Where it moved and why that is the point.** #206 gave a stream position's five rules one home each in `rakaia.offsets`; validity was the one that did not move, and ADR 0005 decision 3 named it as wrong where it was, independent of the ULID question. It is now in `rakaia.offsets` with the other four, and `rakaia.types` no longer decides anything about what an offset may look like. ADR 0005 is updated in place — both its cost 1 and its decision 3 now record this as done, and note that the claim held: widening committed us to nothing.

**Behaviour changes, both predicted by the issue.** The 400 body for a malformed offset goes `b"Invalid offset format"` → `b"Invalid offset"`, asserted in exactly one place; no conformance test asserts it and the protocol does not mandate body text. And `test_rejects_malformed` largely inverts — `abc`, `1.2`, `-2`, `NOW`, `1_`, `_1`, `1_2_3`, `1_-2` are now guard-accepted and store-rejected. Those cases were rewritten to assert the request still 400s **end to end** rather than that a particular layer refused, which is a better assertion regardless: the old form made the server the authority on offset shape, and recorded tokens as "malformed" when the only true statement was "no rakaia store issues this".

Added: a ULID (`01ARZ3NDEKTSV4RRFFQ69G5FAV`) reaches the store rather than being turned away — §6 names ULIDs as acceptable and ADR 0005 keeps rakaia ULID-*tolerant* while declining to mint them.

**Controls.** Reverting the whole guard to the old digits-only enumeration turns 10 assertions red. Each of the three halves was then broken separately — the forbidden-character set, the whitespace check, the length ceiling — and each turns its own cases red and nothing else's (5, 5, 1). Plus an end-to-end control that a real resume round-trips 200 with the expected message, so none of the refusal cases is passing by refusing everything.

Also swept the stale references: seven comments and docstrings across `handler.py`, `store.py`, `offsets.py` and three test files described the old name and the old rule. No reference to `VALID_OFFSET_PATTERN` survives outside the note recording where it went.

Gates: `ruff check`, `ruff format --check` (474 files), `pyright src/` 0 errors, `pytest` 2230 passed / 16 skipped / 1 xfailed, `zensical build`, `just api-reference-check` — all clean.

Draft pending a review loop.

PR #242, opened 2026-08-27
The translations demo keeps its log in files

joshbrooks:
The translations demo used to keep its event log in the database. It now keeps
it in a folder of ordinary text files, one line per event, which you can read
with `less` and follow with `tail -f` while the demo runs.

Live updates still work, and one thing that never worked now does: a change made
from any process reaches every open browser. Previously only saves made inside
the web server itself showed up live, which is why the stress command had to
post over HTTP to be worth watching.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,10 @@ htmlcov/
 django_rakaia_dev.db
 db.sqlite3
 
+# The polyglot sample's event log: a folder of JSON-lines files, written at
+# runtime the way db.sqlite3 above is.
+examples/polyglot/streams/
+
 # Collected static files (chat sample → WhiteNoise)
 examples/chat/staticfiles/
 
@@ -57,3 +61,4 @@ conformance-server.log
 
 # Agent worktrees — transient checkouts, never part of the tree
 .claude/worktrees/
+.gstack/
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -71,7 +71,7 @@ Real-time demos — run the server, open the URL, watch events stream in.
 | Example | Proves | Run |
 |---|---|---|
 | [`chat`](https://github.com/joshbrooks/rakaia/tree/main/examples/chat) | `@stream_model`, multi-stream events per save, live SSE fan-out | `just dev` → `http://localhost:8000` |
-| [`polyglot`](https://github.com/joshbrooks/rakaia/tree/main/examples/polyglot) | Language-scoped streams, live-editable translations over SSE | `just polyglot-dev` → `http://localhost:8001` |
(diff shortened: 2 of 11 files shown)


joshbrooks:
Technical detail.

**Why `RAKAIA_STORE = "jsonl"` was not enough on its own.** Two things blocked it:

* `create_stream_event` is the ORM door. It writes `StreamEvent`/`StreamEntry`
  rows directly and never consults `RAKAIA_STORE`, so the database would have
  kept filling while the log the app served from stayed empty.
  `polyglot/signals.py` now appends a JSON record through `get_store()`. The
  demo gives up the envelope that door builds — provenance, one event fanned
  into several streams — for a plain record, which is all the page renders.
* `django_rakaia`'s SSE endpoint replays history from `StreamEntry` rows and
  waits on the channel layer. Neither exists when the log is a folder.
  `polyglot_project/asgi.py` mounts `rakaia.create_app` on `/protocol/` over the
  same directory; `_handle_sse` reads through the store interface and blocks in
  `store.wait_for_messages`, which `JsonlStreamStore` answers by watching the
  directory. The prefix is stripped by hand — `URLRouter`/`path()` do not strip
  it, and the handler routes on the URL path verbatim.

**Stream paths moved** from `translations:tet` to `/translations/tet`, so the
stream id and the URL that reads it are the same string. The client now reads
the protocol server's named `data`/`control` frames instead of Django's single
unnamed one, from
`/protocol/translations/{lang}?live=sse&offset=0`. The offset is mandatory —
`live=sse` without one is a 400 — and `0` rather than `now` so a reconnect
replays and converges rather than keeping whatever the pane was showing when the
connection dropped.

`django_rakaia.urls` is no longer included. Every route in it is ORM-backed, so
mounted it would serve a permanently empty dashboard next to a visibly streaming
page. `rakaia.W002` is silenced with a comment saying where live delivery went.

**Throughput got worse on the defaults, and the README says so.** 400 `--direct`
saves on one laptop, one writer: 374/s before, 73/s now. Not like-for-like —
SQLite in WAL with `synchronous=NORMAL` never fsyncs on commit, so those events
survived a crashed process but not a power cut; this store fsyncs every append.
`POLYGLOT_STREAM_FSYNC=0` gives 195/s and a `/dev/shm` root 242/s, both new env
knobs alongside `POLYGLOT_STREAM_ROOT`. Twelve writers on one stream gives 84/s:
`flock` serialises appends per path, the file-level counterpart of
`select_for_update()` on the stream row.

**Verified.** Full gate green (ruff, `pyright src/` at 0 errors, 2231 tests,
`zensical build`, api-reference check). Beyond that: an out-of-process save
arriving as an `event: data` frame on an open SSE connection;
`stress_translations --direct` at 20 scrambles plus 11 restores, all 31 seen by
an open reader; and a real browser updating its preview pane both from an
out-of-process ORM save and from the Save button, with no console errors.

PR #243, opened 2026-09-05
An outcome is recorded where the cursor is committed (ADR 0007 + the core seam)

joshbrooks:
A marker of how far a consumer has read cannot say whether it got there cleanly, so
an event that was skipped, declined or lost looks exactly like one that worked.
Nothing being recorded reads as everything being fine. This adds the decision record
for what to record instead, along with the types, the seam and a shared conformance
suite that any backend keeping those records has to satisfy.

The decision is proposed rather than settled, and it is the thing that wants review
— the code is here because writing it is what found the decision's mistakes. Two
backends exist so far, one in memory and one in files, which is what makes the
"works on any backend" claim a measurement rather than an assertion. Storing them in
a database, the loop that writes them, and retrying a failure are all still to come;
retrying is deliberately last, because it writes to the projection and needs the
existing verification commands extended first.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -89,6 +89,28 @@
 - Research notes live in `docs/research/` and are deliberately **not** in the
   nav. They are dated and are not decisions; decisions go in `docs/adr/`.
 
+## Writing it up
+
+- **Two plain paragraphs.** A commit message body, an issue body and a PR body are each
+  at most two paragraphs, written so someone who only reads the notification email gets
+  the whole point. Say what changed and what it means for whoever uses this. That is the
+  deliverable.
+- **Everything else goes in a comment.** Tables, measured counts, SHAs, query output,
+  repro steps, mutation records, per-file reasoning — post them below the body, on the PR.
+  They are evidence for whoever verifies the work, not the summary for whoever reads it. A
+  commit message has nowhere to put them, so they belong on its PR instead.
+- **No machinery in the body.** This library's own words — stream, event, consumer,
+  cursor, offset, replay, backend — are the vocabulary and are fine. Identifiers are not:
+  `DjangoExecutor.apply` is "the part that writes a batch", `_StageBuffer` is "batching a
+  pass together". Save the precise names for the comment, where precision is the job.
+- The test: read the body alone. If it needs the source open, or the point arrives after
+  the evidence, rewrite it.
+- **Docstrings too: two paragraphs**, with one difference — a docstring's reader is
+  looking at the code, so identifiers *are* their vocabulary. What does not belong is the
+  essay: the history of a bug, what an earlier cut did, or the same point twice. A
+  non-obvious constraint earns its own short paragraph; if it took a mutation to find,
+  write it down.
+
 ## Agent skills
 
 ### Issue tracker
--- a/docs/adr/0007-an-outcome-is-recorded-where-the-cursor-is-committed.md
+++ b/docs/adr/0007-an-outcome-is-recorded-where-the-cursor-is-committed.md
@@ -67,6 +67,20 @@ because the decision is store-agnostic; only the place to keep the answer is Dja
 An outcome keyed by `(consumer, stream_path, subject)` is store-agnostic in exactly the
 same way, and there are now three stores.
 
+## What is built, and what this decision only proposes
+
+A decision record says what was decided; it should not be read as saying what exists. Of the
+decisions below, **3, 4, 6, 6a, 6b and 7 describe code in the tree.** Decision 1's core half
+does; its Django half does not. **Decisions 2 and 5 describe nothing that exists** — there is
+no consume loop and no `on_error` parameter — and the Consequences section's "'did we lose
+anything, and which ones?' becomes a query, and it survives a restart" is a claim about the
+finished thing, not about what is merged.
+
+That split is deliberate: what is merged is a record and the two stores that keep it, which
+is the part worth arguing about before the loop that writes them is built. It is called out
+here because an earlier version of this section did not, and a reader would reasonably have
+taken the Decision list for an inventory.
+
 ## Decision
 
 **1. An outcome is a core concept with pluggable storage.** `Outcome`,
@@ -76,8 +90,8 @@ durable place to keep them and nothing else, mirroring `load_cursor`/`commit_cur
 A shared contract suite at `tests/outcome_store_contract.py` is subclassed once per
 backend, so "does this work on JSONL?" is a red test rather than a question.
 
-**2. The record is written where the cursor is committed.** Rakaia gains the consume
-loop it has never had, next to `poll`. It applies, records any outcome in its own
+**2. The record is written where the cursor is committed.** *(Not built — this decision
+proposes it.)* Rakaia gains the consume loop it has never had, next to `poll`. It applies, records any outcome in its own
 transaction — outside the executor's — and then commits the cursor. Nothing is
 recorded from inside an `Effect`, an executor, or a handler.
 
@@ -128,7 +142,8 @@ Nothing is missing; something was rejected. Treating that as a lost append would
 re-drive hunting for a fact that is exactly where it should be, and treating it as an
 unapplied projection would send a replay looking for an event that was never written.
 
-**5. The failure policy is a parameter with per-mode defaults, never inferred.**
+**5. The failure policy is a parameter with per-mode defaults, never inferred.** *(Not
+built — this decision proposes it.)*
 `on_error="skip"` when consuming continuously, `on_error="halt"` when rebuilding. The
 same shape as `on_drift` (`src/rakaia/drift.py:41`), and for the same reason: one code
 path serving two operations with opposite invariants is how a guard ends up correct in
@@ -324,9 +339,11 @@ that is the point of it.
 
 ### Positive
 
-- "Did we lose anything, and which ones?" becomes a query, and it survives a restart.
-- The consume loop lands as a first-class thing. Rakaia has documented at-least-once
-  semantics and, until now, no code expressing them; every consumer wrote the loop.
+- "Did we lose anything, and which ones?" becomes a query that survives a restart — once
+  the loop of Decision 2 writes the records and a durable store keeps them. Neither is in
+  this change; both are what it is for.
+- The consume loop would land as a first-class thing. Rakaia has documented at-least-once
+  semantics and no code expressing them, so every consumer writes the loop itself.
 - `stage` makes the difference between a lost fact and an unapplied one legible at the
   point someone is deciding what to do about it.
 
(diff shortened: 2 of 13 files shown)


joshbrooks:
## Evidence

**Gates.** 2263 passed, 16 skipped, 1 xfailed. `just lint`, `fmt-check`, `typecheck` (0 errors; the 2 warnings are pre-existing on `__init__.py`), `docs` all clean.

**Three things the decision got wrong, each found by implementing it**

| # | What it said | Why it was wrong |
|---|---|---|
| 1 | `gaps()` returns "offsets below the cursor with no outcome recorded" | Under the same decision's exceptions-only rule that is *every event that worked*. Success writes nothing, so absence of a record is the normal case. Dropped. |
| 2 | `stage="append"` means "the fact is gone" | True of a lost append, false of a refusal — a gated consumer drops events on purpose and the fact is safe upstream. Now four cases keyed on `stage` × `status`. |
| 3 | The record carries `resolved_at` | The observations `reasons` names already own resolution. Two answers to one question eventually disagree. Now append-only and attempt-numbered. |

Item 1 stayed invisible until a contract test had to state an expected value — the two halves were in one paragraph and neither is wrong alone.

**A pre-existing bug this found, outside the scope of this PR**

`JsonlStreamStore` has the same path-escape the containment test caught in the new store. `quote(safe="")` does not touch a dot, so a stream named `..` resolves above the root:

```python
s = JsonlStreamStore(root, fsync=False)
s.create(".."); s.append("..", b"{}", AppendOptions())
# writes /tmp/<parent>/000000000000.jsonl — outside root
```

Fixed here as `jsonl_outcomes._safe_name`. Not applied to `jsonl_store._dir` because that file belongs to #242's area — happy to do it here, in its own PR, or as an issue, whichever you prefer.

**Mutations watched red**

| Target | Mutation | Result |
|---|---|---|
| `latest` | take last written, not highest attempt | red |
| `latest` | drop the consumer/stream filter | red *(initially survived — fixture gave three scopes one offset, so the answer collapsed either way)* |
| `latest` | sort offsetless entries first | red |
| `_safe_name` | drop the dot guard | red |
| `_file` | leave consumer unencoded | red *(initially survived — unencoded still round-trips; the test had to assert containment, not read-back)* |
| `_file` | leave stream path unencoded | red |
| `_read` | catch the wrong exception for a torn line | red |

Two of those survived their first version. Both were the same shape: asserting a property that holds either way.

**One test written and then deleted.** An "offsets are never parsed" ordering assertion. Both offset formats are zero-padded precisely so lexicographic and numeric order agree, so no ordering assertion can catch a store that parses one. That property lives in `test_cross_backend_cursors.py`, where a foreign format is refused.

**Open question the ADR names rather than closes.** A sequence key protects ordering *within* a sequence; a stage-2 reducer reads *across* them. So an event parked in one sequence can silently change a value derived in another, with no outcome saying so. Three candidate answers are listed under "What would reopen this" — it is a decision about reducers, not about outcomes.

joshbrooks:
## Correction to commit `3868635`'s message

That message says it pinned three unguarded fixes and that "all three could be undone with the suite green". The commit adds **two** tests, and the third item it names — tolerance for a line written by a later version — was already covered by a test added in the preceding commit `34b070e`. The same message's second paragraph then describes that case as the one that "looked unguarded and was not", so it contradicts itself as well as the diff.

Leaving the history rather than rewriting it: hiding a wrong claim is worse than correcting it in the open. Two other commit messages in this branch were also found to claim more than their diff did (`ed35bed` said an ADR passage was corrected when the edit had silently failed to match; corrected for real in `66f395b`). Three occurrences is a pattern in how these were written, not three accidents.

## Review loop summary

Five rounds, fresh reviewer each time, each round one commit.

| Round | Headline | In the previous round's fix? |
|---|---|---|
| 1 | `latest()` keyed on offset, so refusals collapsed to one row and kept the first | — |
| 2 | `params` stored by reference; the two backends disagreed | yes |
| 3 | `params` **keys** unvalidated — same defect, other half | yes |
| 4 | `reasons` had **both** defects and neither check | yes |
| 5 | five scalar fields declared `str` and never checked | yes |

Four of five rounds found their headline defect inside the previous round's fix, all of one class: a field that survives in memory and changes or raises when written, so the backends disagree about what was recorded. Round 5 was asked for an enumeration rather than a finding and produced both — the two container fields are now provably closed, and the enumeration is what turned up the scalar case.

Findings that corrected the author rather than the code:

- Three commit messages claimed more than their diff (above).
- Round 1's "the mutation was misplaced, the arm is unreachable" was half wrong: unknown fields cannot reach the `TypeError` arm, but a line missing a required field can, and nothing covered it.
- Two reviewer findings were overruled with evidence and the overrules were audited and upheld: `RowResult.failed_flags` is genuinely `tuple[str, ...]`, and `Refusal.params` is genuinely `dict[str, str]` — built in one place by a helper that renders every value, so a reviewer reading the *verdict's* `Mapping[str, Any]` had the wrong object.

Nine ADR claims about the downstream consumer were wrong and are corrected. The most consequential: that consumer records an outcome on success *and* failure so that a missing record means "never ran", which is the inverse of this design's Decision 3 — so the two surfaces cannot be mixed, and that is now a named negative rather than an assumption.

joshbrooks:
## Redesign, and a spike testing whether it worked

The five rounds above each closed one member of one class. `aa4cdb7` and `ef1fe1f` replace that with one translation every store goes through — and the check **walks each field's declared type instead of naming the fields**, because naming them is how it took five rounds: each check was written where the defect was found, so the list was never more complete than the last bug.

Reading the declaration also caught two cases no round had reported: a value outside a `Literal`'s domain (`stage=42` was accepted by both stores) and `bool` arriving where `int` was declared.

### Two corrections to my own work, both found by review

**The first attempt at the redesign was decorative.** Reverting the in-memory store to keeping the object left the entire suite green, because `__post_init__` validates whatever the store does. The commit message had claimed that store-level change *was* the structural fix. Both halves are now pinned — the JSONL half only after removing a second key-sorter that was quietly making the first redundant.

**The redesign orphaned one of its predecessors' guarantees.** With `encode` copying containers, the `params` copy in `__post_init__` became unobservable through any store. It is not redundant — `frozen=True` still promises something to anyone holding an outcome — so it now has a direct type-level test with no store involved.

### The spike: #244

A third backend is the falsifiable test of "is the class closed". Result:

| | outcome |
|---|---|
| Shape | **Closed.** The shared contract failed on the first run of the new backend, on a difference nobody had looked for, with no backend-specific test written. |
| Size | **Not closed, and not closable this way.** A `CharField(max_length=255)` is not a type. A 300-character stream path is kept by both current stores *and* by a database under SQLite, and refused only under Postgres. |

The second row is the more useful finding, and it is the same fault one level further out than the one this PR fixed: the suite's default database does not enforce lengths, so the divergence is invisible where tests usually run — exactly as the in-memory store was the permissive one before. Recorded as a named limit in the ADR; the implication is that the cross-backend comparison has to run somewhere the constraints are real, which is a Phase 3 concern rather than this PR's.

joshbrooks:
## Rebuilt: fifteen commits to five

Same tree, byte for byte — verified with `git diff` against the loop branch before force-pushing. The previous history was the story of the debugging; twelve of the fifteen were rounds, and three of those asserted the defect class was closed when a later round showed it was not.

| | |
|---|---|
| `a90c825` | the decision |
| `f8db4a4` | the record, and one translation to the shape a store keeps |
| `829adc6` | the contract every store must satisfy |
| `9f58fe0` | the file-backed store |
| `590f874` | the tests that compare the stores against each other |

Each is self-contained and green on its own, so a reviewer can stop after any of them.

## The design changed in the rebuild, and the reason matters

It now takes the approach from #245: **the record renders itself to text and every store keeps that exact text**, so there is one representation and nothing for two backends to disagree about. That replaces a hand-written type checker whose job was to *describe* which values survive storage — a set that turned out to be open-ended (a text subclass that flattens, a lone surrogate no path carries, a string longer than a column), which is why six rounds each found one more member of it.

`outcomes.py` is 66 lines shorter and three functions lighter; `json.dumps` does what `_matches`, `_declared` and `_encodable` were attempting.

**One correction to #245 before adopting it.** Dropping every type rule went too far. A number used as an offset is stored *identically* by every backend — no disagreement at all — and then makes `latest()` raise the moment a text offset sits beside it, because the two cannot be compared. Consistent and still broken. So a short list survives, and it is not a type rule: the four fields a store turns into **path segments, keys and sort keys** must be writable text, because the machinery reasons about them. Everything an outcome merely carries is settled by storing it.

## What the loop earned

Five rounds, and **four of the five found their headline defect inside the previous round's fix** — all one class. That count is the argument for having run a loop rather than a single pass, and it is why the sixth pass was a redesign instead of a seventh round.

Findings that corrected me rather than the code, since those are the ones re-reading my own work would never have caught: three commit messages claimed more than their diff; my "the mutation was misplaced" call in round 1 was half self-serving; my first redesign was decorative and I had claimed it was the structural fix; and the type checker skipped the two fields its own docstring named as its reason for existing.

joshbrooks:
Review loop, four rounds, fresh reviewer each round. Rounds 1 to 3 each produced one fix commit; round 4 found nothing blocking.

| Round | Commit | Headline finding | Defect in a previous round's fix? |
|---|---|---|---|
| 1 | 8c6b3b4 | JSONL `record` appended straight onto a torn tail, losing the next outcome too; `Outcome` validation was json-only, so a nested dict in `params`, `stage="bogus"`, `attempt=True` were all accepted while the docstring and ADR 6b claimed a per-field check; `decode` raised on a JSON line that is not an object; JSONL `latest` trusted the file name for scope | no |
| 2 | 95c45e6 | The new walk ran *after* the container copy, so `reasons="missing_total"` became thirteen one-letter codes; the lock docstring claimed no test could fail with the lock removed, and one can | yes, both |
| 3 | f95ccbc | Round 2's lock test never joined the second writer before the final read: 4 failures in 70 runs on a clean tree | yes |
| 4 | — | Nothing blocking | — |

Mutations recorded, all reverted and re-run on the committed tree. Red on the test named for it: torn-tail cut removed; scope filter removed; dict guard in `decode` removed; `_check_fields` returning empty (28 cases); Literal accepts any str; int accepts bool; dict branch skips keys; tuple branch skips elements; str encode check dropped; unconditional container copy (7 cases); `LOCK_EX` removed; flock moved after the torn-tail check; `LOCK_SH` for `LOCK_EX`; plus the ~20 pre-existing mutations round 1 and round 4 re-ran (`_safe_name`, empty names, stage/offset rules, `>=` attempt, `_order`, tier boundary).

Green, and left that way deliberately: `LOCK_UN` removed (closing the file releases the lock; the test docstring says so); the hook skipping the real torn-tail check (the lock test pins the lock window, not the check, and says so); `encode(self)` at the end of `__post_init__` replaced with `pass` (the walk already refuses everything `json.dumps` would, so it is a redundant backstop; kept as one, cost is one `json.dumps` per construction). One more note from round 4: if `record` ever stops calling the torn-tail check, the lock test fails with `RuntimeError: cannot join thread before it is started` rather than an assertion. Still red, just a less helpful message.

PR #244, opened 2026-09-06
Spike: a third outcome store, to test whether the class is closed

joshbrooks:
Not for merge. This exists to answer one question about #243: five review rounds each found one more value that one store kept and another refused, and the shared translation added since was meant to close that class by reading each field's declared type rather than naming the fields. A third store, carrying a limit neither of the others has, is the falsifiable test — do the shared suites catch it without anyone writing a case for it?

For shape, yes: the contract failed on the first run, on a difference nobody had looked for. For size, no — and the reason is worth more than the gap. A name longer than the column allows is kept happily by all three stores locally and refused only against the real database, because the one the suite uses by default does not enforce lengths at all. That is the same fault this whole thread has been about, one level further out, and it argues that the cross-backend comparison needs to run where the constraints are real rather than where they are ignored.


joshbrooks:
Closing — it did its job. The question it existed to answer was whether the shared suites catch a constraint a third backend has and the other two do not, without anyone writing a case for it. They did, on the first run.

Its two durable findings are kept elsewhere: the path escape is now #246, and the limit it measured — a length bound is not a type, and is invisible on the database the suite uses by default — is a named limit in the decision record merged with #243. Nothing here is worth carrying as a branch.

PR #245, opened 2026-09-06
Spike: one representation, instead of describing which values are allowed

joshbrooks:
Seven review findings across #243 were the same thing: a value one store kept and another altered or refused. Each fix was a better description of which values are allowed, and each description leaked, because the set it describes is open-ended — text that is a subclass of text, text no path can carry, text longer than a column allows. The last round found a bug in the description itself, in the two fields its own comment said it existed to protect.

This stops describing the set. The record renders itself to text, every store keeps that exact text, and whether something survives storage is settled by trying rather than by predicting — so the in-memory store is the file-backed one without the file rather than a second implementation that behaves similarly. Eighty-four lines and three functions go, including a hand-written type checker. The trade is real and worth arguing about before this is preferred to #243: a value of the wrong type that stores consistently is now accepted where it used to be refused, moving that check from run time to the type checker.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -89,6 +89,28 @@
 - Research notes live in `docs/research/` and are deliberately **not** in the
   nav. They are dated and are not decisions; decisions go in `docs/adr/`.
 
+## Writing it up
+
+- **Two plain paragraphs.** A commit message body, an issue body and a PR body are each
+  at most two paragraphs, written so someone who only reads the notification email gets
+  the whole point. Say what changed and what it means for whoever uses this. That is the
+  deliverable.
+- **Everything else goes in a comment.** Tables, measured counts, SHAs, query output,
+  repro steps, mutation records, per-file reasoning — post them below the body, on the PR.
+  They are evidence for whoever verifies the work, not the summary for whoever reads it. A
+  commit message has nowhere to put them, so they belong on its PR instead.
+- **No machinery in the body.** This library's own words — stream, event, consumer,
+  cursor, offset, replay, backend — are the vocabulary and are fine. Identifiers are not:
+  `DjangoExecutor.apply` is "the part that writes a batch", `_StageBuffer` is "batching a
+  pass together". Save the precise names for the comment, where precision is the job.
+- The test: read the body alone. If it needs the source open, or the point arrives after
+  the evidence, rewrite it.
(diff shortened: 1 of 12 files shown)


joshbrooks:
## What each approach does with the values that caused the seven findings

Both branches at their current heads. "refused" = at construction; "agree" = both stores read back identically.

| value | #243 (describe the set) | #245 (one representation) |
|---|---|---|
| `stage` as a `str` subclass | refused — *only after round 6 fixed the `Literal` branch* | both flatten to text, agree |
| lone surrogate in a name | refused — *only after round 6* | refused |
| non-text name (`42`) | refused | refused |
| `uuid` subject | refused | refused |
| `offset=42` | refused | **accepted**, both store `42` |
| nested dict in `params` | refused | **accepted**, both store it |
| 300-char name (Postgres) | not caught — see #244 | not caught — same |

The last two rows are the trade. #245 accepts values whose *type* is wrong when both stores keep them identically, because it only enforces what storage cares about; type correctness moves to pyright, which checks the call sites in-repo. The bottom row is unchanged: a column's length bound is not reachable either way.

## The argument

The first four rows are the point. Under #243 they were closed one at a time over five rounds plus a sixth, and the sixth found the type checker failing in exactly the two fields its own docstring named. Under #245 they are closed because there is nothing to disagree about — not because the code knows those cases.

`src/rakaia/outcomes.py` goes 363 → 297 lines; `jsonl_outcomes.py` loses its own JSON handling; `_matches`, `_declared` and `_encodable` all go. What replaces them is `json.dumps`.

## What is still true in both

The remaining checks are about *meaning*, not storage: a subject that names nothing, an appended outcome carrying an offset, an attempt below one. Plus one genuine storage rule the payload codec cannot see — `consumer`, `stream_path` and `subject` become path segments and dictionary keys, so they have to be writable text. One rule in one place, for one reason.

Not merged into #243. Worth a decision on the trade first.

joshbrooks:
Superseded by #243, which took this approach — one shared rendering every store keeps — with one correction made before adopting it.

Dropping every type rule went too far. A number used as an offset is stored identically by every backend, so nothing disagrees, and then makes the read fail the moment a text offset sits beside it because the two cannot be compared. Consistent and still broken. What landed keeps a short rule for the four fields a store turns into path segments, keys and sort keys, and settles everything an outcome merely carries by storing it.

PR #247, opened 2026-09-06
A stream called .. no longer writes outside the store

### LINKED ISSUE #246: A stream called .. writes outside the store root
`JsonlStreamStore` maps a stream name to one directory with `quote(name, safe="")`, which escapes a slash and leaves a dot alone. So a stream called `..` resolves above the root and writes outside the store entirely. Reproduces on `main` today:

```python
s = JsonlStreamStore(root, fsync=False)
s.create(".."); s.append("..", b"{}", AppendOptions())
# writes <root>/../000000000000.jsonl
```

Found while writing a containment test for the outcome store in #243, which hit the identical hole and fixed it on its own side. Worth saying why review had not caught it: an unencoded name still *round-trips*, so a test that reads back what it wrote passes while the file sits somewhere it should never have been. The property that catches it is containment — every file directly under the root — not read-back.

The fix on the outcome side is `_safe_name` in `src/rakaia/outcomes.py`: percent-encode a name that is nothing but dots, and an empty one, since only those can traverse. `jsonl_store._dir` wants the same treatment, and the two should share it rather than each carry a copy — the same argument that put `_discard_torn_tail` at module scope.

joshbrooks:
The file-backed stores turned a stream or consumer name into a folder name by escaping anything that looked like a path separator. A name made only of dots slipped through that, so a stream called `..` put its data in the folder above the store instead of inside it. It went unnoticed because the data still read back correctly — the only thing wrong was where it sat.

Both stores now work the other way round: build the intended location, then check it really is inside the store, and only fall back to further escaping when the check says otherwise. One rule in one place, instead of each store keeping its own list of names it thinks are dangerous. New tests put a range of hostile names through both stores and confirm every file lands where it should. Closes #246.

### DIFF
--- a/src/rakaia/jsonl_outcomes.py
+++ b/src/rakaia/jsonl_outcomes.py
@@ -24,9 +24,10 @@ Layout, one file per consumer and stream::
 
     <root>/<quoted consumer>/<quoted stream path>.jsonl
 
-`quote(safe="")` on both components for the reason `JsonlStreamStore._dir` gives:
-a stream path contains slashes and must map to one name, not a tree, and
-percent-encoding is the mapping that is reversible.
+Both components go through `JsonlStreamStore`'s `_contained`, which maps a name
+to one child of a directory and then checks that it landed there — the same
+single rule the stream store uses, rather than each store carrying its own idea
+of which names are dangerous.
 
 Sharing the stream store's limits, and for the same reasons: `fcntl` only, so no
 Windows; `flock` is unreliable on NFS; readers do not take the lock. A torn
@@ -47,36 +48,14 @@ from __future__ import annotations
 
 import os
 from pathlib import Path
-from urllib.parse import quote
 
-from .jsonl_store import _discard_torn_tail
+from .jsonl_store import _contained, _discard_torn_tail
 from .outcomes import Outcome, _order, decode, encode
 
 try:  # pragma: no cover - platform dependent
     import fcntl
 except ImportError:  # pragma: no cover - Windows
     fcntl = None  # type: ignore[assignment]
-
-
-def _safe_name(name: str) -> str:
-    """One path segment for `name`, guaranteed to stay under its parent.
-
-    `quote(safe="")` maps a slash to `%2F`, which is what keeps a stream path a
-    name rather than a tree, and it is reversible. What it does **not** touch is a
-    dot: `quote("..")` is `".."`, so a consumer or stream called `..` resolves
-    above the root and writes outside the store entirely. Only an all-dot name can
-    do that — anything else already carries an encoded character — so those are the
-    ones encoded here, and an empty name, which would otherwise collapse a
-    directory level.
-
-    Found by a containment test rather than by review, and that is the point worth
-    keeping: an unencoded name still *round-trips*, so reading back what was
-    written passes while the file sits somewhere it should never have been.
-    """
-    quoted = quote(name, safe="")
-    if not quoted or set(quoted) <= {"."}:
-        return quoted.replace(".", "%2E") or "%00"
-    return quoted
 
 
 class JsonlOutcomeStore:
@@ -101,7 +80,7 @@ class JsonlOutcomeStore:
         self.root.mkdir(parents=True, exist_ok=True)
 
     def _file(self, consumer: str, stream_path: str) -> Path:
-        return self.root / _safe_name(consumer) / f"{_safe_name(stream_path)}.jsonl"
+        return _contained(_contained(self.root, consumer), stream_path, ".jsonl")
 
     def record(self, outcome: Outcome) -> None:
         target = self._file(outcome.consumer, outcome.stream_path)
--- a/src/rakaia/jsonl_store.py
+++ b/src/rakaia/jsonl_store.py
@@ -121,6 +121,48 @@ def _discard_torn_tail(fh: Any) -> None:
     fh.truncate(cut)
 
 
+def _contained(root: Path, name: str, suffix: str = "") -> Path:
+    """The path `name` maps to, *proven* to sit directly under `root`.
+
+    A caller-supplied name — a stream path, a consumer id — becomes exactly one
+    child of `root`. `quote(safe="")` does the mapping, because a stream path
+    contains slashes and a stream is one directory rather than a tree: `/a/b`
+    and `/a%2Fb` must not collide, and percent-encoding is the one mapping
+    `list_paths` can reverse.
+
+    What makes it safe is not the encoding, though. `quote` leaves a dot alone,
+    so a stream called `..` used to resolve *above* the root and write outside
+    the store entirely. The tempting repair is to extend the escaping until
+    every dangerous name is covered, and that is the losing move: the bad set is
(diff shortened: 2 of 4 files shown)


joshbrooks:
### Reproduction, before the change

```
files under base: ['.lock', '000000000000.jsonl', 'activity', 'meta.json', 'store', 'store/%retired']
```

`store/` is the root; four of those files are its siblings. After the change everything is under `store/%2E%2E/`, and `list_paths()` still reports `..`.

### The change

| File | What |
| --- | --- |
| `src/rakaia/jsonl_store.py` | new module-scope `_contained(root, name, suffix="")` — quote, `normpath`, assert the result's parent is the root; re-encode dots (and `%00` for empty) and re-check only if it is not; raise if it still escapes. `_dir` and `_retired_path` now go through it. |
| `src/rakaia/jsonl_outcomes.py` | `_safe_name` deleted; `_file` is `_contained(_contained(root, consumer), stream_path, ".jsonl")`. |
| `tests/test_rakaia/test_jsonl_containment.py` | new; containment over `..`, `../escape`, `../../escape`, `a/b`, `a%2Fb`, `%2e%2e`, `%252e%252e`, `.`, `""` — for the stream store, and for the outcome store in either name component. Plus a no-collision test and a `list_paths` reversibility test. |
| `tests/test_rakaia/test_jsonl_outcome_store_contract.py` | its outcome-only containment test moved into the new file, so the shared rule has one test home. |

Deliberately lexical rather than `Path.resolve()`: the encoded name can never contain a separator, so `normpath` is the whole of the canonicalisation, and `_dir` is on every append.

### Mutation runs

Each mutation applied to `_contained` or its callers, then `pytest tests/test_rakaia/test_jsonl_containment.py`:

| Mutation | Result |
| --- | --- |
| containment check never fires (return the first candidate) | 6 failed — both stores' containment tests, plus reversibility |
| `_dir` reverts to `self.root / quote(path, safe="")` | 4 failed — stream containment for `..`, `.`, `""`, plus reversibility |
| empty-name `or "%00"` fallback dropped | 1 failed — stream containment for `""` (raises `ValueError`) |
| repair maps every escaping name to one segment | 2 failed — the no-collision and reversibility tests |

### Gates

`uv run pytest tests/ -q`: 2419 passed, 16 skipped, 1 xfailed (baseline 2400 + 27 new − 8 moved). `just lint`, `just fmt-check` clean; `just typecheck` 0 errors, the 2 pre-existing `__init__.py` warnings; `just docs` builds with no issues. Postgres skipped.

joshbrooks:
### Review finding on the docstring: option 1, the limit is now named

Confirmed the repro on the branch — a symlink inside the root sends `meta.json`, the segment, `.lock` and `activity` into `elsewhere/`, and the parent check passes because `normpath` never sees the link. The docstring did claim the strong property while implementing the weaker one. Fixed by stating the limit rather than by switching to `resolve()`, for the reasons in the review: the cost lands on every append, and the root's contents are the deployer's while only the *name* is caller-supplied.

Three places now say so:

| Where | What |
| --- | --- |
| `_contained` docstring | "the whole of the canonicalisation" replaced with the actual property — lexical containment, which covers every spelling of a name but not a symlink planted in the root — plus the trade and its reason. |
| `jsonl_store` module docstring | the symlink added to the "what it deliberately does not do, as shipped" list, beside Windows and NFS. |
| `test_jsonl_containment.py` | new `test_a_symlink_inside_the_root_does_not_lead_a_write_out_of_it`, `xfail(strict=True)`, in the house style of `test_architecture_spikes.py`. |

The xfail asserts the strong reading and accepts either honest answer — the write contained, or the name refused (`ValueError` is suppressed, then `outside` must be empty) — so a future fix of either shape flips it.

**Verified strict in both directions.** As shipped: `1 xfailed`. With `_contained` switched to `Path.resolve()`: `FAILED … [XPASS(strict)]` — the marker really would come off the day someone resolves. Worth recording that `resolve()` *refuses* rather than contains, since `root/link` resolves outside the base and no re-encoding rescues it; the xfail reason says so.

Gates after the change: `uv run pytest tests/ -q` 2419 passed / 16 skipped / **2 xfailed** (the new one, plus the pre-existing one); `just lint`, `just fmt-check` clean; `just typecheck` 0 errors with the 2 known `__init__.py` warnings; `just docs` clean.

joshbrooks:
### The XPASS reproduces — exact diff, command and output

Re-run just now at `642decc`, in the worktree `../rakaia.worktrees/path-escape`, working tree otherwise clean.

The diff, which is both `normpath` call sites plus the base:

```diff
--- a/src/rakaia/jsonl_store.py
+++ b/src/rakaia/jsonl_store.py
@@ -162,15 +162,15 @@ def _contained(root: Path, name: str, suffix: str = "") -> Path:
-    base = Path(os.path.normpath(root))
+    base = Path(root).resolve()
     encoded = quote(name, safe="")
-    candidate = Path(os.path.normpath(base / (encoded + suffix)))
+    candidate = Path(base / (encoded + suffix)).resolve()
     if candidate.parent == base:
         return candidate
     # Outside, so re-encode and re-check. `%00` covers the empty name, which
     # would otherwise collapse a level away rather than escape one.
     encoded = encoded.replace(".", "%2E") or "%00"
-    candidate = Path(os.path.normpath(base / (encoded + suffix)))
+    candidate = Path(base / (encoded + suffix)).resolve()
     if candidate.parent == base:
         return candidate
     raise ValueError(f"name {name!r} does not map to a path under {root}")
```

```
$ uv run pytest tests/test_rakaia/test_jsonl_containment.py -q -rX
...........................F                                             [100%]
________ test_a_symlink_inside_the_root_does_not_lead_a_write_out_of_it ________
[XPASS(strict)] Known limit, named rather than discovered (#246 review): ...
1 failed, 27 passed in 0.24s
```

And the direct call, under the same mutation:

```
module file: .../rakaia.worktrees/path-escape/src/rakaia/jsonl_store.py
body: ['base = Path(root).resolve()',
       'candidate = Path(base / (encoded + suffix)).resolve()',
       'candidate = Path(base / (encoded + suffix)).resolve()']
_contained raised ValueError: name 'link' does not map to a path under /tmp/tmp_u89kux9/streams
```

Mutation reverted; the tree is back to `27 passed, 1 xfailed`, `git status` clean.

### Why the placement probably missed

`_contained(root, "link")` returning `<root>/link` **unresolved** is not something the mutated body can do — `Path.resolve()` follows an existing symlink unconditionally. So the code that ran was not the code that was edited. This repo currently has at least two trees with their own editable installs pointing at themselves: the main checkout, whose `.venv/…/direct_url.json` reads `file:///home/josh/github/joshbrooks/rakaia`, and this worktree, which reads `…/rakaia.worktrees/path-escape`. Edit in one, run in the other, and the mutation is invisible — silently, with no import error, as long as both trees have the symbol.

The cheap guard, and the one that would have caught all three misplaced mutations this session: before reading a mutation's result, make the process prove it is running the mutated code.

```python
import inspect, rakaia.jsonl_store as m
print(m.__file__)                       # which tree
print(inspect.getsource(m._contained))  # which body
```

A mutation that does not change the printed body has not been placed, and whatever the suite says next is about the unmutated code.

PR #248, opened 2026-09-06
Give a consumer one written loop instead of each writing its own

joshbrooks:
Reading a stream and applying it was four steps every consumer wrote by hand, and the order of those steps decides two things that are easy to get wrong. Recording what happened to an event has to happen after the work, not inside it, or a batch that fails takes the record of its own failure down with it. And the watermark has to move last, one event at a time, or an event nobody applied ends up below it and reads back as an event that worked. There is now one version of that order, and the caller says up front what should happen when an event fails: keep going, which is what a live consumer wants, or stop, which is what a rebuild wants. There is no default, because the two want opposite things and picking one for them is how the wrong one gets used in silence.

The work being applied can also hand back a note of its own rather than raise — that is groundwork for a later change where a value is computed from an incomplete set and someone should be told. Three tests cover the three properties above, and the decision record has been updated: the part of it that said this was only proposed no longer says so. Nothing new is exported, so the surface is unchanged.

### DIFF
--- a/docs/adr/0007-an-outcome-is-recorded-where-the-cursor-is-committed.md
+++ b/docs/adr/0007-an-outcome-is-recorded-where-the-cursor-is-committed.md
@@ -99,13 +99,25 @@ backend, so "does this work on JSONL?" is a red test rather than a question.
 then commits the cursor. Nothing is recorded from inside an `Effect`, an executor, or a
 handler.
 
-Two details of it are the decision rather than the implementation. The commit is **per
-message**, because that is the only granularity at which "the cursor stopped below the event
-that failed" is a true statement — a batch commit puts an unapplied event below the
-watermark, which under Decision 3 reads as an event that worked. And the loop must not be
-called from inside a transaction of the caller's own, or the rollback this decision exists
-to escape swallows the record after all; that is a constraint on the caller, and it is
-stated in the docstring because nothing can enforce it.
+Two details of it are the decision rather than the implementation, and the first is a trade
+rather than a mechanism.
+
+**The commit is per message, and that is a cost accepted knowingly.** It is one durable
+cursor write per event where a batch commit would be one per poll, and a consumer at volume
+will feel the difference. What it buys is the only granularity at which "the cursor stopped
+below the event that failed" is a true statement: commit a batch and an unapplied event sits
+below the watermark, which under Decision 3 does not read as unapplied — it reads as an
+event that worked. So the choice is between a write per event and a mode (`halt`) that
+cannot mean what it says, and this decision takes the write. A consumer that measures this
+as its bottleneck is asking for batch commit back, and the answer is not "make it
+configurable" — it is that `skip` tolerates a coarser commit and `halt` does not, so the
+split would be per mode, like `on_error` itself. Nobody has needed it yet, so it is not
+built.
+
+**The loop must not be called from inside a transaction of the caller's own**, or the
+rollback this decision exists to escape swallows the record after all. That is a constraint
+on the caller, stated in the docstring because nothing in a stdlib-only module can enforce
+it; it is measured and cross-referenced in the Consequences below.
 
 An apply may also **return** outcomes rather than raise, and they are recorded on the same
 path. That is what a reducer needs to say a value was computed from a population with a hole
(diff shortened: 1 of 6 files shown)


joshbrooks:
## Evidence

### API added — `src/rakaia/subscription.py` (protocol-server tier, no new module, tier map unchanged)

```python
OnErrorPolicy = Literal["skip", "halt"]

@dataclass(frozen=True)
class Consumed:
    status: PollStatus
    applied: int
    outcomes: tuple[Outcome, ...]
    cursor: str | None
    halted: bool

def consume(
    store: CursorStore,
    path: str,
    apply: Callable[[StreamMessage], Iterable[Outcome] | None],
    *,
    consumer: str,
    on_error: OnErrorPolicy,          # no default
    cursor: str | None = None,
    commit: Callable[[str], None] | None = None,
    outcomes: OutcomeStore | None = None,
    subject_of: Callable[[StreamMessage], str] | None = None,
    sequence_of: Callable[[StreamMessage], str] | None = None,
) -> Consumed: ...
```

Nothing added to `rakaia.__all__` — ADR 0007 keeps the export for when it is Accepted — so `api-reference` is unchanged and was not regenerated.

### Gates

| Gate | Test | File |
|---|---|---|
| The outcome survives the executor's rollback | `test_the_outcome_survives_the_batch_it_records` | `tests/test_django_rakaia/test_consume_rollback.py` |
| Committing before applying loses nothing | `test_the_cursor_does_not_advance_past_an_event_that_did_not_apply` | `tests/test_rakaia/test_consume_loop.py` |
| `halt` stops where `skip` continues | `test_halt_stops_where_skip_continues` | `tests/test_rakaia/test_consume_loop.py` |

The rollback gate runs against a real `DjangoExecutor` and a real `transaction.atomic`, with an outcome store that writes an actual row, because neither reference store has a transaction to roll back. It asserts both halves at once — the batch is discarded (`Measure.objects.count() == 0`, including the effect that had already succeeded) and the record is present. A second test in the same file writes the record from *inside* the atomic and shows it is gone, which is the design that was rejected.

### Mutations

| # | Mutation | Test that went red | Result |
|---|---|---|---|
| 1 | Drop the record on the failure path | `test_the_outcome_survives_the_batch_it_records` (+2) | killed |
| 1b | Drop the `atomic` from the contrast test — is the store really transactional? | `test_recording_inside_the_executor_loses_the_record` | killed (gate is not vacuous) |
| 2 | Commit the cursor before applying | `test_the_cursor_does_not_advance_past_an_event_that_did_not_apply` (+4) | killed |
| 3 | `halt` behaves like `skip` | `test_halt_stops_where_skip_continues` (+3) | killed |
| 4 | Give `on_error` a default of `"skip"` | `test_on_error_has_no_default` | killed |
| 5 | `skip` does not advance the cursor | `test_halt_stops_where_skip_continues` | killed |
| 6 | Outcomes returned by `apply` are dropped | `test_outcomes_the_apply_returns_are_recorded` | killed |

### Suite

`uv run pytest tests/ -q` — **2412 passed, 16 skipped, 1 xfailed** (baseline 2400 + the 12 added here). `just lint`, `just fmt-check` and `just docs` clean; `just typecheck` 0 errors, 2 pre-existing `__init__.py` warnings. Postgres not run.

### ADR 0007 changes

- "What is built, and what this decision only proposes" — Decisions 2, 3 and 5 move into what is built; what remains outstanding is the Django backend and the export.
- Decision 2 — drops the "not built" marker; adds the two things that are decisions rather than implementation (the commit is per message; the caller must not wrap the loop in its own transaction) and the returned-outcome path.
- Decision 5 — drops the marker; states that there is no default value at all, only which one each mode passes.
- Consequences — the "survives a restart … neither is in this change" entry now says the loop writes the records and the file-backed store keeps them, with a Django place to keep them still outstanding. The "nothing enforces that the loop is used" entry gains the caller-transaction hole.
- "What would reopen this" — option (b) notes the loop already carries a reducer's outcome; what is undecided is the reducer half.

joshbrooks:
## Follow-up (7f8c0ed) — the caller-transaction hole, measured

Reproduced independently on this branch before writing it down, and it matches what you measured:

```
records written inside the block: 1
records surviving the rollback:   0
projection rows:                  0
```

Shape: caller `atomic()`, raising handler, `on_error="skip"`, `DatabaseOutcomeStore`, caller raises to roll back. The probe was a throwaway — not committed, because the number belongs to the guard being built in the Django outcome store, and a test here would pin the defect in the wrong repo tier and go red the moment that guard lands.

### What changed

**ADR 0007 — the negative is now its own entry with the count.** Previously a trailing sentence on "nothing enforces that the loop is used", which read as the same problem. It is not: that one is about a consumer declining the supported path, this one is about a consumer *using* it correctly and still losing the record. Split out, with the table above, and it states why it cannot be fixed in `consume` — `rakaia` is stdlib-only, so it cannot see a Django atomic block, and a core module that could would be the crossing `tests/test_rakaia/test_tier_boundary.py` refuses. Cross-referenced to the unbuilt Django backend as the same finding, to be closed with it rather than separately.

**ADR 0007 Decision 2 — the per-message commit now reads as the trade.** You were right that it read as mechanism. It now leads with the cost (one durable cursor write per event where a batch commit is one per poll), says what that buys, and names the shape a future revisit would take — the split would be per mode, since `skip` tolerates a coarser commit and `halt` cannot, so the answer is not a `batch_commit=` flag. Nobody has needed it, so it is not built.

**`consume`'s docstring** carries the same number and says plainly that it is a constraint on the caller, not a guard.

### The no-default pin

You were right to check. The old test raised `TypeError` on a missing keyword — which *any* missing required argument produces, and which says nothing about a default being added. Replaced with two:

| Test | What it holds |
|---|---|
| `test_on_error_has_no_default` | reads `inspect.signature(consume).parameters["on_error"]` — asserts `.default is Parameter.empty` and `.kind is KEYWORD_ONLY` |
| `test_omitting_on_error_is_an_error_at_the_call` | `pytest.raises(TypeError, match="on_error")` — the absence is felt at the call site, and the match pins *which* argument |

New mutations, both killed:

| # | Mutation | Test that went red |
|---|---|---|
| 4a | `on_error: OnErrorPolicy = "skip"` (keyword-only kept) | both of the above |
| 4b | `on_error` moved before the `*` (positional, still required) | `test_on_error_has_no_default` |

4b is the one the old test could not see at all. Note that under 4a every other test in the file still passes — which is exactly the drift Decision 5 is about, and why the signature is now read directly.

### Gates

`uv run pytest tests/ -q` — **2413 passed, 16 skipped, 1 xfailed** (one more than before: the split no-default test). `just lint`, `just fmt-check`, `just docs` clean; `just typecheck` 0 errors, 2 pre-existing warnings.

PR #249, opened 2026-09-06
Keep a consumer's outcomes in the database

joshbrooks:
An event a consumer could not apply can now be recorded in a table, alongside the
cursor that says how far that consumer got. It joins the two places outcomes could
already be kept — in memory, and in files on disk — and all three write and read the
same text, so none of them can develop its own idea of what was stored. There is a
migration to run; nothing writes to the table unless a consumer asks for it, and a
consumer that never fails leaves it empty.

Recording never refuses. Whatever a consumer names its streams and subjects, of any
length and any character, the record is kept whole — which matters because the code
that will write these records does so while already handling a failure, and a store
that could refuse would turn one bad event into a stopped stream. The columns beside
the record are an index for finding a row rather than a copy of the values, which is
what makes that possible: they hold the same encoded form the file-backed store
already uses to turn a name into something a filesystem will accept.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -39,6 +39,13 @@ else an outcome carries is unbounded. If your names can run longer than that,
 they will not fit, and you will see the refusal when the outcome is recorded
 rather than when it is read back.
 
+One more thing to know if you wrap your own consuming in a transaction. By
+default the store writes on whichever connection is already open, so a record
+written inside a block that later rolls back is rolled back with it — the record
+of the failure goes down with the failure. Pass `using=` a database alias your
+own transaction does not cover if you need the record to survive that; there is a
+test stating both numbers.
+
 ---
 
 # 0.3.0
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -15,6 +15,32 @@ ones you are crossing.
 
 ---
 
+# Unreleased
+
+## A new table for the outcomes a consumer records
+
+There is a new migration, `django_rakaia` `0010`, adding one table:
+`rakaia_consumeroutcome`. Run `manage.py migrate django_rakaia` when you upgrade.
+
+Nothing existing changes and nothing writes to the table on its own. It is where
+`django_rakaia.outcomes.DjangoOutcomeStore` keeps a record of an event a consumer
+could not apply — the third place outcomes can be kept, after the in-memory
+reference and the JSONL files — and only a consumer that asks for that store ever
+puts a row in it. A consumer that never fails, or never uses the store, carries an
+empty table and nothing else; outcomes record exceptions only, so there is no row
+per applied event.
+
+One thing to know before you point a consumer at it. Four of the values are kept
+in bounded columns, at the same widths `ConsumerCursor` already uses — the
+consumer name at 128 characters, the stream path and the subject at 255, the
+offset at 64 — and the store refuses an outcome whose name is longer rather than
+shortening it, because a shortened subject is a different subject. Everything
+else an outcome carries is unbounded. If your names can run longer than that,
+they will not fit, and you will see the refusal when the outcome is recorded
+rather than when it is read back.
+
+---
+
 # 0.3.0
 
 ## A replay applies a whole pass of effects at once, not one event at a time
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -37,10 +37,12 @@ wrong.
 
 The one thing to know if you plan to read the table directly rather than through
 the store. The `payload` column is the record: it holds the whole outcome as text,
-and `rakaia.outcomes.decode` turns it back into one. The columns ending in `_key`
-are an index over that text, not a copy of it — each holds a percent-encoded,
-possibly shortened form of the value, so `stream_path_key` for `submission/tf611`
-reads `submission%2Ftf611`. Query them to find rows; read the payload for values.
+and `rakaia.outcomes.decode` turns it back into one. The two columns ending in
+`_key` are the scope index over that text, not a copy of it — each holds a
+percent-encoded, possibly shortened form of the value, so `stream_path_key` for
+`submission/tf611` reads `submission%2Ftf611`. Use them to find the rows for a
+consumer and a stream; read everything else, the subject and offset included, out
+of the payload.
 
 One more thing to know if you wrap your own consuming in a transaction. By
 default the store writes on whichever connection is already open, so a record
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -30,14 +30,17 @@ puts a row in it. A consumer that never fails, or never uses the store, carries 
 empty table and nothing else; outcomes record exceptions only, so there is no row
 per applied event.
 
-One thing to know before you point a consumer at it. Four of the values are kept
-in bounded columns, at the same widths `ConsumerCursor` already uses — the
-consumer name at 128 characters, the stream path and the subject at 255, the
-offset at 64 — and the store refuses an outcome whose name is longer rather than
-shortening it, because a shortened subject is a different subject. Everything
-else an outcome carries is unbounded. If your names can run longer than that,
-they will not fit, and you will see the refusal when the outcome is recorded
-rather than when it is read back.
+Recording never refuses. Whatever your consumer names its streams and subjects —
+any length, any character — the record is kept whole, so a bad name cannot turn
+into a failed write on the path whose job is to record that something already went
+wrong.
+
+The one thing to know if you plan to read the table directly rather than through
+the store. The `payload` column is the record: it holds the whole outcome as text,
+and `rakaia.outcomes.decode` turns it back into one. The columns ending in `_key`
+are an index over that text, not a copy of it — each holds a percent-encoded,
+possibly shortened form of the value, so `stream_path_key` for `submission/tf611`
+reads `submission%2Ftf611`. Query them to find rows; read the payload for values.
 
 One more thing to know if you wrap your own consuming in a transaction. By
 default the store writes on whichever connection is already open, so a record
--- a/docs/public-api.md
+++ b/docs/public-api.md
@@ -45,8 +45,8 @@ is why the package exported nothing for so long.
 ### Tier 2 — Provisional: the ORM models and the database schema
 
 `django_rakaia.models` — `Stream`, `StreamEvent`, `StreamEntry`,
-`StreamProducer`, `StreamOffsetWatermark`, `ConsumerCursor` — plus the table
-shape behind them.
+`StreamProducer`, `StreamOffsetWatermark`, `ConsumerCursor`, `ConsumerOutcome`
+— plus the table shape behind them.
 
 **These are usable, and deliberately not in `__all__`.** Importing them from
 `django_rakaia.models` keeps the weaker guarantee visible at the import site
--- a/src/django_rakaia/migrations/0010_consumeroutcome.py
+++ b/src/django_rakaia/migrations/0010_consumeroutcome.py
@@ -7,6 +7,10 @@ class Migration(migrations.Migration):
     A record of what happened to an event a consumer could not apply, kept beside
     the cursor that says how far it got. Empty for a consumer that never fails —
     only exceptions are recorded — so the cost of having it is the table itself.
+
+    ``payload`` is the record. The ``_key`` columns are a percent-encoded index
+    over it, cut to the width, so no consumer-supplied name can be a value the
+    column refuses; see the model for why a cut key is still a correct index.
     """
 
     dependencies = [
@@ -26,11 +30,10 @@ class Migration(migrations.Migration):
                         verbose_name="ID",
                     ),
                 ),
-                ("consumer", models.CharField(max_length=128)),
-                ("stream_path", models.CharField(max_length=255)),
-                ("subject", models.CharField(max_length=255)),
-                ("offset", models.CharField(blank=True, max_length=64, null=True)),
-                ("attempt", models.PositiveIntegerField()),
+                ("consumer_key", models.CharField(max_length=128)),
+                ("stream_path_key", models.CharField(max_length=255)),
+                ("subject_key", models.CharField(max_length=255)),
+                ("offset_key", models.CharField(blank=True, max_length=64, null=True)),
                 ("payload", models.TextField()),
                 ("recorded_at", models.DateTimeField(auto_now_add=True)),
             ],
@@ -38,7 +41,7 @@ class Migration(migrations.Migration):
                 "db_table": "rakaia_consumeroutcome",
                 "indexes": [
                     models.Index(
-                        fields=["consumer", "stream_path"],
+                        fields=["consumer_key", "stream_path_key"],
                         name="rakaia_outcome_scope_idx",
                     )
                 ],
--- a/src/django_rakaia/migrations/0010_consumeroutcome.py
+++ b/src/django_rakaia/migrations/0010_consumeroutcome.py
@@ -8,9 +8,11 @@ class Migration(migrations.Migration):
     the cursor that says how far it got. Empty for a consumer that never fails —
     only exceptions are recorded — so the cost of having it is the table itself.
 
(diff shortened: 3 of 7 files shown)


joshbrooks:
## Evidence

### Schema — `rakaia_consumeroutcome` (migration `django_rakaia/0010`)

| column | type | why |
|---|---|---|
| `consumer` | `varchar(128)` | `ConsumerCursor.consumer_id`'s width — a consumer that can hold a cursor can hold an outcome |
| `stream_path` | `varchar(255)` | `ConsumerCursor.stream_path`'s width |
| `subject` | `varchar(255)` | no precedent; a name of the same order as a stream path, so the same width |
| `offset` | `varchar(64)`, nullable | `ConsumerCursor.offset`'s width. NULL for an append-stage outcome — an event that never reached the log has no position |
| `attempt` | `PositiveIntegerField` | highest wins in `latest` |
| `payload` | `TextField` | the whole outcome as `rakaia.outcomes.encode` rendered it |
| `recorded_at` | `DateTimeField(auto_now_add)` | |

Index on `(consumer, stream_path)`. **No unique constraint** — outcomes are
append-only (ADR 0007 Decision 6a), so a repeat is a new row and `latest` takes the
last one written. A constraint would turn an append into an error on the one path
whose job is to record that something already went wrong.

`sequence_key`, `stage`, `status`, `reasons` and `params` are **not** columns. They
live in the payload only, which is Decision 6b taken literally: the payload is the
record, the columns are an index over it so `latest` can filter and order in SQL. A
field added to `Outcome` needs no migration here. `latest` re-checks the decoded
payload's scope and lets it win over the columns, as the JSONL store does.

### What a too-long value does

Bounded columns are checked **in Python**, against the model's own declared widths
(read from the field, so the check cannot drift from the column), and refused with a
`ValueError` naming the field, the length and the limit. Nothing is written.

Without that check, the two databases disagree — measured on this branch with the
check removed, recording a 300-character `stream_path`:

```
SQLITE:   accepted, stream_path stored at 300 characters
POSTGRES: refused by the database -> DataError: value too long for type character varying(255)
```

That is ADR 0007's own forecast ("a backend with a bounded column accepts a name
every other backend keeps and then refuses or truncates it… the suite's default
database does not enforce lengths, so that divergence is invisible where the tests
usually run") reproduced. Refusing rather than truncating, because a truncated
subject is a *different* subject: it would collapse two rows into one and report the
first as speaking for the rest, which is the defect `subject` was introduced to fix.

`sequence_key`, `reasons` and `params` are unbounded and are pinned as unbounded, so
the cheap fix for the above — bound everything — cannot be applied without a red test.

### Both test runs

| | result |
|---|---|
| `uv run pytest tests/ -q` (SQLite) | **2434 passed, 16 skipped, 1 xfailed** (baseline 2400/16/1) |
| `just test-pg` — `RAKAIA_TEST_DB=postgres`, port 55432 | **2448 passed, 2 skipped, 1 xfailed** |

The contract passes on both. `just lint`, `just fmt-check`, `just docs` clean;
`just typecheck` 0 errors, 2 warnings (both pre-existing `__init__.py` `__all__`
warnings). `just api-reference` regenerates unchanged — nothing new is exported.

### The caller's-transaction hole

Raised against this branch while it was in flight: PR #248's consume loop writes the
outcome outside the executor's transaction, but a caller that wraps `consume(...)` in
its own `atomic()` and rolls back takes the record with it.

**Chosen: option 1 (a separate connection), as an opt-in `using=` rather than the
default.** `DjangoOutcomeStore(using="…")` writes and reads on a named alias, the same
seam `DjangoStreamStore(using=…)` and `DjangoExecutor(using=…)` already use, and a
record written there commits independently of the caller's transaction. Two tests
state both numbers:

```
ambient connection, caller atomic() then rollback  ->  0 records survive
using="overlay",    caller atomic() then rollback  ->  1 record survives
```

Why not the others:

- **Option 1 as the *default*** — a store that always opened its own connection would
  also sit outside the test harness's rollback, leaving committed rows between tests,
  and it would be the only Django store here not writing on the ambient connection.
  Making independence the default is a decision about connection management, not about
  outcomes, and it wants its own task.
- **Option 2, refuse inside an atomic block** — the coordinator's own objection stands:
  `consume` swallows handler failures precisely so a store hiccup never breaks a save,
  and a consumer under `ATOMIC_REQUESTS` would be permanently unable to record anything.
  It converts "the record may be lost" into "the record is never written", which is
  worse.
- **Option 3, `transaction.on_commit`** — rejected explicitly. A rolled-back transaction
  is exactly when the record is most wanted; deferring to commit means the record exists
  only when nothing needed recording.
- **Option 4, document and leave it** — the store *can* see what core cannot, so leaving
  it undocumented-in-code would waste the one place the knowledge exists.

### Mutations run (each red on the test named for it)

| mutation | test that went red |
|---|---|
| drop `_check_widths` | 4× `test_a_name_longer_than_its_column_is_refused` — on SQLite **and** Postgres |
| `>` → `>=` in the width check | 4× `test_a_name_exactly_the_width_of_its_column_is_kept` |
| add `sequence_key` to `_BOUNDED` | `test_an_unbounded_field_is_not_length_checked` |
| `create` → `update_or_create` | `test_a_second_attempt_is_a_new_row_not_an_update` |
| store a summary, not `encode(outcome)` | `test_the_row_keeps_the_encoded_text_not_a_column_per_field` |
| trust the columns over the payload | `test_a_row_whose_columns_disagree_with_its_payload_is_not_reported` |
| `latest` takes the last row written | `test_latest_takes_the_highest_attempt_not_the_last_written` |
| `>=` → `>` on attempt | `test_re_recording_an_attempt_replaces_it` |
| drop the shared `_order` sort | `test_latest_orders_by_offset_with_append_stage_last`, `test_the_order_does_not_depend_on_when_things_were_recorded` |
| `record` ignores `self._using` | `test_an_outcome_on_another_alias_survives_the_callers_rollback` |
| `latest` ignores `self._using` | `test_an_outcome_on_another_alias_survives_the_callers_rollback` |

**One mutation stayed green, and it is not a gap.** Dropping the SQL
`filter(consumer=…, stream_path=…)` from `latest` leaves every test passing, because
the decoded payload's scope check catches it. The SQL filter is an index optimisation;
the payload check is the correctness guard, and that one is covered.

### Paperwork

`ConsumerOutcome` added to the Tier 2 list in `docs/public-api.md` and to the
enumeration in `tests/test_rakaia/test_public_api.py`; an `UPGRADING.md` entry under
"Unreleased" announcing migration `0010`, the widths and the transaction caveat.
Nothing is exported, matching ADR 0007 ("exporting is the last step before this is
Accepted"), so the API reference is unchanged.

joshbrooks:
## Re-architecture after review — evidence (commit `d147193`)

All five findings are addressed. Findings 1, 2 and 5 are closed by removing the
mechanism that produced them rather than by extending it; 3 and 4 are the small
fixes the review called for.

### The change in one sentence

The key columns no longer hold the values a consumer chose. They hold
`quote(value, safe="")` cut to the column width — which is `JsonlOutcomeStore._safe_name`'s
answer, borrowed. That store already had to turn an arbitrary consumer-supplied
string into something a filesystem accepts as one path segment; a `varchar` asks the
same question with different edges, and percent-encoded ASCII answers both: total,
injective, no NUL, no control bytes, nothing a text column can refuse, identical on
every backend.

### Why cutting is safe now and was not before

This is the load-bearing part. When a column held the value, truncating meant filing
a record under a name that is not its own — two subjects collapsing into one, the
exact defect `Outcome.subject` exists to fix, and the reason the previous version
refused rather than truncated. **When the column is an index, a prefix is still an
index.** Two names sharing a cut key both come back from the query and the payload
comparison in `latest` drops the one that does not belong. So the widths are a prefix
length rather than a capacity: cutting costs selectivity and never correctness.

That is what makes `record()` **total** — no length, no byte, no value a caller can
construct makes it raise — which is the finding-5 resolution the coordinator asked
for, and it needs no change to #248.

### Schema — `rakaia_consumeroutcome` (migration `0010`, regenerated in place)

| column | type | holds |
|---|---|---|
| `consumer_key` | `varchar(128)` | `quote(consumer)`, cut |
| `stream_path_key` | `varchar(255)` | `quote(stream_path)`, cut |
| `subject_key` | `varchar(255)` | `quote(subject)`, cut |
| `offset_key` | `varchar(64)`, nullable | `quote(offset)`, cut; NULL at append stage |
| `payload` | `TextField` | **the record** — `encode(outcome)`, byte for byte |
| `recorded_at` | `DateTimeField` | |

Renamed with a `_key` suffix on purpose: reading one and expecting the consumer's
value is the mistake the shape invites. `stream_path_key` for `submission/tf611` is
`submission%2Ftf611`.

**`attempt` is gone as a column** (finding 2), by deletion rather than by check. It is
a counter, not a name; nothing queries it — `latest` reads it from the payload — and
as `int4` it was the one field a database could still refuse for a reason unrelated to
text. The four columns that remain are exactly the four `Outcome.__post_init__` itself
names as the ones "a store sorts and names by".

### The divergences, measured before and after

Reverting `_key` to `return value[:limit]` (raw columns) and running the new tests:

```
NUL byte in `consumer`        SQLITE: passed
                              POSTGRES: psycopg.DataError: PostgreSQL text fields
                                        cannot contain NUL (0x00) bytes
```

Restoring `attempt` as a `PositiveIntegerField`, with the migration, on a fresh DB:

```
attempt = 2**31               POSTGRES: django.db.utils.DataError: integer out of range
```

Both now pass on both databases, because neither the byte nor the integer ever
reaches a column.

### Both test runs — final tree

| | result |
|---|---|
| `uv run pytest tests/ -q` (SQLite) | **2437 passed, 16 skipped, 1 xfailed** |
| `RAKAIA_TEST_DB=postgres`, port 55432 | **2451 passed, 2 skipped, 1 xfailed** |

`just lint`, `just fmt-check`, `just docs` clean. `just typecheck` 0 errors, 2
warnings (both pre-existing `__init__.py` `__all__` warnings). `just api-reference`
regenerates unchanged — still nothing exported.

### Findings 3 and 4

**3 — payload byte-identity.** `test_the_row_holds_the_bytes_encode_produced` now
asserts `row.payload == encode(outcome)`. Confirmed the old assertion was vacuous:
swapping `encode(outcome)` for a hand-rolled `json.dumps` of the same fields left the
suite green before and fails this test now.

**4 — I agree the comment was wrong, and "both must agree" is not quite it either.**
With derived keys, a column is not the value, so "agree" has no meaning; and the
asymmetry is real. The rule I documented and tested is **the keys narrow, the payload
decides**: a row the keys admit can still be rejected by its payload, and a row the
keys exclude is never fetched to be judged. Both directions are now pinned —
`test_a_row_whose_payload_names_another_scope_is_not_reported` and
`test_a_row_whose_keys_name_another_scope_is_never_fetched`. The second also makes the
SQL filter load-bearing, so the previously-green mutation is green no longer.

### Mutations — re-run in full against the new design

| mutation | test that went red |
|---|---|
| `_key` returns the raw value, cut | `test_a_key_column_holds_the_quoted_form_not_the_value` (SQLite); 5 errors incl. all four `nul_byte` cases (Postgres) |
| `_key` returns the raw value, uncut | 10 errors: `nul_byte`, `far_longer`, `refuses_nothing` (Postgres) |
| quoted but never cut | 5 errors: `far_longer`, `refuses_nothing` (Postgres) |
| `attempt` restored as `int4` (+migration) | `test_an_attempt_too_large_for_an_integer_column_round_trips` — `DataError: integer out of range` |
| hand-rolled `json.dumps` for the payload | `test_the_row_holds_the_bytes_encode_produced` |
| drop the payload comparison in `latest` | `test_two_names_sharing_a_cut_key_do_not_leak_into_each_other`, `test_a_row_whose_payload_names_another_scope_is_not_reported` |
| drop the SQL key filter | `test_a_row_whose_keys_name_another_scope_is_never_fetched` |
| `create` → `update_or_create` | `test_a_second_attempt_is_a_new_row_not_an_update` |
| `latest` takes the last row written | `test_latest_takes_the_highest_attempt_not_the_last_written` |
| `>=` → `>` on attempt | `test_re_recording_an_attempt_replaces_it` |
| drop the shared `_order` sort | `test_latest_orders_by_offset_with_append_stage_last`, `test_the_order_does_not_depend_on_when_things_were_recorded` |
| `record` ignores `self._using` | `test_an_outcome_on_another_alias_survives_the_callers_rollback` |
| `latest` ignores `self._using` | same |

**Every mutation is now red** — the previously-green one (dropping the SQL scope
filter) is covered by the new finding-4 test.

### Caller-transaction behaviour, unchanged from the last round

```
ambient connection, caller atomic() then rollback  ->  0 records survive
using="overlay",    caller atomic() then rollback  ->  1 record survives
```

joshbrooks:
## Follow-up: the two unread index columns are gone (commit `09e91f8`)

Confirmed the finding first — `subject_key` and `offset_key` were written by
`record` and read by nothing across `src/`; `subject_key` also appeared in
`__str__`, which is not a query. **Dropped both.**

### Why dropping, and not keeping with a comment

ADR 0007's own material argues for it more strongly than I expected:

- **Re-drive is deferred to its own decision.** It sits under "What would reopen
  this", not in a follow-up queue — "it needs a decision of its own about what a
  re-drive may write and which gates must pass first".
- **The admin is argued for, not scheduled.** The ADR calls it "a deviation, not a
  precedent being followed", and frames it as what happens "when the Django backend
  lands". It has landed, in this PR, with no admin — so the deviation has not been
  taken and nothing is waiting on these columns.
- **The ADR's own complaint about the motivating consumer** is that it has "neither a
  retention policy nor an index supporting the query its two hot paths run". That is
  an argument for indexes that match real queries, which is equally an argument
  against carrying ones that match imagined queries.

There is also a narrower point. Even if the worklist arrives tomorrow, a cut,
percent-encoded `subject_key` is probably not the column it would want — an operator
view filters on status and reason codes, which are payload-only, and would want an
exact subject rather than a prefix. So the speculative column is not even
speculatively right.

Against that: adding a column when the query arrives is a migration on a table with
no production rows, since ADR 0007 is Proposed and nothing is exported. Close to
free, as you said.

### Schema now

| column | type | holds |
|---|---|---|
| `consumer_key` | `varchar(128)` | `quote(consumer)`, cut |
| `stream_path_key` | `varchar(255)` | `quote(stream_path)`, cut |
| `payload` | `TextField` | the record — `encode(outcome)` byte for byte |
| `recorded_at` | `DateTimeField` | |

Two columns because `latest(consumer, stream_path)` is the only query. Migration
`0010` regenerated in place. `__str__` is now the scope plus the row id, since a
scope holds many outcomes and naming a subject would suggest otherwise.

The docstrings say which option was taken and why, including your `LIMIT`
observation — `latest` applies none, so a cut key can only over-fetch and never
under-fetch — which is a cleaner statement of the safety argument than what I had.

### New test

`test_the_table_has_exactly_the_columns_the_only_query_needs` pins the concrete field
list. A column added here now has to be a deliberate act that arrives with the query
that reads it, which is the thing review had to catch by hand this round.

### Runs and mutations

| | result |
|---|---|
| SQLite | **2438 passed, 16 skipped, 1 xfailed** |
| Postgres (55432) | **2452 passed, 2 skipped, 1 xfailed** |

`just lint`, `just fmt-check`, `just docs` clean; `just typecheck` 0 errors and the 2
pre-existing warnings; `just api-reference` unchanged.

Re-ran only the mutations touching changed code, as directed — the other seven stand
from the previous round. All six red:

| mutation | test that went red |
|---|---|
| an unread `subject_key` column returns | `test_the_table_has_exactly_the_columns_the_only_query_needs` |
| `latest` stops narrowing by consumer | `test_a_row_whose_keys_name_another_scope_is_never_fetched` |
| `_key` returns the raw value | `test_a_key_column_holds_the_quoted_form_not_the_value` |
| drop the payload comparison | `test_two_names_sharing_a_cut_key_do_not_leak_into_each_other`, `test_a_row_whose_payload_names_another_scope_is_not_reported` |
| drop the SQL key filter | `test_a_row_whose_keys_name_another_scope_is_never_fetched` |
| hand-rolled `json.dumps` for the payload | `test_the_row_holds_the_bytes_encode_produced` |

Issue #118, opened 2026-08-15
The lockfile keeps rewriting itself in unrelated changes


joshbrooks:
A recent change that touched only the version number also carried a few hundred
lines of unrelated churn in the dependency lockfile. Nothing was wrong with it —
it was just a different version of our package tool rewriting the same
information in its newer style, because we never say which version anyone should
be using.

The cost is small but recurring: it makes diffs harder to review, and two people
on different tool versions can undo each other's lockfile indefinitely. Naming
one version for both continuous integration and local work stops it.

joshbrooks:
Observed in #111: a change to `src/rakaia/__init__.py` and one test file also rewrote `uv.lock`, normalising dependency markers (`{ name = "cryptography" }` → `{ name = "cryptography", marker = "python_full_version < '3.11'" }`) across many entries. Semantically identical resolution, different serialisation.

Cause: `.github/workflows/ci.yml` pins the *action* (`astral-sh/setup-uv@v9.0.0`, deliberately, with a comment) but not the uv version it installs, so CI takes whatever is latest. Local uv here is 0.9.22.

Fix: add `version:` to the `setup-uv` step in both jobs in `ci.yml` (and `conformance.yml` if it installs uv), and add a `required-version` to `pyproject.toml`'s `[tool.uv]` so a local uv outside the range refuses rather than silently rewriting. Then regenerate the lock once, deliberately, as its own commit.

PR #250, opened 2026-09-06
Write down what the last two releases changed

joshbrooks:
The guided tour was last updated the day before the first of the last two
releases was tagged, so nothing from either one ever reached it, and the seven
changes merged since had no record anywhere. This writes them all down across the
tour, the changelog and the upgrade notes — including the one trap that fails
silently: pointing the library at a different place to keep its log does not move
the log you already have, and every saved reading position stays valid, so
consumers resume into an empty stream instead of failing. It also records the
rule that would have caught the gap, namely which files a release has to touch.

Two things found while checking that the examples still work. The translations
example demanded a message broker its production command no longer uses, which
locked out anyone without the container tool installed; that is gone. And the
package tool version we require could no longer be installed at all, because the
usual package manager carries only the newest release and our range excluded it,
so the range now covers the current version.

https://claude.ai/code/session_01DmUw5DmbdDmzzE67KVQrEm

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
           # Pin uv itself, not just the action. Unpinned, CI takes whatever is
           # latest and rewrites uv.lock in its newer style — a version-bump PR
           # once carried a few hundred lines of unrelated marker churn (#118).
-          version: "0.11.32"
+          version: "0.12.10"
 
       - name: Set up Python 3.12
         run: uv python install 3.12
@@ -120,7 +120,7 @@ jobs:
       - uses: astral-sh/setup-uv@v9.0.0
         with:
           enable-cache: true
-          version: "0.11.32"
+          version: "0.12.10"
 
       - name: Set up Python 3.12
         run: uv python install 3.12
@@ -144,7 +144,7 @@ jobs:
       - uses: astral-sh/setup-uv@v9.0.0
         with:
           enable-cache: true
-          version: "0.11.32"
+          version: "0.12.10"
 
       - name: Set up Python 3.12
         run: uv python install 3.12
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -38,7 +38,7 @@ jobs:
           # Pin uv itself, not just the action. Unpinned, CI takes whatever is
           # latest and rewrites uv.lock in its newer style — a version-bump PR
           # once carried a few hundred lines of unrelated marker churn (#118).
-          version: "0.11.32"
+          version: "0.12.10"
 
       - name: Set up Python 3.12
         run: uv python install 3.12
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -40,7 +40,7 @@ jobs:
           # Pin uv itself, not just the action. Unpinned, CI takes whatever is
           # latest and rewrites uv.lock in its newer style — a version-bump PR
           # once carried a few hundred lines of unrelated marker churn (#118).
-          version: "0.11.32"
+          version: "0.12.10"
 
       - name: Set up Python 3.12
         run: uv python install 3.12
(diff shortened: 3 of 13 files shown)


joshbrooks:
## Evidence

### The gap

`docs/whats-new.md` last touched **2026-08-22**; `v0.3.0` tagged **2026-08-23**, `v0.3.1` **2026-08-24**. Neither release reached the tour. `[Unreleased]` was consumed by the 0.3.1 prep (#236) and never re-added, so #237–#243 had no changelog home.

| File | Before | After |
|---|---|---|
| `CHANGELOG.md` | 0.3.0, 0.3.1; no `[Unreleased]` | `[Unreleased]` restored, all 7 PRs backfilled |
| `UPGRADING.md` | 0.3.0 only | + 0.3.1, + Unreleased |
| `docs/whats-new.md` | 13 sections, none from 0.3.x | + §14–18 |
| `docs/examples.md` | Known gaps missing 0.3.x APIs | + 6 entries |

New public names since `v0.3.1`, all now documented: `JsonlStreamStore`, `migrate_stream`, `migrate_all`, `Migration`, `get_asgi_app`, `reset_store_cache`.

### Doc samples verified against real signatures

`rebuild_and_verify` was drafted with three wrong arguments (`store=`, `path=`, model *labels*). Real signature is `(stream_path, *, into, live_models: Sequence[type[Model]])`. Corrected, and every name in every new snippet was import-checked:

```
from rakaia import migrate_stream, migrate_all, Migration, JsonlStreamStore
from django_rakaia import rebuild_and_verify, GuardNotArmed, ScratchAliasNotEmpty, get_asgi_app, reset_store_cache
```

Also corrected `report.ok` → `report.verdict` (`ok` is vacuously true on an empty population; `verdict` separates `GREEN` from `VACUOUS`).

### Examples: all 13 exercised against this branch

`just demos` covers 11. The two it skips need a running server, so they were driven by hand:

- **chat** — posted through the form, asserted the message arrived on `/streams/api/streams/room:1:messages/sse/`.
- **polyglot** — saved from a **separate process**, asserted delivery on `/protocol/translations/tet?live=sse&offset=now`. This is #242's headline claim and nothing in CI covers it.

### The broker was genuinely dead

Only reference to a channel layer anywhere in `examples/polyglot/` is a docstring explaining why it is *not* used. Proof it was vestigial — `settings_prod` + hypercorn, 2 workers, `REDIS_URL` pointed at a **dead** port 6399:

```
GET /                              -> 200
GET /protocol/translations/tet     -> 200
cross-worker SSE delivery          -> 1 (delivered, no broker)
redis/connection errors in log     -> 0
```

`redis-up` and `REDIS_URL` stay in the justfile — the chat `serve` recipe still uses them legitimately.

### uv pin

Tested in a throwaway copy before changing anything:

| Test with 0.12.10 | Result |
|---|---|
| `uv lock` on committed lockfile | unchanged |
| `uv sync --extra dev --extra django` | unchanged |
| `uv lock` from scratch vs 0.11.32 from scratch | **byte-identical**, both `revision = 3` |

So #118's serialisation-churn risk does not apply to this jump. Six sites moved: `pyproject.toml` + `ci.yml`×3 + `publish.yml` + `conformance.yml`. `uv.lock` is untouched in this PR.

No changelog entry for the pin, matching precedent — `4629aa3 Move the uv pin to 0.11.32` was not changelogged, and 0.3.0 changelogs no tooling. The rule added to `CLAUDE.md` states that exemption explicitly.

### Gate (Python 3.12, matching CI)

`ruff check` · `ruff format --check` · `api-reference-check` · `pyright src/` **0 errors** · `zensical build` · **2400 passed, 16 skipped, 1 xfailed** · `just demos` all passed.

### Not fixed here

`just api-reference-check` fails on Python 3.14 and passes on 3.12 — `inspect.signature` renders `Optional[Literal[...]]` vs `Literal[...] | None`. A contributor on a newer Python sees a spurious failure and could commit a regeneration that breaks CI. Left alone; it needs a decision (pin the generator's Python, or normalise the annotations).

Issue #257, opened 2026-09-07
The reason a failure was recorded is just an internal error name


joshbrooks:
A failure record keeps its reasons as codes rather than a sentence, so that "how many failed for this reason" is a question you can answer by counting. When the library itself is what failed, though, the code it writes down is simply the name of whatever internal error class was raised. Renaming a class quietly rewrites the vocabulary an operator has been reading, and anything unexpected lands under a single generic name that says nothing about what went wrong.

There is a second half to it: the six things replay can raise share no common type, so a consumer that wants to catch "anything the library could not apply" has to list all six or catch everything. Giving them one shared type with a stable code attached fixes both — the record gets a vocabulary that survives a refactor, and callers get one thing to catch. Worth deciding first whether those codes become a promise we keep, since the decision record currently says reason codes belong to the consumer and is silent on what happens when the library is the one that failed.

joshbrooks:
### Where it is

`src/rakaia/subscription.py:276-296` — the consume loop's handler:

```python
except Exception as exc:
    _record(Outcome(..., reasons=(type(exc).__name__,)))
```

So the reason code for a genuine library failure is a class name, and every unanticipated bug arrives as `ValueError` — indistinguishable from a decode failure, a missing reader, or a bad merge key.

### The six, with no shared base

- `HandlerGapError`, `UpcasterChainError` — `src/rakaia/registry.py:47,55`
- `EffectCollisionError`, `UnresolvedRefError`, `DuplicateProducesError` — `src/rakaia/effects.py:286-298`
- `HandlerDriftError` — `src/rakaia/drift.py:45`
- plus `ValueError` from `replay.py:269` (staged replay with no reader), `:816-821` (decode), `:753-757` and `:780-802` (merge keys)

`replay()` catches none of them. A caller wanting "anything replay can raise" enumerates all six or catches bare `Exception`, which is what the loop does.

### Decision needed before this is agent-ready

ADR 0007 Decision 6 says reason codes are the consumer's and opaque to rakaia. It never addresses the case where rakaia itself failed. Either the library's own codes become a promised, documented set (and the ADR gets an amendment), or the record says only "unhandled" and puts the type in `params`. Second one is cheaper and honest; first one is what makes #252's screen readable. My preference is the first, with a small closed set.

### Conflicts

Edits the `except` clause in `subscription.py` that #255 is deliberately staying out of, plus `registry.py`, `effects.py`, `drift.py`, and the export files. Wave 2, after #255 merges.

[...]

Issue #254, opened 2026-09-07
Nothing demonstrates the new consuming loop


joshbrooks:
The library now offers a single loop that reads events, processes them, commits the reading position, and records anything it could not apply. The example project still does part of that by hand — it polls for events one call at a time and threads the position through itself, which is what the new loop replaces.

Until an example uses it, the documentation describes something no working code demonstrates, and the first person to try it is also the first to find out whether the documentation is right. The gap is already written down honestly in the examples page rather than hidden; this closes it.

Issue #180, opened 2026-08-20
The event log cannot be pointed at a second database, but everything around it can


joshbrooks:
Rebuilding projections into a throwaway database is the way we prove a rebuild is trustworthy without touching production. The two pieces that write and read those projections can both be aimed at another database. The log itself cannot, so a rebuild cannot read its own events from the throwaway one.

The workaround is written down as instructions rather than code: copy the log into memory first, in about six lines, at every place that wants to do this. That is a missing option presented as a ritual, and because the log is untestable off the default database, the six lines every consumer writes are tested nowhere.

PR #258, opened 2026-09-07
Hold a consumer as one thing, so it cannot forget to record

### LINKED ISSUE #255: Setting up a consumer takes ten decisions, and two of them fail quietly
Reading a stream and applying it is now one supported loop, but wiring one up means supplying ten separate things at the call site: where the log is, which stream, what to do with each event, the consumer's name, the failure policy, where to load the reading position from, how to save it back, where to keep the record of anything that could not be applied, and two optional ways of naming what a record is about. The consumer's name has to be typed three times and nothing checks the three agree.

Two of those ten do nothing if you leave them out. Omit the one that saves the position and the loop still runs, re-reading the same events forever; omit the one that keeps the failure records and it runs recording nothing, which is the exact condition this whole feature exists to prevent. Both look like a working consumer. This asks for one small module that holds the scope, the position and the failure records together, so a consumer is built once and cannot be built half-configured.

joshbrooks:
Consuming a stream took six arguments and two of them could be left out without anything complaining: omit the place outcomes are kept and the loop still polls, still applies and still moves the watermark, while recording nothing. Under the rule this library just adopted, a stream with no records is a stream where everything worked — so the wiring that records nothing looks exactly like a clean run. The consumer's name and the stream path also had to be repeated in each of the three places they were needed, and getting one of the three wrong files the outcomes under one identity and the watermark under another.

A consumer is now something you build once and run. The name and the path are said once and reach both halves, and somewhere to keep outcomes is a required argument, so the silent shape cannot be built at all. The Django entry point fills in the durable places for both and refuses to start inside a transaction the caller opened — the one hole the loop could never close on its own, because a rollback out there takes the record of the failure with it and the dependency-free core cannot see a database transaction. How to stop and start it stays the caller's decision, and the choice between skipping a bad event and stopping on it still has no default.

Closes #255


joshbrooks:
## Evidence

### Red → green, per test

| Test | Red how | Green after |
|---|---|---|
| `tests/test_rakaia/test_consumer_module.py::TestRecordingIsOnByConstruction::test_a_consumer_without_an_outcome_store_cannot_be_built` | `ModuleNotFoundError: No module named 'rakaia.consumer'` (collection error) | `src/rakaia/consumer.py` — `outcomes` is a `@dataclass` field with no default, so omitting it is a `TypeError` |
| `tests/test_rakaia/test_consumer_module.py::TestTheNameAndPathAreSaidOnce::*` | same collection error | `Consumer.run` threads `self.name` / `self.path` into `consume(consumer=…, path=…)`, the cursor load and the commit closure |
| `tests/test_django_rakaia/test_consumer_entry.py::TestTheCallersTransactionIsRefused::*` | `ModuleNotFoundError: No module named 'django_rakaia.consumer'` | `DjangoConsumer.run` checks `connections[self.alias].in_atomic_block` |

### The guard is load-bearing (mutation)

Replaced `if connections[self.alias].in_atomic_block:` with `if False:` —

```
FAILED test_consumer_entry.py::TestTheCallersTransactionIsRefused::test_running_inside_an_open_transaction_raises
E   Failed: DID NOT RAISE <class 'django_rakaia.consumer.CallerTransactionOpen'>
1 failed, 1 passed
```

### `transaction=True` is load-bearing

Downgraded the marker to plain `@pytest.mark.django_db` —

```
FAILED test_consumer_entry.py::TestTheCallersTransactionIsRefused::test_running_outside_one_records_and_commits
E   django_rakaia.consumer.CallerTransactionOpen
1 failed, 1 passed
```

pytest-django has already opened the transaction the guard detects, so the ordinary path is refused too and the green half of the pair cannot be written. The #180 failure mode arriving from the other side.

### Gates

| Gate | Result |
|---|---|
| `just fmt-check` | 492 files already formatted |
| `just typecheck` (pyright 1.1.411) | 0 errors, 2 warnings (both pre-existing `reportUnsupportedDunderAll` on the two lazy `__init__.py`) |
| `uv run pytest` | 2476 passed, 16 skipped, 2 xfailed |
| `just api-reference-check` | clean after regenerating |
| `uv run zensical build` | No issues found |

### Files

- `src/rakaia/consumer.py` — `Consumer`, `CursorLedger` (protocol), `InMemoryCursorLedger`. Imports `subscription`, so it is classified `PROTOCOL_SERVER` in `test_tier_boundary.py`; no new crossing.
- `src/django_rakaia/consumer.py` — `django_consumer`, `DjangoConsumer`, `DjangoCursorLedger`, `CallerTransactionOpen`.
- `src/django_rakaia/subscription.py` — `load_cursor` / `commit_cursor` gained keyword-only `using=None` (additive) so the ledger can point at an alias without duplicating the two queries. `consume()`'s body is untouched — #257 owns that.
- `scripts/gen_api_reference.py` — the seven new names grouped into "Consuming, and what happened to an event" rather than left in "Everything else".
- Public surface pins updated in `tests/test_rakaia/test_public_api.py`.

### Notes for review

- The guard is on `run`, not on construction: a consumer built at startup and run from a request handler is fine, and one built outside a transaction and run inside one is the case being refused.
- It inspects the connection for the alias the cursor and outcome rows are written on, since it is that connection's transaction their survival depends on — `using=` therefore both selects where the rows go and what is checked.
- `CHANGELOG.md` will conflict with the three sibling branches; resolve by keeping every entry.

joshbrooks:
### Review round 1 — two blockers, both fixed in `8a3bffb`

**1. The guard inspected a connection the rows were not on.** `alias` was its own field defaulting to `default`, and only the factory tied it to `using=`. Since `DjangoConsumer` is exported, building one directly with both stores on another alias left the guard checking `default` — it passed precisely when it should have refused, and a caller reads a pass as permission.

Fixed by deriving the alias from the stores themselves: each is asked where it writes (`using`, now public on both), and every named connection is checked. A cursor and an outcome kept on different aliases are both covered, and a store that names no alias is not guarded at all — it is not database-backed, so no transaction of the caller's can roll back what it keeps.

**2. The `using=` seam had no coverage.** The reviewer stripped `.using(using)` from both cursor helpers and the full suite stayed green. Now pinned.

**3. `CursorLedger` renamed to `ConsumerCursorStore`** (`InMemoryConsumerCursorStore`, `DjangoConsumerCursorStore`). It sat one line from `CursorStore` — the *event* store a subscriber polls — meaning nearly the opposite thing, and it broke the `…Store` convention its sibling `OutcomeStore` uses. Renaming after #254 writes an example against it costs a deprecation; renaming now costs a sed.

### Mutations, each proved landed before the result was believed

| mutation | occurrences before → after | result |
|---|---|---|
| strip `.using(using)` from both cursor helpers | 2 → 0 | `test_an_alias_sends_both_rows_to_that_database` red, 6 green |
| `_aliases()` returns `[DEFAULT_DB_ALIAS]` (the original defect) | 1 → 0 | 4 red — the three alias tests and the not-guarded one |
| `_alias_of` treats "names no alias" as the default alias | branch removed | `test_a_store_that_names_no_database_is_not_guarded` red, 6 green |

Tree confirmed clean after each revert.

### Gates

`ruff check` clean · `ruff format --check` 493 files · `pyright src/` 0 errors (2 pre-existing `__all__` warnings) · `uv run pytest` 2573 passed / 16 skipped / 2 xfailed · `just api-reference-check` no drift. Merged with `main` at `c841db2`, so the new docs-names gate ran against the renamed prose.

### Not fixed, deliberately

`docs/whats-new.md` section 20 collides with #261, which appends its own. Textual only — grepped for cross-references to section numbers and found none, so whichever lands second renumbers to 21 and no content becomes wrong.

joshbrooks:
### Review round 2 — nothing blocking; five informational findings, four taken

Round 2 confirmed all three of round 1's findings are genuinely fixed, and — importantly — **attacked the reasoning behind the fix rather than just the diff**. The inverted defect I was most worried about is not present: it enumerated every `OutcomeStore` and cursor-store implementation in the tree and confirmed the only database-backed ones both expose `using`, so no connection that can swallow a record goes unguarded. It also proved the two stores are *separately* load-bearing — making `DjangoOutcomeStore.using` return `None` reddens a test on its own — so the guard is not passing on the strength of one of them.

It agreed that not guarding the event store is right: `consume` only reads from it, so a rollback there loses nothing.

Fixed in this commit:

- **Neither seam declared the attribute the guard reads.** `_alias_of` duck-types on the spelling `using`, so a future database-backed store naming it `db`, or keeping it private, would be silently unguarded with no signal anywhere. Both protocol docstrings now say it. Nothing reachable today — this is a note for whoever writes the second implementation.
- **`ATOMIC_REQUESTS = True` is how a real deployment meets this error first**, without having written a transaction at all. Now in the error's own docstring and in `docs/subscriber-cursors.md`, with the answer: run the consumer outside the request, or on an alias that setting does not cover.
- **The rename left its own word behind** in a test name, a local and two comments. Retired.

Not actioned: `str(using)` on a store exposing `using` as a method yields a bad alias and raises `ConnectionDoesNotExist` — loud, not silent, and the right direction.

On the shape question for #254, round 2's answer: `django_consumer(store, path, name, using=…)` is the whole entry point, `run` is the only method, and the guard needs no argument from the caller, so there is nothing for an example to keep in step.

### Gates

`uv run pytest` 2573 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `api-reference-check` no drift. Collected node ids diffed against `c841db2`: +11 new tests, nothing removed or renamed away.

PR #259, opened 2026-09-07
Decide the order a batch is applied in once

### LINKED ISSUE #256: The order writes reach the database is written out three times
When a batch of changes is applied, they always go in the same order: writes first, then deletions, then retirements. That single rule is currently stated separately in three places — once where a batch is buffered up during replay, and once inside each of the two things that can apply a batch. Nothing forces the three to agree with each other, and only two of them are covered by the shared conformance suite.

The consequence is small today and grows: adding a fourth kind of change, or altering the order, means three edits that have to land together, and getting one wrong writes data in the wrong order rather than failing. The sibling rule about which fields a batch may write already lives in one shared place for exactly this reason; this one belongs beside it.

joshbrooks:
A batch of changes reaches storage as writes, then deletes, then retires. That ordering was written out three separate times — once where replay decides how many events can share a batch, and once inside each of the two things that write one — so two of the three could agree while the third quietly drifted, and nothing in the suite would have caught it. The rule now lives beside the other batch-level rule it belongs with, and all three read it from there.

Nothing about the order itself changes, so nobody using the library sees a difference and there is no changelog entry: by long precedent that file documents the library, not the internal structure. A new conformance test pins each of the three readers to the shared rule and was watched fail with each one put back on its own copy in turn; the evidence is in a comment below.

Closes #256

### DIFF
--- a/src/django_rakaia/effect_executor.py
+++ b/src/django_rakaia/effect_executor.py
@@ -35,6 +35,7 @@ from django.db import DEFAULT_DB_ALIAS, connections, transaction
 from django.db.models import Q
 
 from rakaia.effects import (
+    _WRITE_ORDER_PASSES,
     ApplyReport,
     Delete,
     Effect,
@@ -44,6 +45,7 @@ from rakaia.effects import (
     SpareKeys,
     Update,
     Upsert,
+    _write_order_rank,
     check_disjoint_defaults,
 )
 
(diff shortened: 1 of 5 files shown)


joshbrooks:
### What moved

`_write_order_rank` now lives in `src/rakaia/effects.py`, beside `check_disjoint_defaults` / `_WrittenFields`, together with `_WRITE_ORDER_PASSES = (0, 1, 2)`. Both stay private — no `__all__` entry, so `docs/api-reference.md` and `tests/test_rakaia/test_public_api.py` are untouched (`just api-reference-check` regenerates to no diff).

The three readers:

| Site | Before | After |
| --- | --- | --- |
| `src/rakaia/replay.py` (`_StageBuffer`) | defined its own `_write_order_rank` | imports it |
| `src/rakaia/executors.py` `InMemoryProjections.apply` | three hard-coded `for` passes | `for rank in _WRITE_ORDER_PASSES:` with `_write_order_rank(eff) != rank: continue` |
| `src/django_rakaia/effect_executor.py` `DjangoExecutor.apply` | three hard-coded passes, `Update` run-collapsing in the first | same rank loop; the `pending` run is opened per pass and flushed at its end, so #199's collapsing is unchanged |

### Mutation evidence — each site restored to its own copy, in turn

New test file: `tests/test_django_rakaia/test_write_order_conformance.py` (6 tests). Three identity tests (`module._write_order_rank is effects._write_order_rank`) and three behaviour tests that monkeypatch each module's name with a reversed rule (`2 - rank`) and assert the observed order follows it.

| Mutation | Conformance file | `test_replay_batching.py` + `test_executor_contract.py` (64 tests) |
| --- | --- | --- |
| A — `replay.py` re-declares its own `_write_order_rank`, import dropped | FAIL `test_every_site_reads_the_one_rule[buffer]` | 64 passed |
| B — `executors.py` back to three hard-coded passes | FAIL `test_the_in_memory_executor_takes_its_passes_from_the_rule` | 64 passed |
| C — `effect_executor.py` back to three hard-coded passes | FAIL `test_the_django_executor_takes_its_passes_from_the_rule` | 64 passed |

Each mutation was reverted and the file re-run green. Both halves of the test are load-bearing: A left no reference to the shared name and only the identity assertion caught it; B and C left the import in place (ruff had hoisted it to the top of the import block, so my edit did not remove it) and only the behaviour assertion caught it. In all three, the existing suites stayed green — which is the gap this issue named.

One test-file gotcha worth recording: the modules are grabbed with `import_module("rakaia.replay")`, because `from rakaia import replay` hands back the re-exported `replay` *function*, and every assertion would then have been about the wrong object.

### Gates

- `just lint` — All checks passed
- `just fmt` / `ruff format --check` — 489 files already formatted
- `just typecheck` — 0 errors, 2 warnings (the two pre-existing `reportUnsupportedDunderAll` on `__init__.py`)
- `uv run pytest` — 2476 passed, 16 skipped, 2 xfailed
- `just api-reference-check` — no drift (155 names, 0 ungrouped)

Not run: the Postgres leg and `zensical build` (no docs extra synced in this worktree). Neither is touched — no locking, no docs change.

joshbrooks:
Review round 1 found nothing blocking, so this converges in one round.

The reviewer verified no behaviour change three ways: a spy on the update-collapsing run sizes across six batch shapes, old code against new, identical in every case; a differential fuzz of 4,400 random batches comparing rows, report and exception type between the old and new modules, zero mismatches, with the harness shown sensitive by flipping a rank; and a check that within each pass the emission order is untouched. It also re-ran the three-site mutation evidence and confirmed the identity and behaviour halves of the conformance test each catch regressions the other misses.

Issue #253, opened 2026-09-07
Old failure records accumulate with no way to remove them


joshbrooks:
Failure records are only written when something goes wrong, so the table stays small in normal use. But nothing ever removes them, and there is no way to say "delete anything older than a year" without writing SQL by hand.

This needs a command that takes an explicit age, deliberately with no default — the right retention period differs by installation, and a wrong guess deletes evidence of a problem someone is still investigating. It should refuse to run and change nothing when the age is left out. The lookup it performs is also the first real justification for indexing the record's timestamp, so the two belong together.

PR #260, opened 2026-09-07
An administrator can look at the failure records

### LINKED ISSUE #252: An administrator cannot look at the failure records
The library now keeps a record of every event a consumer could not process — what it was, and why it was refused or lost. There is no screen for looking at them, so the only way to read one is to query the database directly.

This needs a read-only page listing them, newest first, with the reason shown. Read-only on purpose: these are a record of what happened, not something to edit, and removing them is a separate retention job. There is an existing pattern in the same file to copy for formatting, and the decision record already argues why this table is the exception that gets a screen when the others do not.

joshbrooks:
The library keeps a record of every event a consumer could not process, and until now the only way to read one was to query the database directly. There is now a page in the Django admin listing them newest first, showing the consumer, the stream, the subject, the position, the stage, the status and the reasons behind it.

The values shown come out of the stored record, not the two index columns beside it — those hold a percent-encoded form of the name and would print something no consumer ever passed. The page is read-only throughout: these are a record of what happened, and clearing them out is the retention job in #253 rather than a button here. A record written by a version this one cannot read still gets a row, marked as unreadable, because a record dropped from the page is the failure the whole table exists to prevent.

Closes #252

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ runnable demo for each.
 
 ### Added
 
+- **A screen for the outcome records.** The database-backed store keeps a row for
+  every event a consumer could not apply, and until now reading one meant querying
+  the table by hand. The Django admin lists them newest first, with the consumer,
+  the stream, the subject, the position, the stage, the status and the reasons —
+  decoded from the stored record rather than read off the two index columns beside
+  it, which hold a percent-encoded form of the value and would show
+  `submission%2Ftf611` for a stream named `submission/tf611`.
+
+  Read-only on purpose: nothing can be added, changed or deleted from it. These
+  are a record of what happened, and clearing them out is a retention job rather
+  than a button. A record written by a version this one cannot read still gets a
+  row, marked as unreadable, because a dropped record is the failure the table
+  exists to prevent. (#252)
+
 - **`JsonlStreamStore` — keep a log in plain text files instead of a database.**
   A stream is a directory of JSON-lines segments; an event is a line. Nothing but
   the filesystem is involved, so a consumer who wants durability without running
--- a/src/django_rakaia/admin.py
+++ b/src/django_rakaia/admin.py
@@ -336,14 +336,20 @@ class ConsumerOutcomeAdmin(admin.ModelAdmin):
         nothing and said so as an empty page — an operator reads that as "there
         is no record", which is the one conclusion this table exists to prevent.
 
-        Escaping the term the same way the payload was escaped makes the two
-        comparable. For a term that is already ASCII this is the identity, and a
-        term carrying a quote or a backslash is escaped to exactly the form the
-        payload stores.
+        **Only the characters that were escaped are escaped back, one at a
+        time.** Escaping the whole term through `json.dumps` was the first
+        attempt and it broke quoted phrases: Django splits the search box on
+        spaces and *then* strips the quotes around a bit, so a term arriving as
+        ``\"two words\"`` never matches that branch and is searched literally,
+        which returned nothing at all for a query that used to work. Anything
+        already ASCII is left exactly as the operator typed it, so every search
+        that worked before this method existed still works, character for
+        character.
         """
-        return super().get_search_results(
-            request, queryset, json.dumps(search_term)[1:-1]
+        escaped = "".join(
+            char if char.isascii() else json.dumps(char)[1:-1] for char in search_term
         )
+        return super().get_search_results(request, queryset, escaped)
 
     def has_add_permission(self, request) -> bool:  # noqa: ARG002
         return False
--- a/src/django_rakaia/admin.py
+++ b/src/django_rakaia/admin.py
@@ -302,7 +302,11 @@ class ConsumerOutcomeAdmin(admin.ModelAdmin):
     # `recorded_at` can tie — a batch of failures lands in one transaction — so
     # the primary key breaks it, and the newest row is first either way.
     ordering = ["-recorded_at", "-pk"]
-    date_hierarchy = "recorded_at"
+    # No `date_hierarchy`. It costs two `COUNT(*)`s and a `SELECT DISTINCT` over
+    # a truncated `recorded_at` on every page load, none of which an index on
+    # that column can serve, and this table has no bound on its size until a
+    # retention job is scheduled. Newest-first ordering answers the question the
+    # hierarchy was there for.
     # The payload is the record, so searching it searches every field of every
     # outcome at once, in the values the consumer passed rather than the keys.
     search_fields = ["payload"]
@@ -322,6 +326,24 @@ class ConsumerOutcomeAdmin(admin.ModelAdmin):
         "payload",
     ]
     readonly_fields = fields
+
+    def get_search_results(self, request, queryset, search_term):
+        """Search the payload for what the consumer passed, not for its spelling.
+
+        `encode_outcome` renders with `json.dumps`, which escapes anything
+        outside ASCII, so a stream a consumer named `café/tf611` sits in the
+        column as ``caf\u00e9/tf611``. Searching the term as typed matched
+        nothing and said so as an empty page — an operator reads that as "there
+        is no record", which is the one conclusion this table exists to prevent.
+
+        Escaping the term the same way the payload was escaped makes the two
+        comparable. For a term that is already ASCII this is the identity, and a
+        term carrying a quote or a backslash is escaped to exactly the form the
+        payload stores.
+        """
+        return super().get_search_results(
+            request, queryset, json.dumps(search_term)[1:-1]
+        )
 
     def has_add_permission(self, request) -> bool:  # noqa: ARG002
         return False
--- a/src/django_rakaia/admin.py
+++ b/src/django_rakaia/admin.py
@@ -6,7 +6,7 @@ in the Django admin.
 """
 
 import json
-from typing import Any
+from typing import Any, cast
 
 from django.contrib import admin
 from django.utils.html import format_html, format_html_join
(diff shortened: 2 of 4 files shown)


joshbrooks:
## What landed

`ConsumerOutcomeAdmin` in `src/django_rakaia/admin.py`, registered for `ConsumerOutcome` (`rakaia_consumeroutcome`).

- `ordering = ["-recorded_at", "-pk"]` — the timestamp can tie when a batch of failures lands in one transaction, so the primary key breaks it.
- `list_display`: `recorded_at`, `consumer`, `stream_path`, `subject`, `offset`, `stage`, `status_badge`, `reasons`. **None of these is a model field** — each is a display method reading the decoded record. The detail view adds `sequence_key`, `params`, `attempt` and the raw `payload`.
- `has_add_permission` / `has_change_permission` / `has_delete_permission` all return `False`. `has_view_permission` is left alone, so the row still links to a read-only detail page. Deletion is #253's retention job.
- `_decoded(obj)` calls `decode_outcome` once per row and caches the result on the instance (`_rakaia_decoded`, via `getattr`/`setattr` so pyright is happy — `obj.__dict__` is typed `MappingProxyType` and `reportIndexIssue` fires on assignment). Eight columns per row would otherwise decode eight times and, for an undecodable payload, log eight warnings for one row.
- `status_badge` carries the undecodable case (grey `UNREADABLE` badge, `_STATUS_COLORS` beside `_BADGE_COLORS`); every other decoded column falls to `_ABSENT` (`—`). Putting the marker on the one column every row has avoids a row of em-dashes that reads as a rendering bug rather than a record from another version.
- `search_fields = ["payload"]` searches the record text, so a search for `submission/tf611` matches — the key columns hold `submission%2Ftf611` and would not.
- Existing idiom reused: `_truncate` for every cell, the `format_html` badge span copied from `_event_badge`, `format_html_join`/`mark_safe` for `params` as in `streams_list`, `date_hierarchy` as on `StreamEntryAdmin`.

Why this table gets a screen at all: ADR 0007 Consequences, "The admin registration will break a local pattern" — `ConsumerCursor`, `StreamOffsetWatermark` and `StreamProducer` have no admin because bookkeeping is not browsed; a dead-letter table is the argued exception because looking at it *is* the feature. Implemented as decided, not re-litigated. The class docstring says so at the point of deviation.

## Red → green

Tests are `TestTheOutcomeScreenReadsThePayload` in `tests/test_django_rakaia/test_admin.py`, asserting on what `result_list(changelist)` renders (the same call the changelist template makes) rather than on what a display method returns — the file's existing convention, and the distinction that let #195 and #208 each fix one surface and leave another wrong.

First red was the `ImportError` for `ConsumerOutcomeAdmin`. To get a red that discriminates rather than a red that just fails to import, the tests were then run against a deliberately naive stub — `list_display = ["consumer_key", "stream_path_key", "payload", "recorded_at"]` and nothing else:

| test | against the naive stub |
|---|---|
| `test_the_screen_shows_the_values_not_the_encoded_keys` | FAIL — `submission%2Ftf611` rendered |
| `test_nothing_can_be_added_changed_or_deleted` | FAIL — `has_add_permission` was `True` |
| `test_a_payload_this_version_cannot_read_still_renders` | FAIL — no marker on the undecodable row |
| `test_the_newest_record_is_first` | passed incidentally (Django's `-pk` fallback) → **strengthened** |
| `test_a_record_from_before_the_log_renders_without_an_offset` | passed incidentally (stub dumped raw payload) → **strengthened** |

The two incidental passes were tightened before implementing: the ordering test now writes rows oldest-name-first and then dates them in the opposite order, so a screen falling back to insertion order gets it exactly backwards; the append-stage test now also asserts `"None" not in row`, so printing Python's word for the absent offset fails.

Each test was then shown to fail with its target removed, one mutation at a time on the finished implementation:

| mutation | red test |
|---|---|
| drop `ordering` | `test_the_newest_record_is_first` |
| `status_badge` marks an undecodable payload `"skipped"` instead of `"unreadable"` | `test_a_payload_this_version_cannot_read_still_renders` |
| `has_add_permission` returns `True` | `test_nothing_can_be_added_changed_or_deleted` |
| `stream_path` returns `obj.stream_path_key` | `test_the_screen_shows_the_values_not_the_encoded_keys` |
| `_outcome_field` returns `str(value)` unconditionally (no `_ABSENT` for an empty value) | `test_a_record_from_before_the_log_renders_without_an_offset` |

Each mutation reddened exactly one test and left the other five green. `test_the_changelist_renders` is a deliberate smoke case and is not claimed to discriminate.

One test-harness note: `result_list` reads `cl.formset`, which only `changelist_view` sets. The helper sets it to `None` — the value that view sets when nothing on the screen is editable.

## Gates

| gate | result |
|---|---|
| `just fmt` | 488 files left unchanged |
| `just typecheck` | 0 errors, 2 warnings (both pre-existing `reportUnsupportedDunderAll` on the two `__init__.py`) |
| `uv run pytest` | 2476 passed, 16 skipped, 2 xfailed |
| `just api-reference-check` | no drift — nothing exported changed |

SQLite leg only. Nothing here takes a lock or is alias-aware, so no `transaction=True` case was added and the Postgres leg has nothing extra to prove.

## Docs

`CHANGELOG.md` under `[Unreleased] / Added`. Nothing for `UPGRADING.md` — nothing breaks and nothing behaves differently. `docs/examples.md` *Known gaps* re-checked: its outcome entry says what is missing is an example calling `consume()`, which is still true and is not what this PR adds, so it stands unedited.

joshbrooks:
### Review round 1 — nothing blocking, three findings taken anyway (`a5a5ee0`)

**1. A search for a non-ASCII value silently found nothing.** `encode_outcome` renders with `json.dumps`, so a stream named `café/tf611` is stored as `café/tf611`; the screen showed the accent and the search box did not match it. An empty result set reads as "there is no record", which is the conclusion this table exists to prevent — and the docstring claimed searching finds the values the consumer passed, true for ASCII only.

Fixed by escaping the term the same way the payload was escaped before comparing. That is the identity for an ordinary name, and a term carrying a quote or a backslash is escaped to exactly the form the payload stores.

**2. The change page printed the percent-encoded key.** Django puts `str(obj)` in the title, breadcrumb and action labels, and `__str__` returned `ledger@submission%2Ftf611#1` — the exact confusion this screen was built to avoid, on the page most likely to be believed. `__str__` now unquotes; the key columns stay encoded.

**3. `date_hierarchy` removed.** Measured at two `COUNT(*)`s and a `SELECT DISTINCT` over a truncated `recorded_at` per page load, none servable by #261's index, on a table with no size bound until a retention job is scheduled. Newest-first ordering answers the question it was there for.

### Mutations

| mutation | landed | result |
|---|---|---|
| remove the search override | `get_search_results` 1 → 0 | non-ASCII search test red; the ASCII one stays green, so it is not a duplicate |
| `__str__` back to the encoded key | `unquote(` 1 → 0 | page-title test red, 45 green |

Both reverted, tree confirmed clean.

**Not pinned, and said plainly:** the `date_hierarchy` removal has no test. A test asserting an absent option is a restatement of the option, not a check on behaviour — if it comes back, it should come back with the measurement that justifies it.

### Gates

`ruff check` clean · `ruff format --check` 489 files · `pyright src/` 0 errors · `uv run pytest` 2567 passed / 16 skipped / 2 xfailed · `api-reference-check` no drift. Merged with `main` at `c841db2`.

joshbrooks:
### Review round 2 — found a regression in round 1's own fix (`5ea7c27`)

**The non-ASCII search fix broke quoted-phrase search.** Django splits the search box on spaces and *then* strips the quotes around a bit; escaping the whole term first turns the operator's `"` into `\"`, so that branch never fires and the phrase is searched literally. Measured on ASCII-only data: `"submission/tf611"` found 1 row before the override and 0 after; `"alpha beta"` went 1 → 0. That is the same empty page the override was written to prevent, reintroduced by the override — and it was invisible to round 1 because the code it broke did not exist yet.

Fixed by escaping **only the characters that were escaped**, one at a time. Anything already ASCII is left exactly as typed, so every search that worked before this method existed still works character for character, and the docstring no longer claims something about the pipeline that is not true of it.

Two tests were also passing on their names rather than their subjects, both found by round 2:

- `test_a_search_for_an_ordinary_value_still_finds_it` could not see the slice at all — `json.dumps(term)` keeping both quotes left all 46 tests green, because Django strips them back off. There is now a quoted-phrase test, and it is what catches the whole-term escaping.
- The page-title test watched only half of what it renders: the fixture's consumer was `ledger`, which encodes to itself, so removing `unquote` from the consumer half left 996 tests green. The fixture now carries a character the index encodes on both halves.

### Mutations, each proved landed

| mutation | landed | result |
|---|---|---|
| escape the whole term (round 1's version) | `isascii` 1 → 0 | quoted-phrase test red, 46 green |
| drop the slice, `[1:-1]` → `[1:]` | 1 → 0 | non-ASCII search test red, 46 green |
| unquote only the stream half of the title | 1 → 0 | page-title test red, 46 green |

Round 2 also confirmed the `__str__` round trip against `%`, `\`, `+` and a key truncated mid-escape-sequence at every cut point (no exception, U+FFFD or literal passthrough), and agreed the `date_hierarchy` removal needs no test.

### Gates

`uv run pytest` 2568 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `api-reference-check` no drift.

joshbrooks:
### Review round 3 — narrow, and the transform holds

One question this round: can the search override ever find *fewer* rows than no override at all? Answered against 21 real payloads recorded through `DjangoOutcomeStore`, comparing three counts for each of 32 terms — the override, the unbound `ModelAdmin` method, and ground truth read out of every payload in Python.

Proved, with rows rather than reasoning: a quoted phrase containing an accent (`"café tf611"`) finds its row where no override finds none; an emoji outside the Basic Multilingual Plane matches, because the two escapes `json.dumps` emits are byte-identical to what `encode_outcome` writes; a combining mark matches per code point, exactly as the payload is escaped; `%` and `_` still reach Django's own LIKE escaping untouched; and an empty, whitespace, quote-only or backslash-only term is identical to no override.

One narrowing, left as it is: a term containing non-ASCII *whitespace* stops being split into two bits, so it demands adjacency. The row that drops out is one the operator did not ask for, and the row they did type is found only with the override. Stricter, and more correct.

Three rounds, converged. Round 2's finding — that round 1's fix broke quoted-phrase search — is the one that earned the loop here.

PR #261, opened 2026-09-07
Delete old failure records, on an age the operator names

### LINKED ISSUE #253: Old failure records accumulate with no way to remove them

joshbrooks:
Closes #253.

A record is written whenever a consumer cannot apply an event, and nothing ever removed one — so an installation carrying a long-running intermittent problem accumulated years of them, with no way to clear the old ones short of writing SQL by hand. There is now a command that deletes records older than an age you give it: `python manage.py prune_outcomes --older-than-days 365`. It will tell you what would go before anything does (`--dry-run` prints the cutoff date and the count and changes nothing), it deletes in batches so a first pass over a long backlog is not one long lock on a table consumers are still writing to, and the timestamp it filters on gained an index in the same change.

The age has no default and the command refuses to run without one. How long these records are worth keeping is a question about an installation's own obligations rather than about this library, and a default would be applied by whoever ran the command without reading the page — what a wrong guess destroys is the evidence of a problem someone may still be working through. Deployment guidance now has a retention section, and the guided tour has a section for the new command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,23 @@ runnable demo for each.
 
 ### Added
 
+- **`manage.py prune_outcomes` — a retention sweep for the outcome table, plus an
+  index on the timestamp it filters.** Failure records are written only when a
+  consumer cannot apply an event and nothing ever removed them, so an
+  installation with a long-running intermittent problem accumulated them with no
+  way to clear the old ones short of hand-written SQL. The command takes the age
+  to keep and deletes what is older:
+  `python manage.py prune_outcomes --older-than-days 365`.
+
+  **`--older-than-days` has no default and the command refuses to run without
+  it** — the right retention period differs by installation, and a guess destroys
+  the evidence of a problem someone may still be investigating. `--dry-run`
+  reports the count and deletes nothing, deletion is batched (`--batch-size`,
+  default 1000) so a first prune over a backlog is not one long lock, and
+  `--database` targets a named alias. "Older than" is strict: a record stamped
+  exactly on the cutoff is kept. Migration `0011` adds the index the sweep needs;
+  see `docs/deployment.md`. (#253)
+
 - **`JsonlStreamStore` — keep a log in plain text files instead of a database.**
   A stream is a directory of JSON-lines segments; an event is a line. Nothing but
   the filesystem is involved, so a consumer who wants durability without running
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -215,6 +215,39 @@ For an actual deployment, you'd want at least:
 4. **Process supervision.** systemd, s6, or your container orchestrator.
 5. **Static files.** The chat sample has none, but if you add any, run
    `collectstatic` and serve them via the reverse proxy or WhiteNoise.
+6. **A retention policy for outcome records.** Nothing deletes them on its own.
+   See the next section.
+
+## Retention: pruning old outcome records
+
+When a consumer cannot apply an event, the failure is recorded so it can be
+looked at later. Only failures are recorded, so on a healthy system the table
+stays small — but nothing ever removes a row, and an installation that has run
+for years with an intermittent problem will have years of them.
+
+`prune_outcomes` deletes the old ones:
+
+```bash
+# What would go, without deleting anything:
+python manage.py prune_outcomes --older-than-days 365 --dry-run
+
+# Actually delete:
+python manage.py prune_outcomes --older-than-days 365
+```
+
+**There is no default age, and the command refuses to run without one.** How long
+these records are worth keeping is a question about your obligations, not about
+this library: a fortnight is right for one installation and seven years for
+another, and a default would be applied by whoever ran the command without
+reading this page. What it deletes is the evidence of a problem someone may still
+be working through, so the age is yours to state.
+
+"Older than" is strict — a record stamped exactly at the cutoff is kept — and
+deletion is done in batches (`--batch-size`, default 1000) so a first prune over
+a large backlog does not hold one long lock on a table your consumers are still
+writing to. `--database` prunes a named database alias, matching the alias the
+outcome store writes on. Run it from cron or your scheduler at whatever interval
+suits; running it twice deletes nothing the first run did not.
 
 ## Troubleshooting
 
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -186,6 +186,11 @@ No example exercises these yet — a good place to contribute a demo:
   stores). Exported and supported since #251, so the old reason for this gap —
   that an example would document an unstable surface — has expired. What is
   missing now is simply the example: nothing under `examples/` calls `consume()`.
+- Outcome retention. `manage.py prune_outcomes` deletes old failure records
+  (`docs/deployment.md`), but no example records a failure and then prunes it, so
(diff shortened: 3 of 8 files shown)


joshbrooks:
## What is here

- `src/django_rakaia/management/commands/prune_outcomes.py` — the command. Options: `--older-than-days` (int, `default=None`), `--dry-run`, `--batch-size` (default 1000), `--database` (default `DEFAULT_DB_ALIAS`).
- `src/django_rakaia/models.py` — `models.Index(fields=["recorded_at"], name="rakaia_outcome_age_idx")` on `ConsumerOutcome`, plus the docstring paragraph saying why it lands now and not with `0010`. The "there are exactly two because `latest()` is the only query" claim was amended to "the only query that needs a scope", because it is no longer the only query.
- `src/django_rakaia/migrations/0011_outcome_recorded_at_index.py` — `AddIndex`, one operation.
- `tests/test_django_rakaia/test_prune_outcomes_command.py` — 9 tests.
- Docs: `CHANGELOG.md` (`[Unreleased]` → Added), `docs/deployment.md` (new "Retention: pruning old outcome records" section + item 6 in the hardening list), `docs/whats-new.md` (section 20), `docs/examples.md` (Known gaps — a prune demo needs a `consume()` demo first, which is the gap directly above it).

## Decisions asked for

**Dry run: yes.** `--dry-run` runs the same filter through `.count()` and writes the same result line with a `[DRY RUN]` prefix, matching `replay.py`'s existing `[DRY RUN]`/`[APPLIED]` convention. The line carries the cutoff timestamp, not just the age, so a cron log says which date was used — `test_reports_the_cutoff_it_used` pins that. The argument for it is the same as the argument for having no default: the operator is choosing a number whose consequence is irreversible deletion, and the cheapest way to make that choice safe is to let them see the count first.

**Chunking: yes, `--batch-size`, default 1000.** The normal case does not need it — the table is empty on a healthy system. The case this command exists for is the abnormal one, where an installation has never pruned and the first run is over an unbounded backlog. One `DELETE` over that is a long lock on a table the consume loop writes to *from inside its own error handler*, so blocking it turns a slow prune into stalled consumers. The loop re-selects primary keys each pass rather than deleting a sliced queryset (Django cannot `DELETE` a queryset with a `LIMIT`); it terminates because the filter is on a timestamp that only ever moves rows *into* the selection.

**No `transaction.atomic()` anywhere**, so there is no missing-`using=` trap of the kind CLAUDE.md warns about. Each batch's `.delete()` is Django's own transaction on the queryset's own connection, and `--database` is threaded through as `.using(alias)` on every queryset. Plain `django_db` is therefore the right marker here.

**Negative ages are refused** rather than treated as zero. An age in the future selects nothing, so it fails silently as a no-op — more likely a typo than a request.

## Red-then-green

All 9 tests were written first and all 9 were red before any source existed:

```
9 failed in 0.66s
```

Two things worth recording about getting there:

1. One test (`test_negative_age_is_refused`) was **vacuously green** on the first run — it asserted only `pytest.raises(CommandError)`, which `call_command` raises for an unknown command name. Tightened to assert the message names `--older-than-days` before writing any source; it then went red with the other eight.
2. `test_row_on_the_cutoff_is_kept` was initially written against a moving clock — create a row at `now() - 30d`, then have the command compute its cutoff microseconds later. That is unpinnable: the row is *always* microseconds older than the cutoff. Rewritten to freeze the command module's clock via `monkeypatch` and place three rows at cutoff − 1 µs, cutoff, cutoff + 1 µs.

## Mutation testing

| Mutation | Result |
| --- | --- |
| `if days is None: raise CommandError(...)` → `if days is None: days = 365` | **RED** — `test_no_age_refuses_and_deletes_nothing` |
| `recorded_at__lt` → `recorded_at__lte` | **RED** — `test_row_on_the_cutoff_is_kept` |
| `values_list("pk", flat=True)[:batch_size]` → `values_list("pk", flat=True)` (no batching) | **survived** on the first attempt |

The third one is the interesting one. The batching test originally asserted only which rows survived, which one big `DELETE` satisfies just as well. It was strengthened to count `DELETE` statements through `CaptureQueriesContext` — five stale rows at `--batch-size 2` is three statements — and the same mutation is now **RED**.

## Gates

| Gate | Result |
| --- | --- |
| `just fmt` / `ruff format --check` | 491 files unchanged |
| `just typecheck` | 0 errors, 2 warnings (both pre-existing `reportUnsupportedDunderAll`) |
| `uv run pytest` | 2479 passed, 16 skipped, 2 xfailed |
| `makemigrations --check --dry-run` | `No changes detected`, exit 0 |
| `just api-reference-check` | no drift (nothing new is exported — a management command is not a `__all__` name) |
| `uv run zensical build` | `No issues found` |

Not run: the Postgres leg (`just test-pg`). Nothing here locks rows, and the two backend-sensitive details — microsecond precision on the boundary comparison and `pk__in` batching — are the same on both. CI's `test-postgres` job will confirm.

## Expected conflict

`CHANGELOG.md` — the entry is a new first item under `[Unreleased] → Added`. Sibling branches will land entries in the same place; keep all of them.

joshbrooks:
### Review round 1 — nothing blocking; two unpinned guards now pinned

The reviewer traced every destructive path and found none that deletes when it should not. What it did find was two guards the suite could not see:

**`--database` had no coverage.** Stripping `.using(alias)` from both querysets left the full suite green — the code was right and the net was absent. There is now a test that writes a stale row to two databases, prunes one by name, and asserts the other still has its row.

**The `--batch-size < 1` guard had no coverage.** Deleting the guard left the suite green. Without it, `--batch-size 0` slices `[:0]`, breaks on the first pass and reports `deleted=0` — a cron entry that looks like a healthy prune of an empty table.

### Mutations

| mutation | landed | result |
|---|---|---|
| strip `.using(alias)` from both querysets | 2 → 0 | both `TestTheDatabaseFlag` tests red, 11 green |
| delete the batch-size guard | 1 → 0 | both `TestTheBatchSize` tests red, 11 green |

The alias tests are marked `django_db(transaction=True, databases=[...])`. There is no `atomic()` in this command for a plain marker to mask today; the marker is what keeps that true if one ever appears.

### Accepted rather than fixed

Two argument paths exit on a raw traceback instead of a `CommandError` — `--older-than-days 999999999999` overflows, `--database nope` raises `ConnectionDoesNotExist`. Both happen before any delete is issued, both exit non-zero, and every arrival of the flag was probed against a table holding one stale row: **nothing deletes on any of them**. Cosmetic, and wrapping each arithmetic path would add guards with no failure behind them.

The model docstring calls this command the first query on `recorded_at`. With #260 merged the admin's ordering is a second one; the sentence stays true as written, so I have left it.

### Gates

`ruff check` clean · `ruff format --check` 492 files · `pyright src/` 0 errors · `uv run pytest` 2571 passed / 16 skipped / 2 xfailed · `api-reference-check` no drift · migration `0011` applies, reverses and reapplies cleanly. Merged with `main` at `c841db2`.

PR #262, opened 2026-09-07
Sweep where the prose actually lives, and pin that it still does

joshbrooks:
The check that documentation names things which exist read the docs and the library source and stopped there. A comment in an example — read by the very person copying a snippet out of it — could name something gone, and nothing would say so. Widening it to the examples, the scripts and the tests adds thirty-two names, two of which were already dead and both of which turned out to be honest references to history.

The wider half is that nothing held the sweep to its own promise: a reviewer reverted it to reading markdown alone and the suite stayed green, which is this file's own failure mode turned on itself. The same was true of the rule that an exemption forgives one named file rather than a name everywhere. Both now have a test that goes red, and the count the file used to quote about itself is gone, because it went stale within a day — which is exactly what it exists to catch.


joshbrooks:
## What the loop found

Five rounds. **No round was clean, and no finding was a defect in production code** — the diff is one test file. Every round found the same thing one step further out: an input that decides what this checks, sitting downstream of the guard the previous round added.

| Round | The input it narrowed | Result of the mutation |
|---|---|---|
| 1 | the file-set guard covered *kinds*, not patterns | deleting a glob left the count identical, green |
| 2 | the regex was unpinned | dropped one package entirely, green but for an accident |
| 3 | `_mentions()`' loop body — not a declaration at all | `break` took 103 checks to 61, every pin green |
| 4 | the parametrise expression, downstream of every floor | filtering it made the test a tautology, whole suite green |
| 5 | the file set passed per name, **and the floors' own values** | zeroing all eight floors and dropping a directory: green |

Round 5's verdict was that "there is no defence beyond this" was me declaring victory, and it proved the point by finding two more inputs in one round at no cleverness. Its recommendation was to cut rather than harden, and that is what landed: the floors added in round 3 are gone, along with roughly sixty lines arguing for them.

**Three of my own errors are worth recording.** Round 4 proposed folding the parametrised test into one loop, arguing that this made shrinking the iteration and shrinking the data the same edit — I checked instead of accepting it and the argument is false; both narrowings survived the fold, and what actually closes it is a one-line count. Round 2's commit turned out not to contain a docstring fix I believed was in it, because `commit` after a soft reset does not pick up unstaged changes. And rounds 2, 3, 4 and 5 each found a stale repository count in this file's own prose, twice in text I had written in the same round to explain why counts do not belong there.

What survives is the part that works: it catches prose that has rotted, and it found four dead references to one renamed function that three earlier review rounds had found one spelling at a time.

PR #263, opened 2026-09-07
Name the two files after what they serve

joshbrooks:
Two files in this package were named after things they are not. One is the protocol server — the web surface, the wire rules, producer fencing — and carried the name of the small pure function that turns an event into intended changes, which lives somewhere else entirely. The other holds the marker that lets a cache collapse identical waiting requests, and was named for the reading position a consumer keeps between runs, which is also somewhere else. Both reviewers who explored this codebase this week opened the wrong file first and said so.

Nothing importable from the package moved or changed name, so this breaks only code that reached into one of those files directly, and it breaks loudly on the first import rather than quietly later. The changelog and the upgrade notes still name the old paths because they are describing the past, and the check that every name in the prose resolves has a place to record that deliberately.


joshbrooks:
### What moved

| was | is | what it actually holds |
|---|---|---|
| `src/rakaia/handler.py` | `src/rakaia/protocol_server.py` | the ASGI protocol server — PUT/POST/GET/HEAD/DELETE, SSE, producer fencing (~1240 lines) |
| `src/rakaia/cursor.py` | `src/rakaia/response_cursor.py` | the response cursor for CDN request collapsing (protocol §8.1) |
| `tests/test_rakaia/test_handler.py` | `tests/test_rakaia/test_protocol_server.py` | |
| `tests/test_rakaia/test_cursor.py` | `tests/test_rakaia/test_response_cursor.py` | |

The domain sense of *handler* — the pure `event → Effect` function that `docs/glossary.md` defines — lives in `src/rakaia/registry.py` and is untouched. The consumer's reading position lives in `src/rakaia/subscription.py` and `consumer.py`, also untouched.

### Two tests failed on purpose, and both were the point

`test_tier_boundary.py` asserts the ADR 0002 tier map covers every module, so it went red the moment a module was renamed out from under it — updated to list the new names in the same tier. `test_docs_names_resolve.py` (new on main this week) caught `rakaia.handler.STORE_FAILURE_STATUS` in a historical changelog entry, and then caught my own upgrade note naming both old paths. All three are recorded in `_DELIBERATELY_ABSENT`, scoped to the files allowed to say them, which is the pattern already there for the module removed in #249.

### One over-reach, found by the suite

The first pass replaced `rakaia.handler` by substring, which also rewrote `django_rakaia.handlers_autodiscover` — the app failed to start on the next run. Fixed and re-run. Worth stating because it is the argument for doing this rename while nothing else is in flight: a mechanical edit across 19 files needs the whole suite green to be trusted, and it cannot be if branches are landing underneath it.

### Gates

`uv run pytest` 2605 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` no drift after regeneration.

### Not done

No compatibility shim at the old paths. The names in `__all__` did not move, the library is pre-1.0, and a shim for a path nothing outside this repo is known to import would be two files kept alive to serve a hypothesis.

PR #267, opened 2026-09-07
Give batching a file of its own, with its rules written down

### LINKED ISSUE #264: The batching machinery is hidden inside the replay file
When events are replayed, the changes they produce are gathered up and sent to the database in batches rather than one at a time. The rules for when a batch has to be sent before carrying on are subtle — four of them, each guarding against a different way a batch can produce the wrong result — and they live in the middle of an eight-hundred-line file that is otherwise about walking through events and calling the right function for each one.

Nothing else in that file uses them, so they can move out into a file of their own with a small, obvious way in and out. The gain is that someone looking for how batching works can find it by name, and the four rules end up documented in one place instead of being discovered by reading past everything they are not.

joshbrooks:
When events are replayed the changes they produce are gathered into batches rather than sent one at a time, and four rules decide when a batch has to be sent before carrying on. Those rules lived in the middle of the file that walks through events and calls a handler for each one, so the only way to learn them was to read past eight hundred lines they have nothing to do with. They now sit in a file named after what they do, with all four stated together at the top.

Nothing about how a replay behaves changes — the same effects reach the database in the same batches, and the tests that pin those boundaries through a whole replay pass untouched. What is new is a set of tests that exercise the rules directly, without a stream or a handler in sight, so a change to one of them fails on the rule it broke rather than somewhere downstream. Nothing here is visible to anyone using the library, so there is no changelog entry: by long precedent that file documents the library, not the workshop.

Closes #264

### DIFF
--- a/docs/adr/0007-an-outcome-is-recorded-where-the-cursor-is-committed.md
+++ b/docs/adr/0007-an-outcome-is-recorded-where-the-cursor-is-committed.md
@@ -42,7 +42,7 @@ out the obvious implementation.
 `transaction.atomic` (`src/django_rakaia/effect_executor.py:204`). An outcome row
 written inside that block rolls back with the batch whose failure it exists to
 record. Worse, it could not name the event anyway: a stage-0 pass buffers many events
-into one `apply()` (`_StageBuffer`, `src/rakaia/replay.py:434`), and `RowEffect`
+into one `apply()` (`_StageBuffer`, `src/rakaia/batching.py:62`), and `RowEffect`
 carries `model_label` and `lookup` and nothing else (`src/rakaia/effects.py:176-189`).
 By the time effects reach an executor, which event produced which effect is not
 recoverable.
--- a/src/rakaia/batching.py
+++ b/src/rakaia/batching.py
@@ -1,0 +1,146 @@
+"""Holding a replay pass's effects back so the executor sees few, large batches
+— and flushing the moment holding one back would change the answer.
+
+Replay used to call ``executor.apply()`` once per event, so a batch of one was
+the norm: the contiguous-`Update` collapsing in the Django executor could never
+engage (a run of one is never collapsed), and nine events meant nine
+``transaction.atomic()`` blocks, measured at 18 of the 135 statements one form
+save issues (#207). Widening the batch is only free if the executor cannot tell
+the difference, so **the whole design is in the four rules that force a flush**:
+
+* **Before a reorder.** A batch is applied writes-then-deletes-then-retires (see
+  `_write_order_rank`), so an effect ranking below anything pending would be
+  hoisted across an event boundary — flush before adding it.
+* **Before a collision.** A second write to the same column of the same row is
+  ordinary between events (event 2 supersedes event 1) and an error inside one
+  batch, which `check_disjoint_defaults` raises with nothing applied. Asked
+  through the same `_WrittenFields` that check uses, so there is one rule.
+* **Before a duplicate ``produces=``.** Two producers of one correlation id in a
+  batch is `DuplicateProducesError`; in separate events it is normal.
+* **After a transition retire.** Its transitions are synthesised from the report
+  of the call that applied it, and the replay result's external list is
+  documented in handler-emission order, so nothing may ride along behind it.
+  Opted-in retires are rare, so this costs almost nothing.
+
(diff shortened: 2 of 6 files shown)


joshbrooks:
### What moved

`src/rakaia/replay.py` -> new `src/rakaia/batching.py` (private, no `__all__` entry, so `docs/api-reference.md` and `test_public_api.py` are untouched):

| moved | note |
| --- | --- |
| `_StageBuffer` | body unchanged except the constructor |
| `_self_collides` | verbatim |
| `_drain` | verbatim, now typed against a local `_Buffered` protocol (`buffer: _StageBuffer \| None`) so the module need not import `_ReplayCtx` |

`_write_order_rank` stays in `rakaia.effects` (#259) and the buffer still imports it from there.

The one interface change: `_StageBuffer(ctx)` becomes `_StageBuffer(executor, on_report)`. Taking the whole replay context was what tied the buffer to the orchestrator; a sink for each `ApplyReport` is all it actually needed. `replay._make_buffer` wires it, and `_synth_transitions` stays in `replay.py` because `_apply_effects` uses it too.

Followers: `test_tier_boundary.py` gains `"batching"` in `FRAMEWORK`; `test_write_order_conformance.py` now names the batching module as the third reader of the write-order rule (it was patching `rakaia.replay._write_order_rank`, which the buffer no longer reads); ADR 0007's `src/rakaia/replay.py:434` pointer is repointed at the new file.

`replay.py` 826 -> 726 lines. `tests/test_rakaia/test_replay_batching.py` is unmodified and green — that is the behaviour-unchanged evidence.

### Mutation evidence

Each rule removed from `src/rakaia/batching.py` one at a time, with `tests/test_rakaia/test_batching.py` run against it. All five go red at the new module's own interface:

| mutation | new test that goes red | replay-level test that also goes red |
| --- | --- | --- |
| drop the rank arm of `_conflicts` (flush before a reorder) | `test_a_write_behind_a_delete_starts_a_new_batch` | `TestWhatForcesAFlush::test_a_write_after_a_delete_starts_a_new_batch` |
| drop the `_written.collides` arm (flush before a collision) | `test_a_second_write_to_the_same_column_starts_a_new_batch` | `..._same_column_starts_a_new_batch`, `TestBatchingChangesNoRow[same_column]` |
| drop the `_produces` arm (flush before a duplicate `produces=`) | `test_a_repeated_produces_id_starts_a_new_batch` | `..._a_repeated_produces_id_starts_a_new_batch` |
| drop the trailing `Transition` arm of `add` (flush after a transition retire) | `test_a_retire_asking_for_notifications_ends_its_batch` | `test_a_transitions_notification_still_precedes_a_later_events_own` |
| drop `_self_collides` from `add` | `test_an_event_that_collides_with_itself_is_applied_alone` | `test_an_events_own_colliding_effects_still_raise_on_their_own` |

The transition-retire test needed a second look: the first cut followed the retire with an `Upsert`, which the *reorder* rule separates anyway, so it stayed green under the mutation. Following it with a second `Retire` — same rank, no column written, no correlation id — leaves rule four as the only thing that can force the boundary, and it goes red as it should. Worth recording, because it is exactly the "passed with and without the rule" failure mode.

One extra mutation, on the seam the extraction introduced: dropping the `_on_report(...)` call in `flush` (applying but never handing the report back) reddens `test_every_applied_batch_hands_its_report_back` and the replay-level transition-ordering test.

### Gates

`just lint` clean; `just fmt` 499 files unchanged; `just typecheck` 0 errors in `src/` (the 2 pre-existing `reportUnsupportedDunderAll` warnings on `rakaia/__init__.py` are unrelated); `uv run pytest` 2614 passed, 16 skipped, 2 xfailed; `just api-reference-check` clean (no drift — nothing was exported). SQLite leg only; nothing here touches a lock.

joshbrooks:
Review round 1 found nothing blocking, so this converges in one round.

The reviewer verified it is genuinely a move: an AST-level diff of the three relocated bodies against the base shows two byte-identical and one differing only in a type annotation, with the sole code change being the two lines that now hand the report to a sink rather than reaching through the replay context. It then ran a full eight-by-eight mutation matrix over the new module and got a perfect diagonal — every mutation failing exactly one test, and rule four isolated by the rewritten case rather than by the reorder rule that had masked it. It also confirmed the repointed conformance test still catches all three sites through both its identity and behaviour halves.

PR #268, opened 2026-09-07
One recorder in the core for what a replay wrote

### LINKED ISSUE #265: There is no supported way to ask what a replay actually wrote
A replay reports how many changes it applied, but not what they were. Anything that needs the detail — the guarded rebuild that compares a replay against the live rows, for one — has to wrap the thing that does the writing in its own recorder to find out. That workaround currently lives in the Django half of the library, which means the core cannot answer a question about itself, and the same idea is now built twice in two packages under two names.

One recorder that wraps any writer, offered by the core package, replaces both. It makes the compare-a-replay-against-live-rows check possible without Django, which is where it belongs, since nothing about the question is Django-shaped.

joshbrooks:
A replay tells you how many changes it applied, never which ones. Anything that needed the detail had to wrap the thing doing the writing in its own recorder, and the only version of that lived in the Django half of the library — so the core could not answer a question about its own replay, and the same idea existed twice under two names. There is now one recorder in the core package: it wraps any writer, applies through it, and keeps what it passed on. Comparing a replay against the live rows no longer needs Django, which is the point, since nothing about that question is Django-shaped.

It is deliberately invisible from the outside. Whatever the wrapped writer reports comes straight back, an error it raises still reaches the caller, and the batch is read once so a one-shot sequence is never recorded and then applied nowhere. The guarded rebuild now uses this instead of its private copy and produces the same report as before. The dry-run recorder that writes nothing is unchanged: it is the case where no writing code exists to run, which is what makes it safe to point at production.

Closes #265

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,18 @@ runnable demo for each.
 
 ### Added
 
+- **`RecordingExecutor` — apply for real, and keep what you applied.** A replay
+  reports how many effects it applied, never which ones, so anything needing the
+  detail had to wrap the executor itself; that workaround lived privately in the
+  Django package, which meant the core could not answer a question about its own
+  replay. `RecordingExecutor(inner)` wraps any executor, applies through it and
+  holds the effects it passed on in `.effects`. It is transparent by contract:
+  the wrapped executor's report comes back unchanged, an exception it raises
+  propagates, and the batch is materialised so a generator is never recorded and
+  then applied nowhere. `rebuild_and_verify` now composes it in place of its
+  private copy, with no change to the report it returns. `CollectingExecutor` is
+  unaffected — it stays the terminal, writes-nothing case. (#265)
+
 - **`Consumer` and `django_consumer()` — the consume loop as a thing you hold.**
   `consume()` takes the store, the stream, the consumer name, somewhere to load
   the cursor, somewhere to commit it and somewhere to keep outcomes on every
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -112,6 +112,7 @@ page is the contract.
 |---|---|---|---|
 | `Executor` | `rakaia` | `(*args, **kwargs)` | Applies a batch of effects to durable storage. |
 | `CollectingExecutor` | `rakaia` | `() -> 'None'` | An Executor that records effects instead of applying them. |
+| `RecordingExecutor` | `rakaia` | `(inner: 'Executor') -> 'None'` | Applies through another `Executor` **and** keeps what it passed on. |
 | `DjangoExecutor` | `django_rakaia` | `(*, skip_unchanged: 'bool' = False, using: 'str \| None' = None, normalizers: 'Sequence[Normalizer] \| None' = None, batch_updates: 'bool' = False) -> 'None'` | Apply Effects via Django's ORM. |
 | `InMemoryProjections` | `rakaia` | `() -> 'None'` | An in-memory `Executor` **and** `ProjectionReader` over dict-backed tables. |
 | `ProjectionReader` | `rakaia` | `(*args, **kwargs)` | Read-only view over materialised projections. |
@@ -254,16 +255,10 @@ page is the contract.
 | `UpcasterChainError` | `rakaia` | — | Cannot upcast: missing or ambiguous link in the upcaster chain. |
 | `UpcasterConflictError` | `rakaia` | — | Two upcasters were registered for the same (event_match, from_version). |
 
-## Everything else
-
-| Name | Import from | Signature | What it does |
-|---|---|---|---|
-| `RecordingExecutor` | `rakaia` | `(inner: 'Executor') -> 'None'` | Applies through another `Executor` **and** keeps what it passed on. |
-
 ---
 
 ## Appendix — coverage
 
-163 exported names across 16 sections. 144 carry a docstring; 19 do not and show `—` above.
+163 exported names across 15 sections. 144 carry a docstring; 19 do not and show `—` above.
 
 Undocumented: `AnyEffect`, `DEFAULT_NORMALIZERS`, `ENVELOPE_TS`, `Effect`, `GREEN`, `HANDLERS_META_STREAM`, `Normalizer`, `OnErrorPolicy`, `OutcomeStatus`, `PollStatus`, `ProducerValidationResult`, `RED`, `REDUCERS_META_STREAM`, `SCRATCH_PATH`, `Stage`, `UPCASTERS_META_STREAM`, `VACUOUS`, `__version__`, `app`.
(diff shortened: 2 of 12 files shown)


joshbrooks:
## What landed

`rakaia.executors.RecordingExecutor(inner)` — exported from `rakaia`, added to `_EXPORTS` and to the pinned public-API set. Six lines of behaviour:

```python
def apply(self, effects: Iterable[Effect]) -> ApplyReport | None:
    batch = list(effects)
    self.effects.extend(batch)
    return self.inner.apply(batch)
```

`django_rakaia.rebuild._RecordingExecutor` is deleted and `rebuild_and_verify` composes the core one. `Iterable` and `Effect` are no longer imported in `rebuild.py`.

## The `CollectingExecutor` decision: it stays as it is

It is expressible as `RecordingExecutor(NullExecutor())`, and that is the argument against it. `CollectingExecutor`'s promise is not "records effects" — it is **"no code that writes exists in this path"**, which is what makes `replay(store, "orders", CollectingExecutor())` safe to aim at production and what `docs/dry-run-and-executors.md` sells it on. Re-expressing it as a wrapper puts a real `apply()` call in the middle of a dry run, adds a second class (the no-op inner) to keep it honest, and buys no shared code — the body it would share is two lines. The two are different shapes, not two spellings of one shape: one is terminal, the other is a decorator. It also stays a supported export used by 8 test modules, 4 docs pages and the `replay` management command's `--dry-run`, so churning it would cost a deprecation cycle for a refactor with no user-visible benefit.

Recorded in the docs page rather than only here, since that is where the next reader asks the question.

## Contract suite: yes, it belongs

`tests/executor_contract.py` is unchanged; `tests/test_rakaia/test_executor_contract.py` gains `TestRecordingExecutorIsTransparent`, which runs the **whole** existing suite through `RecordingExecutor(InMemoryProjections())` with the inner object as the reader. Transparency is the recorder's only promise, and this is the cheapest way to hold it: every invariant the suite already pins — three ordered passes, one `RefResolver` per batch, `check_disjoint_defaults` before the first write, upsert counts, retire flips, `isinstance(x, Executor)` — has to survive the wrap. A recorder that reordered, re-batched, deduplicated or swallowed would fail there rather than in a consumer's rebuild gate. One extra case (`test_the_contract_run_left_a_recording`) covers the one thing the suite would not notice, since it asserts on rows: an empty recording.

That is 25 additional passing cases, not a new suite.

## Red-then-green

| Test | Red before source | How it failed |
|---|---|---|
| `test_executors.py::TestRecordingExecutor` (7 cases) | yes | `ImportError: cannot import name 'RecordingExecutor'` |
| `test_executor_contract.py::TestRecordingExecutorIsTransparent` | yes | same import error |
| `test_public_api.py::test_all_is_exactly_the_pinned_set` | yes | `Extra items in the right set: 'RecordingExecutor'` |
| `test_rebuild.py::test_the_gate_records_through_the_shared_recorder` | yes | `ImportError` on `rakaia.executors.RecordingExecutor` |
| `test_rebuild.py::test_the_diff_report_is_what_the_private_tee_produced` | **no, by design** | it is the baseline: written and run green against the private tee, then re-run green after the swap. That is the behaviour-preserving evidence — `(RED, compared=2, 1 problem, on row "b")` before and after. |

## Mutation record

Three mutations of the six-line body, each run against the new tests:

| Mutation | Result |
|---|---|
| `self.inner.apply(batch)` then `return None` (swallow the report) | RED — `test_the_inner_report_passes_through_unchanged` |
| drop `batch = list(effects)`, pass `effects` to both | RED — `test_a_generator_is_materialised_before_the_inner_executor_sees_it` |
| record *after* `inner.apply` returns | RED — `test_an_inner_exception_propagates_and_what_reached_it_stays_recorded` |

The third is the one worth naming: recording after applying looks equivalent and is not — a batch that raised partway would vanish from the record, and the recording is what a rebuild gate diffs. Record-first is also what the private tee did, so this is behaviour-preserving as well as correct.

## Gates

- `just lint` — All checks passed
- `ruff format --check` — 497 files already formatted
- `just typecheck` — **0 errors**, 2 warnings (both pre-existing `reportUnsupportedDunderAll` on the two PEP 562 package roots)
- `uv run pytest` — **2638 passed, 16 skipped, 2 xfailed** (SQLite leg)
- `just api-reference` — regenerated and committed; 163 names, was 162
- `uv run zensical build` — No issues found

## Docs

- `CHANGELOG.md` under `[Unreleased] / Added`.
- `docs/dry-run-and-executors.md` — the family table goes from three to four, the mermaid flow gains the wrap, and a new *Asking what a replay actually wrote* section carries the snippet, the transparency contract, and the `CollectingExecutor` argument above.
- `docs/examples.md` *Known gaps* — a new public API with no example is a gap, and this one has none: the demos either write or preview, never both at once.
- No `docs/whats-new.md` section. It is a seam consolidation, not a headline capability, and the tour's bar is "a problem, a snippet, and a one-command demo" — there is no demo to point at, per the gap above. Happy to add one as §22 if you disagree.
- `UPGRADING.md` untouched: nothing breaks and nothing changes quietly. The removed name was private.

Expect a `CHANGELOG.md` / `docs/api-reference.md` conflict with sibling branches — keep both entries and re-run `just api-reference` for the count line.

joshbrooks:
### Review round 2 — nothing blocking; two corrections taken

Round 1's three fixes all verified landed, with the reviewer running eight mutations on the six-line body and killing all eight, each proved live in the executing file first. The grouping fix is confirmed: `just api-reference` reports 0 ungrouped and the name sits between `CollectingExecutor` and `DjangoExecutor` in the executors section, with no catch-all emitted.

Two things it found, both now fixed:

**The class's own description still made the claim the page had been corrected to deny** — "every effect that reached the database". That description is what an editor hover and the generated reference show, so the two surfaces contradicted each other with the wrong one in the more-read place.

**And the corrected page overshot in its second half.** It said a batch that raises partway is "recorded and never lands", which is only true when something outside supplies a transaction — this recorder rolls nothing back, so what the inner executor already wrote stays written. `rebuild_and_verify` is the case that makes it true, and the page now says which.

The reviewer also attacked the `CollectingExecutor` argument properly, finding the best counter — subclass it, keep the greppable guarantee, share the code — and then defeating that counter on two grounds the page did not have: the return types differ, so a dry-run caller is always handed a report, and the subclass would inherit a public `inner` that means nothing for a dry run. Its verdict that the first paragraph does no work and the second carries the load is fair; the second is the one that stayed.

### The merge with #269

Resolved by keeping both sides of the export list. Round 2 flagged this as the same shape as the resolution that cost a day earlier — each branch added a name adjacent to the other's, and taking either whole drops the other silently. The pinned public-API list catches it, and did in the reviewer's test. The reference count is regenerated rather than hand-picked: **170 names, 0 ungrouped**, which is neither side's number.

### Gates

`uv run pytest` 2600 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` no drift.

PR #269, opened 2026-09-07
Give a failure record a name that does not move

### LINKED ISSUE #257: The reason a failure was recorded is just an internal error name

joshbrooks:
A record of an event that could not be applied used to carry the name of whatever class was raised. That meant renaming something internal quietly rewrote the vocabulary an operator had been counting and filtering on, and a genuine bug in a consumer's own code was recorded under the same name as a bad merge key. Everything rakaia raises while applying an event now carries a short code from a published, closed set — written out, never derived from a class name — and anything that is not rakaia's own is recorded under one code with the exception's type beside it and its message left out.

Three failures inside a replay that used to arrive as a bare, unnameable error now have names of their own, and one clause catches anything the library raises from an apply instead of importing six classes from three modules. Nothing that was being raised changed name or stopped being catchable the way it was, so this is additive for anyone already handling these. What does change is the value stored on new failure records, which is why the upgrade notes call it out.

Closes #257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,19 @@ runnable demo for each.
 
 ### Added
 
+- **A failure record now names the failure, not a class name.** Everything rakaia
+  raises while applying an event carries a short, stable reason code, and those
+  codes are a published closed set. The loop records the code rather than the
+  exception's class name, so renaming something internal no longer rewrites the
+  vocabulary an operator has been counting and filtering on. Anything that is not
+  rakaia's own — a bug in your own apply — is recorded under one code,
+  `unhandled`, with the exception's type name beside it and its message left out.
+  Three failures in a replay that used to arrive as a bare `ValueError` (a staged
+  replay with no reader, an event that will not decode, a missing or unusable
+  merge key) now have names of their own; each still subclasses `ValueError`, so
+  code already catching that keeps working. Catch `RakaiaError` to catch anything
+  the library raises from an apply in one clause. (#257)
+
 - **`Consumer` and `django_consumer()` — the consume loop as a thing you hold.**
   `consume()` takes the store, the stream, the consumer name, somewhere to load
   the cursor, somewhere to commit it and somewhere to keep outcomes on every
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -32,6 +32,20 @@ either submodule directly — `from rakaia.handler import ServerOptions` — cha
 path. The failure is an `ImportError` at the import, so nothing gets past a first
 run.
 
+## Recorded reason codes changed value, and nothing raises about it
+
+A failure the loop recorded used to carry the exception's class name as its reason
+code — `HandlerGapError`, `UpcasterChainError`, `ValueError`. It now carries a
+short code from a published closed set: `handler_gap`, `upcaster_chain`,
+`undecodable_event`, and so on, with anything that is not rakaia's own recorded as
+`unhandled` and its type name in `params` under `exception_type`.
+
+Nothing warns and nothing breaks at import. A dashboard, alert or query filtering
+on the old strings simply stops matching, and under this library's own model an
+absent record reads as nothing having gone wrong — which is the failure worth
+naming here. Records already stored keep the old values; only new ones use the new
+codes. Update whatever reads them, and see the reference for the full set.
+
 ## Outcomes are part of the stable surface now
 
 `Outcome`, `OutcomeStatus`, `Stage`, `OutcomeStore`, `InMemoryOutcomeStore`,
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -44,7 +44,8 @@ Nothing warns and nothing breaks at import. A dashboard, alert or query filterin
 on the old strings simply stops matching, and under this library's own model an
 absent record reads as nothing having gone wrong — which is the failure worth
 naming here. Records already stored keep the old values; only new ones use the new
-codes. Update whatever reads them, and see the reference for the full set.
+codes. Update whatever reads them; the full set is tabulated in
+`docs/subscriber-cursors.md`, under "The reason codes rakaia records for itself".
 
 ## Outcomes are part of the stable surface now
 
--- a/docs/adr/0007-an-outcome-is-recorded-where-the-cursor-is-committed.md
+++ b/docs/adr/0007-an-outcome-is-recorded-where-the-cursor-is-committed.md
@@ -218,22 +218,6 @@ will eventually disagree with the first. So outcomes are append-only and attempt
 and "is this still failing?" is the latest outcome for the key, not a column. `unresolved`
 is therefore a derived query, not stored state.
 
-**6c. Rakaia's own codes are a promised set. A consumer's are still the consumer's.**
-Decision 6 says the codes are opaque to rakaia, and it was answering the question of what a
-*consumer* records. It never addressed the case where rakaia itself is what failed, and the
-loop was answering that with `type(exc).__name__` — so the vocabulary an operator reads was
-the internal class names, and any rename rewrote it silently. Everything rakaia raises from
-an apply now inherits one base type and carries a written-out `code`; the set of them is
-closed, published in `rakaia.REASON_CODES`, and changed with the care any other public name
-gets. The code is never derived from the class name, which is the whole defect.
-
-Anything outside that set is recorded as `unhandled` with the exception's type name in
-`params`, and nothing else from it — an exception's message is exactly where a field value
-would leak, which Decision 6 already refused. That keeps the vocabulary countable rather than
-an open set of class names, while still telling two unanticipated bugs apart. None of this
-touches Decision 6: reason codes on an outcome a consumer records remain the consumer's, and
-rakaia still has no opinion about them.
-
 **6b. One translation decides what a stored outcome looks like, and every backend uses it.**
 `encode_outcome`/`decode_outcome` is the only crossing between an outcome and its stored form, the in-memory
 reference included. That store previously kept the object as handed to it while the durable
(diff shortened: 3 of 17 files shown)


joshbrooks:
## The two decisions the issue asked for

**The code lives on the class, not the instance.** `code: ClassVar[str]` on `RakaiaError`, set literally on each subclass. It is a property of the failure *kind*, so a reader can look it up (`HandlerGapError.code`) without constructing one, and there is no per-raise site that can pass the wrong one. `exc.code` reads identically either way, so the loop is unaffected by the choice. The base defaults to `"unhandled"` so an exception from outside the closed set can never break the recording path; `TestTheSetIsClosed` walks `__subclasses__` and fails if any rakaia error is left sitting on that default.

**`unhandled` carries the type name in `params` under a documented key.** `EXCEPTION_TYPE_KEY = "exception_type"`, exported, so a caller filtering on it is not hard-coding a string this repo could change. The type *name* only — never `str(exc)`. ADR 0007 Decision 6 refuses an interpolated message because that is where field values leak, and this library runs on submissions carrying personal and financial data; `test_the_type_name_is_the_only_thing_carried` pins that with a message containing a fake taxpayer id. `params` is `{}` for every code that is ours, so the key's presence is itself the signal that rakaia could not name the failure.

## The closed set

`rakaia.REASON_CODES`, ten members: nine of ours plus `unhandled`.

| Code | Class | Module |
|---|---|---|
| `handler_gap` | `HandlerGapError` | `registry.py` |
| `upcaster_chain` | `UpcasterChainError` | `registry.py` |
| `effect_collision` | `EffectCollisionError` | `effects.py` |
| `unresolved_ref` | `UnresolvedRefError` | `effects.py` |
| `duplicate_produces` | `DuplicateProducesError` | `effects.py` |
| `handler_drift` | `HandlerDriftError` | `drift.py` |
| `missing_reader` | `MissingReaderError` | `replay.py` (new) |
| `undecodable_event` | `UndecodableEventError` | `replay.py` (new) |
| `merge_key` | `MergeKeyError` | `replay.py` (new) |
| `unhandled` | — | the loop, for anything else |

`MergeKeyError` covers three `raise` sites (`replay.py` — no envelope `event_ts` under `ENVELOPE_TS`, `order_key` absent from a payload, values not mutually comparable). All three are one recovery for an operator, so they are one code; splitting them would be a promise the ADR's own recovery table has no use for.

The three new types are `(RakaiaError, ValueError)`. `ValueError` is what those sites raised before, so a caller already catching it keeps catching it — the same reasoning `types.py` records for its store failures ("each subclasses the builtin it replaced"). The duplicate-`stream_paths` `ValueError` at `replay.py:769` is deliberately left bare: it is an argument error raised before any event is read, not something an apply can hit, so it is not part of this family.

## Where the base lives

New `src/rakaia/errors.py`, stdlib-only. It had to be classified in `test_tier_boundary.py`, and it is **SHARED**: the framework tier raises these and the consume loop — which is server tier — reads the code off one. Same reason `outcomes` is shared. Adding it to `FRAMEWORK` would have made `subscription → errors` a tier crossing.

## Red-then-green, per test

Every test in `tests/test_rakaia/test_error_codes.py` was written before the source existed; the first run was `ModuleNotFoundError: No module named 'rakaia.errors'` (collection error, 0 collected). After the implementation, 42 passed.

Two mutations, run against the finished source, because "it passes" is not evidence a test bites:

**1. Derive the code from the class name** — the exact defect being fixed. Replaced `exc.code if isinstance(exc, RakaiaError) else UNHANDLED` in the loop with a `re.sub` camel-to-snake of `type(exc).__name__` minus the `Error` suffix. That derivation produces the *correct* code for all nine classes, so it passes every code-mapping test:

```
9 failed, 33 passed
FAILED TestACodeSurvivesARename::test_renaming_the_class_does_not_move_the_code[HandlerGapError]
FAILED ... [UpcasterChainError] [EffectCollisionError] [UnresolvedRefError]
FAILED ... [DuplicateProducesError] [HandlerDriftError] [MissingReaderError]
FAILED ... [UndecodableEventError] [MergeKeyError]
```

Only `TestACodeSurvivesARename` catches it — nine reds, one per class, from `type("ThoroughlyDifferentName", (cls,), {})` run through the loop. That is the whole point of the file, and the 33 passing tests are the proof that nothing else would have.

**2. Attach `params` unconditionally** — dropped the `if code == UNHANDLED else {}` guard so every failure carried `exception_type`. `9 failed, 33 passed`: every `test_the_loop_records_the_class_s_promised_code[…]` case, via its `assert params == {}`. So the guard is covered per class, not just in aggregate.

Both mutations were reverted; `git diff` against the pre-mutation copy is empty.

## Coverage of the nine, by layer

- **Through the loop:** `TestEachSiteRecordsItsOwnCode`, parametrized over all nine — asserts `reasons == (code,)` and `params == {}`, and cross-checks `Consumed.outcomes` against `InMemoryOutcomeStore.latest`, so the assertion holds for the stored form as well as the returned one.
- **At the real raise sites:** the six pre-existing classes already have raise-site tests across `test_registry.py`, `test_effects.py`, `test_refs.py` and `test_drift_ledger.py` (all still green, unchanged — the classes kept their names). The three new ones get `TestTheRealRaisingSitesRaiseTheNewTypes`: a stage-1 handler with `reader=None` through `require_reader`, `_decode_event(b"not json at all", …)`, and a `merge_replay` on a payload with no `when` key.
- **One clause:** `TestOneClauseCatchesThemAll` — `pytest.raises(RakaiaError)` for each of the nine, plus `rakaia.RakaiaError is RakaiaError` through the lazy package root, plus `issubclass(cls, ValueError)` for the three replay types.
- **`unhandled`:** a locally-defined `RuntimeError` subclass records `("unhandled",)` with `{"exception_type": "ConsumerBlewUp"}`; a bare `ValueError` and an `UndecodableEventError` now record *different* codes, which is the issue's opening complaint stated as an assertion.
- **The set is closed:** `REASON_CODES` equals the nine plus `unhandled`; every code is distinct; every code survives `Outcome`'s string check; no code is literally its class name (so mutation 1 could not have passed by coincidence on one member).

## Exports

`RakaiaError`, `REASON_CODES`, `UNHANDLED`, `EXCEPTION_TYPE_KEY`, `MissingReaderError`, `UndecodableEventError`, `MergeKeyError` — seven names, `169` total (was `162`). `test_public_api.py`'s pinned set and `GROUPS` in `scripts/gen_api_reference.py` both updated (errors into "Errors", the three constants into "Constants"); `docs/api-reference.md` regenerated and committed, `0` ungrouped.

## Docs

- `CHANGELOG.md` under `[Unreleased] / Added`.
- `UPGRADING.md` — the stored value of a reason code changed, and **nothing raises about it**: a dashboard filtering on `HandlerGapError` simply stops matching, and under this library's own model an absent record reads as nothing having gone wrong. That is exactly the quiet-failure case `CLAUDE.md` says belongs here.
- `docs/adr/0007-…md` gains **Decision 6c**, between 6b and 7: rakaia's own codes are a promised set, `unhandled` carries the type and nothing else, and Decision 6 is untouched — a consumer's codes stay the consumer's. Decision 7 gains the one-line note that `sequence_key` is accepted as-is: nothing reads it, removing it is no longer free now that it is exported, and enforcement is revisited on demand.
- `docs/examples.md` *Known gaps* — a new public surface with no example is a gap, and it follows the existing consume-loop gap: nothing under `examples/` runs the loop, so there is no recorded failure for an example to read a code off.

**No `docs/whats-new.md` section.** It would have been 22, and I drafted one and dropped it: `CLAUDE.md` says every section owes a problem, a snippet **and a one-command demo**, and there is no demo to point at — `docs/examples.md` records, in the entry directly above the one I added, that nothing under `examples/` calls `consume()`. A section citing a `just` recipe that does not exist would have been worse than none. Worth reconsidering in whichever PR adds the consume-loop example, where a real demo would exist to cite.

## Gates

| Gate | Result |
|---|---|
| `just fmt-check` | 499 files already formatted |
| `just typecheck` | `0 errors, 2 warnings` (both pre-existing `reportUnsupportedDunderAll` on the two lazy package roots) |
| `uv run pytest` | `2648 passed, 16 skipped, 2 xfailed` (2645 on `main` + 42 new, less the parametrize consolidations) |
| `just api-reference` | `169 names, 0 ungrouped`, committed |

Postgres leg not run: nothing here touches a `select_for_update` site or an alias, and no test in the new file is `django_db`.

## Merge note

Expect a conflict with sibling branches on `CHANGELOG.md` (keep both entries) and on `docs/api-reference.md` (do not hand-merge — rebase, then `just api-reference` and commit; the count line at the bottom is one line two branches will both have rewritten).

joshbrooks:
### Review round 2 — nothing blocking; four findings taken (`b443c77`)

Round 1's four are confirmed genuinely fixed, each re-verified by the reviewer rather than taken on trust. What round 2 found:

**The guard I widened last round still had the same hole, one module short of where it was found.** It recognised a class by the prefix `rakaia.`, and the package root's module name is `rakaia` with no dot — so a codeless type defined there was invisible, and unlike the earlier case there was no unrelated import to catch it by luck. Now recognises both spellings; a codeless subclass in `src/rakaia/__init__.py` reddens the check on its own file (43 passed → 1 failed, 42 passed).

**Two published descriptions were untrue of some of their raise sites.** That matters more than its severity suggests, because the table is now the published meaning of a promised code. `unresolved_ref` covers both a reference no earlier effect produced *and* a field the produced row lacks; `upcaster_chain` covers a missing link, two upcasters that both match, and an event already above the target — not just "could not carry it to the current version".

**Two stale exception mentions** in `merge_replay`'s own argument notes, and one in `replay`'s, still named `ValueError` where the typed errors now raise.

**And the reviewer was right about my justification.** I wrote that the duplicate-paths check "is not a failure the consume loop can record a code for". Not literally true — called from inside an `apply` it reaches the loop and is recorded as `unhandled` with the type in `params`. The decision stands (a code there would be one the published set carries and no outcome can hold); the sentence now says what actually happens instead of overstating it.

### Why no round 3

The changes here are one test-only filter widening, proven by the mutation above, plus documentation prose. Nothing changed in behaviour that a further round could probe. That is the stop condition, stated rather than skipped.

### Gates

`uv run pytest` 2567 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` 169 names, 0 ungrouped.

PR #270, opened 2026-09-07
An example that actually runs the consuming loop

### LINKED ISSUE #254: Nothing demonstrates the new consuming loop

joshbrooks:
Closes #254.

The library documents one loop that reads a stream, applies each event, records anything it could not apply and commits the reading position — and nothing under `examples/` used it, so the first person to wire a consumer was also the first to find out whether the documentation was right. This adds a small Django project that submits progress forms with repeating rows and consumes them, and each of the three ways an event can fail leaves the record that says how to recover it: a row the submitting side's rules decline never reaches the log, because the fact is still sitting upstream in the form; an event that fails to apply stays in the log with the reading position stopped below it, so the next run delivers it again once the missing village is registered; an event for a closed reporting month is declined on purpose, and recorded so a later reader can tell that from never having seen it.

It is a Django example rather than a standalone script because the half worth showing is what is still there afterwards — the walkthrough ends by building a fresh consumer that reads the position and all four records back, and by contrasting a run that halts below a failure with one that skips past it. It runs as part of the demo suite so CI keeps it honest, and the examples page no longer claims a gap that is now filled.


joshbrooks:
## What is here

| Path | What it is |
|---|---|
| `examples/partisipa_intake/intake/gate.py` | The rules a row is checked against, before the log. Pure, no short-circuit, reason codes out. |
| `examples/partisipa_intake/intake/ingest.py` | Splits a form into row events, appends the acceptable ones, records a refusal per declined row. Model-free. |
| `examples/partisipa_intake/intake/consumer.py` | What applying one event means: apply, raise `UnknownSuku`, or return a `skipped` outcome. Writes through a `ProgressRows` protocol. Model-free. |
| `examples/partisipa_intake/intake/rows.py` | The database-backed `ProgressRows` (`update_or_create`, so the apply is idempotent). |
| `examples/partisipa_intake/intake/management/commands/demo_intake.py` | The six-check walkthrough. |
| `tests/test_examples/test_partisipa_intake.py` | 8 tests over the real `django_consumer`, `ConsumerCursor` and `ConsumerOutcome`. |

Wiring: `just intake-demo`, and `just demos` now runs twelve demos instead of eleven. `RAKAIA_STORE = "durable"` — the log is in the database beside the cursor and the outcomes, because "it survives a restart" is the claim.

The example uses `django_consumer(store, path, name)` and `run(apply, on_error=…)` throughout, never the loose-argument `consume()` underneath it.

## TDD record

Tests were written first and run before any source existed:

```
tests/test_examples/test_partisipa_intake.py:36: in <module>
    from intake.consumer import UnknownSuku, make_apply
E   ModuleNotFoundError: No module named 'intake'
1 error in 0.24s
```

Then, after the three model-free modules: `8 passed in 0.65s`.

Mutation check, to show the tests bite rather than merely run — deleting the `continue` in `ingest.submit_form`, so a refused row is appended after being recorded:

```
FAILED tests/test_examples/test_partisipa_intake.py::TestARefusedRowNeverReachesTheLog::test_the_siblings_are_appended_and_the_refused_row_is_not
1 failed, 7 passed
```

## What the demo asserts

| Check | Claim |
|---|---|
| [1] REFUSED | 5 rows appended, 1 refused; the refused subject appears in no log entry; `stage="append"`, `offset is None` |
| [2] HALT | `halted`, position committed at offset 2 while the failure is at offset 3, 2 projection rows, one `project`/`failed` record naming that offset |
| [3] RECOVER | after the village is registered: 3 events handed over (not 5 — nothing already committed is re-delivered), one `skipped` record for the closed month, 2 new rows |
| [4] RESUME | `caught_up`, 0 events handed over, projection unchanged |
| [5] SKIP | same class of failure, `on_error="skip"`: position advances to the head, both failures recorded |
| [6] DURABLE | a fresh `DjangoStreamStore` + fresh `django_consumer`: position at the head, exactly `[(append, refused), (project, failed), (project, failed), (project, skipped)]` |

Check [3] counts the calls into `apply` rather than the projection rows, because a row proves an event was applied at some point and cannot say whether it was applied twice.

## Test placement

`tests/test_examples/__init__.py` puts the example project on `sys.path` so the model-free modules can be imported; the projection half needs the example's own settings, so it is covered by the demo command that `just demos` runs. The consumer tests are `django_db(transaction=True)` — under a plain marker pytest-django has already opened the transaction `DjangoConsumer` refuses to start inside.

## Docs

- `CHANGELOG.md` — new entry under `[Unreleased]`.
- `docs/examples.md` — matrix row, three Django-integration coverage rows, and the *Known gaps* entry for the loop replaced: what is left is the two non-Django outcome stores as an example's own choice, and outcome retention (which now has four records to prune, but still no demo running the command).
- `docs/whats-new.md` — section 22. Sections 19 and 20 each carried a "Not yet demonstrated" paragraph pointing at the gap; both now point at the example instead, which is a correction rather than a rewrite. **Renumber to 23 if a sibling branch lands a 22 first.**
- `okf/` — one example document plus the index and the Django-integration concept, kept in step with `docs/examples.md`.

## Gates

| Gate | Result |
|---|---|
| `just lint` | pass |
| `just fmt-check` | 518 files formatted |
| `just typecheck` | 0 errors, 2 pre-existing `__all__` warnings |
| `uv run pytest` | 2613 passed, 16 skipped, 2 xfailed |
| `just api-reference-check` | no drift (162 names) |
| `uv run zensical build` | no issues |
| `just demos` | all twelve pass, including this one |

Expect a `CHANGELOG.md` conflict with sibling branches — keep every entry.

joshbrooks:
### Review round 2 — nothing blocking; one fix had no cover, and a merge changed a fact

Round 2 confirmed the value check from round 1 is closed **in the right place**: mutating the real projection to write a constant fails `just demos` with the rows it found, while the test suite stays green — proving the walkthrough is the sole cover for the real database projection, exactly as intended. My first attempt only covered the in-memory stand-in and I caught that by measuring rather than reading; both halves are covered now.

**What it found:** the other round-1 fix had no cover at all. Reverting the skipped record's subject from the row back to its offset left both the tests and the walkthrough green — nothing compared that name anywhere, so the one change in this PR that alters observable output could rot without a red build. The walkthrough now asserts and prints it, and reverting it fails the demo.

**And it answered the question I left open.** Two of the four records name a row and two name a position, because the loop-written ones cannot be named by the consumer yet. Rather than hide that, the walkthrough and the README now explain it and point at #272, filed against `django_consumer()` for not passing those arguments through. Keeping the library change out of an example PR was the right call; leaving the table silent was not.

### A cross-PR interaction that neither review could see

Merging #269 changed what this example demonstrates. The loop now records `unhandled` with the exception's type in `params`, not the class name, so this PR's assertion of `("UnknownSuku",)` became wrong the moment #269 landed — neither PR's own CI could have caught it, because each was correct against its own base.

That is a better teaching moment than what it replaced, so the example now shows it: the walkthrough prints `unhandled, UnknownSuku` and says why a consumer's own exception is not one of rakaia's promised codes. `docs/examples.md`'s new gap entry from #269 also claimed nothing under `examples/` runs the loop, which this example makes untrue; it is narrowed to what is still missing — no example *reads* a code off a record.

### Gates

`uv run pytest` 2609 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` 170 names, 0 ungrouped · `just demos` all pass, and CI runs them.

PR #271, opened 2026-09-07
Put back the work a merge resolution dropped

joshbrooks:
Merging the prose sweep resolved every conflicting file in favour of the branch, which had been cut before five pieces of work landed. The result removed about two thousand lines across forty-six files: the consumer object and its Django entry point, the command that expires old failure records together with the index and migration it needs, the tests covering the failure-record screen, and the renaming of two files that were named after things they are not. Nothing complained, because what remained was internally consistent — deleted tests do not fail, and the published names went out along with the code behind them — so every check passed on a tree that was missing a day of work.

This restores that tree exactly as it stood and keeps the sweep, which is the half of the merge that was wanted. The sweep's own check is what found the last piece: it holds a list of names allowed not to resolve because the prose is deliberately describing the past, and the two renamed files had entries there that the older copy of the file did not carry. Those are back, and the check now passes over the restored names.


joshbrooks:
### How it was built, so it can be checked rather than trusted

The good tip is `86e2f55`, the merge's own first parent. The repair is that tree plus the sweep:

1. `git checkout 86e2f55 -- .` — restores every regressed path.
2. `git checkout origin/main -- tests/test_django_rakaia/test_docs_names_resolve.py` — keeps the sweep's rewrite of its own file.
3. Deleted the four files the revert re-created, which step 1 cannot remove: `src/rakaia/cursor.py`, `src/rakaia/handler.py`, and their two test files. This is the gap in `checkout <sha> -- .` — it restores and overwrites, it does not delete.
4. Re-applied three entries to `_DELIBERATELY_ABSENT` for `rakaia.handler`, `rakaia.handler.STORE_FAILURE_STATUS` and `rakaia.cursor`, scoped to `CHANGELOG.md` and `UPGRADING.md`.

**The check on the result:** `git diff --cached --stat 86e2f55` is one file — `test_docs_names_resolve.py`, +171 / −83. Nothing else differs from the tree as it stood before the merge.

### Steps 3 and 4 are the interesting ones

Step 3 was found by a test, not by reading: with both spellings of the module present at once, the tier map reported `cursor` and `handler` as unclassified. Step 4 likewise — the restored tree names the old paths in the changelog and upgrade notes on purpose, and the sweep's rewritten copy of its own file had no entry allowing that, so it failed on exactly the three names the rename had recorded. Each failure named precisely what was missing.

### Test count

2515, against 2605 on `86e2f55`. The difference is the sweep's own deliberate design change: it replaced a per-name parametrised test with one test over every name, and says why in its docstring — a parametrise expression sits downstream of everything that builds the list, so narrowing it shrank what was checked with the suite green. Roughly 88 parameters became 1.

### Verified back

`rakaia.Consumer`, `rakaia.ConsumerCursorStore`, `django_rakaia.django_consumer`, `DjangoConsumer`, `DjangoConsumerCursorStore`, `DjangoOutcomeStore` all import. `prune_outcomes` and migration `0011_outcome_recorded_at_index` are present, and that command's 13 tests pass. `protocol_server.py` and `response_cursor.py` are the only spellings.

### Gates

`uv run pytest` 2515 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` no drift, 162 names, 0 ungrouped.

### Afterwards

The four open PRs (#267–#270) are built on `86e2f55`'s tree, so they apply cleanly once this lands. #268's fix for the ungrouped name is already pushed. Landing order: #267, #269, #268, #270, rebasing each for the changelog, which conflicts three ways.

PR #273, opened 2026-09-07
Cut 0.4.0

joshbrooks:
Six changes since the last release, and the shape of them is one story: the library now delivers events to a consumer and keeps an honest record of what happened to each one. A consumer is something you build once and run rather than six arguments repeated at every call, and it cannot be built in the shape that quietly records nothing; an administrator has a screen for the failures and a command to expire the old ones on an age they name, with no default because a wrong guess destroys evidence; and every failure the library raises itself now carries a short code that renaming something internal cannot change.

The rest is groundwork that shows through. Two files are named after what they serve rather than after things they are not, the rule deciding what order writes reach the database lives in one place instead of three, and there is at last a worked example that runs the whole loop against real shapes and demonstrates each of the ways an event can fail to be applied. This is the prep only — version, changelog and upgrade notes — with the tag to follow once it lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ runnable demo for each.
 
 ## [Unreleased]
 
+## [0.4.0] - 2026-09-08
+
 ### Changed
 
 - **Two modules renamed for what they serve.** `rakaia.handler` is the Durable
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -15,7 +15,7 @@ ones you are crossing.
 
 ---
 
-# Unreleased
+# 0.4.0
 
 ## Two module paths changed, and there is no shim
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@
 # PyPI by an unrelated placeholder. The import name is unaffected:
 #   pip install rakaia-streams   ->   import rakaia
 name = "rakaia-streams"
-version = "0.3.1"
+version = "0.4.0"
 description = "Python server implementation of the Durable Streams protocol"
 readme = "README.md"
 requires-python = ">=3.10"
--- a/uv.lock
+++ b/uv.lock
@@ -1215,7 +1215,7 @@ wheels = [
 
 [[package]]
 name = "rakaia-streams"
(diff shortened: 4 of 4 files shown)


joshbrooks:
### The three release steps, in order

1. `pyproject.toml` — `0.3.1` → `0.4.0` (and `uv.lock` regenerated with it, which is a fourth file the version lives in and is easy to miss).
2. `CHANGELOG.md` — `[Unreleased]` dated as `[0.4.0] - 2026-09-08`, with an **empty `[Unreleased]` put straight back**. Leaving it absent is how a later PR ends up appending into a released section.
3. `UPGRADING.md` — re-headed by release: the `Unreleased` heading becomes `0.4.0`, joining `0.3.1` and `0.3.0` below it.

### What is in it

| # | |
|---|---|
| #255 / #258 | a consumer you hold, which cannot be built recording nothing, and refuses to start inside your transaction |
| #252 / #260 | a read-only screen for the failure records |
| #253 / #261 | `prune_outcomes`, with no default age, plus the index it needs |
| #257 / #269 | reason codes as a published closed set |
| #254 / #270 | a worked example that runs the whole loop |
| #256 / #259, #264 / #267, #265 / #268, #263 | write-order in one place, batching in its own module, one recorder, two files renamed |

Plus #266, which writes down how far this library goes into handling an event and what stays out, and #271, which restored work a merge resolution had silently dropped.

### Verified before proposing the tag

`uv run pytest` 2609 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` 170 names, 0 ungrouped · `uv run zensical build` no issues · `just demos` all twelve pass · `rakaia.__version__` reports `0.4.0` from the installed package, not just the file.

### Breaking, and in the upgrade notes

Two module paths moved (`rakaia.handler` → `rakaia.protocol_server`, `rakaia.cursor` → `rakaia.response_cursor`) with no shim; every exported name still imports from `rakaia` unchanged. And the reason codes stored in existing failure records changed value without anything raising — a dashboard filtering the old strings simply stops matching, which is why it is written down.

PR #274, opened 2026-09-07
Document what the release added, and make the bundle keep up on its own

joshbrooks:
Two things the last release shipped had no home outside the release notes. The screen an administrator uses to read what a consumer could not apply was described nowhere at all, which is a poor result for the most operator-facing thing in it; it now sits beside the command that expires those records, because the same person needs both. And the short codes a failure is recorded under had no entry in the guided tour — deferred at the time because the tour requires a runnable demo per entry and nothing demonstrated them, a reason that expired when the worked example landed in the same release.

The machine-readable bundle under `okf/` had drifted further, because nothing checked it: no entry for the consuming loop, the failure records, the codes, the retention command or the screen, and an update log untouched since the day it was created. It gains all of that as a seventh concept group, plus two checks so the next gap is loud rather than silent — every example directory must have a page there, and the names in its prose are swept for resolution along with the rest of the documentation. It only stayed current this week because somebody happened to notice the directory by hand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ runnable demo for each.
 
 ## [Unreleased]
 
+### Changed
+
+- **Documentation for what 0.4.0 added.** The failure-record screen had no page
+  outside the release notes, and the reason codes had no entry in the guided
+  tour; both are written up now, the screen beside the retention command it
+  belongs with. The machine-readable knowledge bundle under `okf/` gained the
+  consuming loop as a seventh concept group — it had no entry for any of it —
+  and two checks so it cannot drift again: every example directory must have a
+  page there, and the names in its prose are swept for resolution along with the
+  rest of the documentation.
+
 ## [0.4.0] - 2026-09-08
 
 ### Changed
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -216,7 +216,42 @@ For an actual deployment, you'd want at least:
 5. **Static files.** The chat sample has none, but if you add any, run
    `collectstatic` and serve them via the reverse proxy or WhiteNoise.
 6. **A retention policy for outcome records.** Nothing deletes them on its own.
-   See the next section.
+   See [Retention](#retention-pruning-old-outcome-records); the screen for reading
+   them is the section before it.
+
+## Looking at the failure records
+
+The records are only useful if somebody reads them, so they get a screen — the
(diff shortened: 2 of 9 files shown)


joshbrooks:
### The hardening

Two checks in `tests/test_django_rakaia/test_docs_names_resolve.py`:

1. **`okf/**/*.md` joins the swept file set**, so a name in the bundle's prose must resolve like everywhere else. Note the sweep is pinned twice on purpose — the file set and a `_COVERAGE_DECLARATIONS` copy of it — with a message saying to change both deliberately or not at all, because a previous round narrowed the sweep with the suite green. Both were widened. The bundle passed on the first run, so its existing names are all live.
2. **`TestTheKnowledgeBundleKeepsUp`** — every directory under `examples/` must have a page in `okf/examples/`. Mutation: moving `partisipa-intake.md` out reddens it, and it names the missing example plus what to do.

The check runs **one way on purpose**, and the docstring says why: a page with no directory is allowed, because the bundle also describes variants living inside another example (`formkit-submission-stream` is one). Requiring the converse would forbid that on no evidence.

### Documentation

- **`docs/whats-new.md` §23** — the reason codes, with the problem, a snippet, and `just intake-demo` as the one-command demo the tour requires. It points at the demo's `[2] HALT` step, which shows the boundary directly: `unhandled, UnknownSuku`.
- **`docs/deployment.md`** — a section for the screen, before the retention one. Covers where it appears with no wiring, that read-only is enforced rather than conventional, and the two things worth knowing before reading a row: the displayed values are decoded from the payload rather than read from the index columns (`submission/tf611` is indexed as `submission%2Ftf611`), and an unreadable record still gets a row marked as such, because a record vanishing from the page is the failure the table exists to prevent.

### The bundle

New seventh concept, `okf/concepts/consuming-and-outcomes.md`: `poll`/`consume`/`Consumer`, the `Outcome` value and its three stores, the published reason codes, `django_consumer` and its transaction refusal, the retention command, the screen. It carries the two orderings that are load-bearing and unenforceable — the record written outside the apply's transaction, the position committed last per message — and the honest "not yet covered by an example" list. `RecordingExecutor` added to the executors concept. `okf/log.md` written for the first time since 28 July.

### Gates

`uv run pytest` 2610 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` 170 names, 0 ungrouped · `uv run zensical build` no issues.

PR #275, opened 2026-09-07
Let a consumer name the records written on its behalf

### LINKED ISSUE #272: A consumer cannot say how to name the records the loop writes for it
Every failure record names the thing it is about. When a consumer writes the record itself it can use whatever the row is called; when the loop writes one because the work raised, it falls back to the event's position in the log, because the Django entry point does not let a caller say how a message should be named. The result is a list where some records name a row and others name a position, which is exactly the comparison an administrator is trying to make when they open it.

The loop itself already accepts both a way to name what a record is about and a way to say what it is ordered within. The entry point simply does not pass them through. The new intake example shows the gap in its own output and says so rather than hiding it, which is the honest version of the problem but not a fix.

joshbrooks:
A failure record names the thing it is about, and until now a consumer could only choose that name for the records it wrote itself. The ones the loop wrote when the work raised fell back to the event's position in the log, because that is the only name the loop has on its own. The result was a list meant to be read as one thing that held two kinds, and comparing them is exactly why somebody opens it.

The loop has always accepted both a way to name what a record is about and a way to say what it is ordered within; the Django entry point simply did not pass them on, and now does. The worked example passes both and asserts that every record it produces names a row — so the thing it used to point at as a known gap is now the thing it demonstrates.

Closes #272.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ runnable demo for each.
 
 ## [Unreleased]
 
+### Added
+
+- **A consumer can say how the records written for it are named.**
+  `django_consumer()` now takes `subject_of` and `sequence_of` and passes them to
+  the loop, which has always accepted them. Without this the records a consumer
+  writes itself name a row while the records the loop writes for a failed apply
+  name the event's position in the log — two kinds of thing in one column, on the
+  screen where comparing them is the whole point. The worked example passes both
+  and asserts that every record it produces names a row. (#272)
+
 ### Changed
 
 - **Documentation for what 0.4.0 added.** The failure-record screen had no page
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -181,7 +181,7 @@ page is the contract.
 | Name | Import from | Signature | What it does |
 |---|---|---|---|
 | `Consumer` | `rakaia` | `(store: 'CursorStore', path: 'str', name: 'str', cursors: 'ConsumerCursorStore', outcomes: 'OutcomeStore', subject_of: 'Callable[[StreamMessage], str] \| None' = None, sequence_of: 'Callable[[StreamMessage], str] \| None' = None) -> None` | One consumer of one stream, with everywhere it keeps things attached. |
-| `django_consumer` | `django_rakaia` | `(store: 'CursorStore', path: 'str', name: 'str', *, using: 'str \| None' = None) -> 'DjangoConsumer'` | A consumer of `path`, named `name`, keeping both its cursor and its outcomes in the database. |
+| `django_consumer` | `django_rakaia` | `(store: 'CursorStore', path: 'str', name: 'str', *, using: 'str \| None' = None, subject_of: 'Callable[[StreamMessage], str] \| None' = None, sequence_of: 'Callable[[StreamMessage], str] \| None' = None) -> 'DjangoConsumer'` | A consumer of `path`, named `name`, keeping both its cursor and its outcomes in the database. |
 | `DjangoConsumer` | `django_rakaia` | `(store: 'CursorStore', path: 'str', name: 'str', cursors: 'ConsumerCursorStore', outcomes: 'OutcomeStore', subject_of: 'Callable[[StreamMessage], str] \| None' = None, sequence_of: 'Callable[[StreamMessage], str] \| None' = None) -> None` | A `Consumer` that checks the caller has no transaction open. |
 | `ConsumerCursorStore` | `rakaia` | `(*args, **kwargs)` | Somewhere to keep a consumer's watermark between runs. |
 | `InMemoryConsumerCursorStore` | `rakaia` | `() -> 'None'` | Watermarks in a dict: the reference `ConsumerCursorStore`, for tests and demos. |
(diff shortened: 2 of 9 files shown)


joshbrooks:
### The change

`django_consumer()` gains `subject_of` and `sequence_of`, passed to the `Consumer` it builds. `Consumer` has carried both fields since #255 and `consume()` since #248; only the factory was in the way.

### Red first, and each pass-through separately load-bearing

Three tests written before the change: `TypeError` on the unexpected keyword for the two new ones, and the third — that saying nothing still names the position — passed from the start, which is correct, since it documents behaviour that must not change.

| mutation | result |
|---|---|
| drop `subject_of=subject_of` from the call | only `test_a_record_is_named_by_the_caller` red |
| drop `sequence_of=sequence_of` from the call | only `test_the_sequence_a_record_belongs_to_is_the_callers_too` red |

### The example

`demo_intake` now passes both, so all four records it produces name a row. The narrated line that used to explain the gap is replaced by an assertion — `positional = [r.subject for r in records if "/" not in r.subject]` raises if any record names a position instead. `just demos` runs it, so the property is held by CI rather than by prose.

The README's paragraph about the gap becomes the snippet showing how to close it, and `docs/subscriber-cursors.md` and the `okf/` concept page both gained the argument.

### Why `sequence_of` too, when nothing reads it

ADR 0007 Decision 7: the key is recorded and not yet acted on. The field exists so that adding enforcement later does not mean re-deriving groupings a consumer has already decided — which is exactly the argument for exposing it now rather than when something reads it.

### Gates

`uv run pytest` 2613 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` 170 names, 0 ungrouped · `just demos` all pass · `zensical build` no issues.

joshbrooks:
### Review round 1 — nothing blocking; three findings taken (`cfbf44f`)

Both mutations I reported reproduced exactly, and the reviewer added a third I had not run: the **subject** default was covered by nothing before this PR either — the test I wrote for it is what closes that, which I had not realised was new coverage rather than existing coverage.

**The gap in my own testing.** Replacing the *sequence* default (`sequence_of = sequence_of or subject_of`) with a constant left the whole suite green — 2614 passed, 0 failed. The docstring promised the behaviour and nothing read it. There is now a test that passes only `subject_of` and asserts the sequence follows the subject; the same mutation reddens exactly that test and nothing else.

**Two consequences of letting a caller choose a name**, both now in the docstring, and both properties of the choice rather than of this diff:

- `subject_of` is called **only when an apply fails**, so a mistake in it surfaces on the one path you least want surprises on, and it propagates out of the run instead of being recorded. Verified by the reviewer: a namer raising `KeyError` gives 0 rows and no committed cursor.
- The subject is what makes a record distinct — `latest` keeps the newest per subject — so a non-injective namer shows one record where two were written (`GROUPED n=1 rows=2`). Neither failure was possible under the old default, an offset being always readable and always unique.

**And a better argument than the one I gave** for exposing `sequence_of` now. I said recording it later means re-deriving groupings. The reviewer's version is stronger and is what the docstring says: the field is written on every record either way, so omitting the argument does not omit the grouping — it stores a duplicate of the subject in its place. The choice is between recording something true and recording noise.

### Corrections to my own brief

I asked about the subject reaching a derived, percent-encoded, truncated index column. There is no such column — `subject` and `offset` were columns for one round and were removed in #249; only the consumer and stream path are indexed that way. The reviewer probed six shapes anyway: empty, a lone surrogate and a non-string are refused by `Outcome`'s own validation before any write; 5000 characters, slashes, percent signs and non-ASCII all round-trip and group correctly.

I also said nothing reads `sequence_key`. The admin's detail view renders it as "Sequence", so exposing this changes one visible thing — the intended improvement rather than a regression.

### Also fixed

`docs/examples.md`'s coverage row for this example now names the two arguments, which the doc rules require and I had missed.

### Gates

`uv run pytest` 2614 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` no drift · `just demos` all pass.

joshbrooks:
### Review round 2 — converged, nothing blocking

All three of round 1's fixes verified genuinely fixed, and every claim the diff makes about neighbouring code checked against that code rather than taken on trust: that `latest` keeps the newest per subject (`outcomes.py:126-155`, and the collapse reproduced — 2 written, 1 returned), that the admin renders `sequence_key` as "Sequence" (`admin.py:320` and `:381-383`), that ADR 0007 Decision 7 says what the docstring claims it says, and that `subject_of` is called only inside the failure handler (`subscription.py:281-286`) with nothing between it and the caller to catch what it raises.

The sequence-default mutation that survived the whole suite in round 1 now reddens exactly one test, proven landed in the running body. Removing `subject_of` from the demo still fails `just intake-demo` with the message naming the positional subjects.

Two non-blocking notes, both left as they are:

- The demo's `"/" not in subject` is a proxy for "names a row" rather than the property itself. It catches today's regression — offsets carry `_`, never `/` — and the reviewer proved it does. It would go quiet only if an offset format ever contained a slash, which would be a different change with its own review.
- `docs/whats-new.md` is untouched. The tour's bar is a headline capability with a runnable demo, and this is a pass-through of two arguments that already existed — a refinement of sections 20 and 22 rather than a new section. The changelog entry carries it.

Two rounds, and both earned their cost: round 1 found a default that the whole suite could not see, and round 2 confirmed the fix for it was real rather than plausible.

PR #276, opened 2026-09-07
Write down the third way a saved position meets the wrong log

joshbrooks:
The decision record about changing where a log is kept carries a table of what a consumer's saved position does when it meets the wrong store, and that table was missing a case: the same kind of store in two places. Two databases, or two directories, hold two independent logs numbering their events from the same start — so there is not even a difference of format to notice, and a saved position crosses between them more freely than it does between two different kinds of store.

It is the most reachable of the three cases, because this codebase uses that seam deliberately: a from-scratch rebuild replays into a disposable database precisely so it can be verified without touching the real one. Nothing on that path resumes a consumer, so nothing is broken today. This changes no behaviour — it adds the row to the table and pins it with tests, the way the other rows are pinned, so that leaving it is a decision rather than an oversight. Found while investigating #232, which is what would change it.

### DIFF
--- a/docs/adr/0006-changing-backends-is-a-copy.md
+++ b/docs/adr/0006-changing-backends-is-a-copy.md
@@ -38,14 +38,22 @@ All are symmetric — a format is refused in whichever direction it is carried,
 and accepted in whichever direction too.
 
 **The third row is not a variation of the second, and it is the more reachable
-one.** Two `DjangoStreamStore`s on different database aliases, or two
-`JsonlStreamStore`s at different roots, are two independent logs numbering their
-events from the same start. Nothing about them differs in format, because nothing
+one.** Two `DjangoStreamStore`s on different database aliases are two independent
+logs numbering their events from the same start. Nothing about them differs in format, because nothing
 about them differs at all except where they keep their data — so there is not even
 a format to compare, and a cursor crosses between them as freely as within one.
 Measured on this tree, three events on one alias and ten on another: a position
 saved against the first resumes against the second as `advanced`, delivering seven
 events and skipping three permanently.
+
+Worth knowing while reading the tests: the second row is *modelled* by two
+`JsonlStreamStore`s at different roots, which is literally this third row — the
+file says so, on the grounds that `rakaia.offsets` sees a format rather than a
+class, and for the offset module that is exactly right. What demonstrates the
+second row for real is the pin that `DjangoStreamStore` issues the same format
+this one does (`tests/test_django_rakaia/test_offset_format_pin.py`). So one
+setup has been standing in for both rows, and only the alias case below separates
+them.
 
 It is the more reachable row because this codebase uses that seam on purpose.
 `DjangoStreamStore(using=…)` exists so a from-scratch rebuild can replay into a
@@ -54,7 +62,7 @@ argument, and `rebuild_and_verify` composes all three. Nothing in that path
 resumes a consumer, so nothing is wrong today — but "the same store class" reads
 as "the same store", and it is not.
 
-The second entry is new. Until `JsonlStreamStore` existed, every pair of stores
+The `DjangoStreamStore` ↔ `JsonlStreamStore` entry is new. Until `JsonlStreamStore` existed, every pair of stores
 disagreed about format, so every cross-store cursor was refused by accident of
 shape. That protection is gone for one pair of three, and it was never a designed
 guarantee — `rakaia.offsets` refuses a cursor whose format it can see belongs to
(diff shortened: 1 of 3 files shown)


joshbrooks:
### The measurement

Three events on `default`, ten on `overlay`, a position saved against the first and resumed against the second:

```
DEFAULT cursor = 00000000000000000003
OVERLAY head   = 00000000000000000010
resumed -> status=advanced  rewound=False  messages=7
```

Seven delivered where ten exist. Three skipped permanently, no error and no rewind — and `poll` is behaving correctly, because there is nothing for it to see: both offsets are the same format and the same value. `test_the_offsets_are_indistinguishable` asserts exactly that, `mine == theirs`.

### Why this is not row two in different clothing

Row two is two *backends* that happen to share an offset format — a coincidence of shape, and `rakaia.offsets` refuses cross-format pairs precisely because it can see the difference. Row three is one backend at two locations: nothing differs except where the data is kept, so there is no format question to ask and no seam where an answer could be given. That is why the fix #232 proposes — a short identifier naming the backend — would not catch it: `DjangoStreamStore()` and `DjangoStreamStore(using="overlay")` would both report `durable`. Recorded on that issue.

### These tests assert a defect, so the mutation runs the other way

There is nothing to break to make them red. The check that they are reading real behaviour is to make the two stores *one* log — pointing the second `_seed` at the default alias — which turns the resumed position into a caught-up one and fails `test_a_cursor_from_one_alias_skips_events_on_the_other` (1 failed, 1 passed). Reverted, tree clean.

They live in `test_using_seam.py` rather than beside the cross-backend pair because the seam is the subject: `DjangoStreamStore(using=…)` is what this file is about, and the third row is a property of that argument rather than of the offset module.

### Gates

`uv run pytest` 2612 passed / 16 skipped / 2 xfailed · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` no drift. No source changed.

joshbrooks:
### Review round 1 — nothing blocking; three findings taken (`303c93c`)

The reviewer verified every claim in the diff against the code — the `using=` argument on all three classes, that `rebuild_and_verify` composes them, that **nothing on that path resumes a consumer** (no `poll`, `Consumer` or `load_cursor` in `rebuild.py` or `hermeticity.py`), and it reproduced both measurements independently, including the two-file-store case.

**The finding that mattered was a sentence contradicting the repo's own tests.** The paragraph said "the third row is not a variation of the second" and then offered *two file stores at different roots* as its example — which is precisely the arrangement `test_cross_backend_cursors.py` uses to stand in for the second row, with the same 3-vs-10 numbers and the same expected delivery. So the new row's canonical example was the old row's demonstration.

Fixed by naming only the two-database case, which is unambiguous and measured, and by saying what the discovery actually exposed: one arrangement has been demonstrating both rows, and only the alias case separates them. What demonstrates the second row for real is `test_offset_format_pin.py`, pinning that `DjangoStreamStore` issues the same format the file store does.

**Two smaller ones.** "The second entry is new" became a positional reference the moment the table grew to three — in a record whose companion test file explicitly refuses positional references, because three drafts miscounted this very table. It names the row now. And the two test files pinning neighbouring halves did not point at each other, so a reader landing on either could not tell what was already covered; both now do.

### A misplaced mutation of my own, and how it was caught

I set out to reproduce the reviewer's best mutation — tagging offsets with their alias, the shape a #232 fix would take. My first attempt tagged `get_current_offset` and came back **green**, which reads exactly like a weak test. It was not: `poll(...).cursor` comes from the last *message's* offset, not the head, so the mutation never reached the assertion. Tagging the offsets returned by `_read_since` reddens `test_the_offsets_are_indistinguishable` as it should.

Worth recording because the wrong conclusion was one step away, and it is the expensive one — I would have reported my own test as vacuous.

That mutation is also the useful answer to my question about whether these tests merely restate their setup. They do not: `test_the_offsets_are_indistinguishable` is a tripwire that goes red the day offsets become alias-aware, which is the day #232 lands.

### Both judgement calls upheld

**Placement stays.** `test_cross_backend_cursors.py` is deliberately database-free, and `test_offset_format_pin.py` records that precedent verbatim — split out rather than folded in *because it needs a database*. Two aliases plus `django_db(transaction=True, databases=[…])` cannot live in the core package. A pointer was added instead.

**No changelog entry is right.** No behaviour change, no exported name, nothing a consumer could notice. If an entry is owed anywhere it belongs to #232, which is what would change the behaviour.

### Postgres

The reviewer ran it — the one leg where an alias plus `transaction=True` might differ. 25 passed, no difference.

### Why no round 2

This round's fix touched prose and two test docstrings. No test body and no source changed, so there is no behaviour a further round could probe. Stated rather than skipped.

### Gates

`uv run pytest` 2612 passed / 16 skipped / 2 xfailed — baseline 2610 plus the two new tests · `pyright src/` 0 errors · `ruff check` and `format --check` clean · `just api-reference-check` no drift.

PR #279, opened 2026-09-10
Name the change labels and metadata keys an event carries

### LINKED ISSUE #277: Give the event envelope's metadata a declared shape
## Summary

`StreamMessage.metadata` and `AppendOptions.metadata` are `dict | None`. The keys inside are a
convention — `user`, and by documented intent `url` and `causation` — and nothing enforces the
shape at either end. A producer that writes `actor` instead of `user` gets no error; the actor
simply stops appearing, and the first sign is a history screen with a blank column.

This is one of the three gaps the library's own survey names in its boundary surface, and it is
the one with a consumer already exposed to it. Everything else on that boundary is
contract-grade — `Effect` is an exhaustive union of frozen dataclasses, `Outcome` is frozen and
runtime-validated with a stable codec, the store seams are `@runtime_checkable` Protocols with
shipped conformance suites. The envelope is the exception.

## Suggestion

A `TypedDict` for the known keys, `total=False`, with `dict[str, Any]` still accepted so nothing
breaks:

```python
class EnvelopeMetadata(TypedDict, total=False):
    user: Any
    url: str
    causation: str
```

Additive, matches the house style (declare the shape, keep the runtime permissive), and it gives
`envelope.append_event` — already the one chokepoint where the actor key is centralised — a type
to name rather than a convention to remember.

Worth noting for the Python floor: `total=False` on the whole class keeps this portable to 3.10,
where `NotRequired` is unavailable without `typing_extensions`.

## Context

Found while designing typed contracts across the three repositories that meet here
(partisipa-import, formkit-ninja, rakaia). Not blocking anything — partisipa pins
`rakaia-streams==0.3.*` and cannot consume a new release until that pin moves — so this is a
candidate for whenever the next minor goes, not a request to cut one.

### LINKED ISSUE #278: The change label is a closed set everywhere but its type
## Summary

`StreamMessage.label` and `AppendOptions.label` are free-form `str`, but the read side treats
them as a closed set: `history.label_marker` special-cases `insert` / `create` / `delete` and
maps everything else to `~`. So a typo lands in the log as a change marker rather than an error,
and the log is the system of record — the wrong marker is permanent.

Every other closed vocabulary in this library is a `Literal` alias rather than an `Enum` —
`Stage`, `OutcomeStatus`, `PollStatus`, `OnErrorPolicy`, `OnDriftPolicy`, `DeleteEventType`.
`label` is the one that got missed, and it is the one that reaches durable storage.

## Suggestion

```python
ChangeLabel = Literal["insert", "create", "update", "delete"]
```

…as the annotation on the two `label` fields, with `str` still accepted at run time so an
application-specific label is not a hard error. That keeps the existing behaviour — anything
unrecognised still maps to `~` — while making the four the library actually acts on visible to a
consumer's type checker.

Two things to decide, which is why this is a suggestion rather than a patch: whether `update`
belongs in the set given `label_marker` reaches it only via the fallback, and whether narrowing
the annotation on a field consumers already write to is a minor or a patch under the pre-1.0
policy.

## Context

Found while designing typed contracts across partisipa-import, formkit-ninja and rakaia. Not
blocking — partisipa's `==0.3.*` pin means no rakaia release is consumable there until the pin
moves. A candidate for the next minor rather than a reason to cut one.

joshbrooks:
An event's change label and its metadata were typed as a bare string and a bare dictionary, so nothing told a reader which labels the history view recognises or which metadata key holds the actor. This adds two named types listing the values rakaia itself reads and writes: the four change labels, and the `user` / `url` / `causation` metadata keys. The append options and stored messages now use them, so an editor offers the known values.

Both types stay open on purpose. An application label such as `import`, or an extra metadata key, is still accepted — which also means a misspelling is still not a type error. That was a deliberate choice: the library's own code and tests use labels outside the four, so enforcing the set would have been a breaking change. Closes #277, closes #278.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ runnable demo for each.
   name the event's position in the log — two kinds of thing in one column, on the
   screen where comparing them is the whole point. The worked example passes both
   and asserts that every record it produces names a row. (#272)
+- **The change label and the event metadata now have named types.**
+  `ChangeLabel` lists the four labels rakaia acts on (`insert`, `create`,
+  `update`, `delete`), and `EnvelopeMetadata` lists the metadata keys it reads or
+  writes (`user`, `url`, `causation`). The `label` and `metadata` fields on
+  `AppendOptions` and `StreamMessage` are annotated with them, so an editor can
+  offer the known values. Both still accept any string or dictionary, so an
+  application label such as `import`, or an extra metadata key, is not a type
+  error — and for the same reason a misspelling is not one either. (#277, #278)
 
 ### Changed
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -53,7 +53,9 @@
   the lockfile. Do **not** set it to `latest`, which this file used to advise:
   pyright nags that a newer release exists, and taking its suggestion typechecks
   against a different pyright than CI — the nag is cosmetic, the divergence is
-  not. Django's synthesised attributes are declared explicitly (see the
+  not. The recipe silences the nag; a bare `uv run pyright` still prints it.
+  Upgrading pyright is its own change: bump the pin in `pyproject.toml` and the
+  `justfile` together. Django's synthesised attributes are declared explicitly (see the
   `if TYPE_CHECKING` blocks in `django_rakaia/models.py`) rather than waved
   through with ignores.
 - **Lint and format take no path arguments.** `[tool.ruff]` in `pyproject.toml`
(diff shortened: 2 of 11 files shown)


joshbrooks:
## Detail

**What changed**
- `rakaia.types`: new `ChangeLabel = Literal["insert", "create", "update", "delete"]` and `EnvelopeMetadata(TypedDict, total=False)` with `user: Any`, `url: str`, `causation: str`. Both exported from `rakaia` and grouped under *Reading and writing streams* in `gen_api_reference.py`; `docs/api-reference.md` regenerated.
- `AppendOptions.label` / `StreamMessage.label` → `ChangeLabel | str`; `.metadata` → `EnvelopeMetadata | dict[str, Any] | None`. `label_marker` takes `ChangeLabel | str`.
- `context.merge_provenance(explicit=...)` widened from `dict[str, Any]` to `Mapping[str, Any]`: a TypedDict is not assignable to `dict[str, Any]` (it is to `Mapping`), which pyright flagged at `store.py:293` and `jsonl_store.py:842/930`. The body already only reads it.
- `history.envelope_actor` reads `.get("user")` once instead of `.get()` and then `["user"]`: pyright does not narrow a TypedDict key through `.get(...) is not None`.

**Why not enforced (the question #278 left open)**
Labels in use outside the four: `fold_events(label="import")` default in `django_rakaia/envelope.py`, and in tests/examples `created`, `archived`, `verify`, `""`. A strict `Literal` on `AppendOptions.label` would make all of those type errors. Picked "named, not enforced" with the maintainer before writing. `update` stays in the set because `@stream_model` writes it, even though `label_marker` only reaches it through the fallback. With nothing narrowed, this is additive, so it can go in a patch release.

**Tests**
- `test_history.py`: `ChangeLabel` members equal the four, and between them map to all three markers (catches drift between the alias and `label_marker`); a label outside the set is still accepted; `EnvelopeMetadata` is all-optional and its `user` is the key `envelope_actor` reads; extra metadata keys pass through.
- `test_public_api.py`: both names added to the expected public set.

**Gates run locally**
`ruff check` clean, `ruff format --check` clean, `just typecheck` 0 errors, `pytest` 2620 passed / 16 skipped / 2 xfailed, `just api-reference-check` clean, `zensical build` clean. The Postgres leg was not run locally; nothing here touches locking or the database.

